PR #51, opened 2024-07-05
Issue 46 sysadmin delete button

rlamacraft:
The delete user sysadmin menu item is now disabled if the deployment property is set, with an explanation as such.
image

Furthermore, I've fixed the error handling for the network request to ensure that any other reason for the deletion operation failing is made apparent.


rlamacraft:
@mKowalski256 Do you want to double check that this works as intended before I merge it in?

mKowalski256:
Yes, will take a look.

mKowalski256 (approved):
Re-tested on [AWS build](https://issue-46-sysadmin-delete-button-4.researchspace.com/), looks good to me!

PR #55, opened 2024-07-11
Rsdev 148 api performance

rs-fraser:
This PR improves the performance of:
- api search
- workspace 'View all' (equivalent of an empty - i.e. returning all results - search)

Previously in both of the above features, we would first limit the list of records which could be searched in various ways, including by permissions and filtering.

This caused slow performance in cases such as on pangolin85 as sysadmin or for large customer installs, where there are lots of documents. 

On the UI search which performs well, we first search all of the indexed documents, then apply any filters or permission-based view rights to the smaller list of search results. 

In addition, in the case of an empty search or 'View All' on the workspace where we list all documents, we would first of all loop through each of the users which the subject user can view (e.g. members of a PI subjects groups, or everyone for a sysadmin) and perform a db call for each to get their documents. Instead, the db call now takes a list of user ids and performs an IN query so that there only needs to be 1 db call. 

The above changes have improved performance on pangolin85, operating as sysadmin, as follows:
- search via api for specific term (chemical):
  - from ~30s to under 2s (i.e. not logged as slow call in the logs, where the limit is set to 2s)
- 'View All' in the workspace or empty search:
  - from ~30s to ~6s

There is still probably further we could go with this as in the case of an empty (i.e. all records) on an instance with 100k records in scope this will still be slow, but hopefully won't time out. I think another avenue to address would be the way we handle permissions checking.

Additionally, there are a bunch of code cleanup changes included in the files which I touched.


src/main/java/com/researchspace/api/v1/controller/DocumentsApiController.java
```diff
@@ -144,7 +144,7 @@ public ApiDocumentSearchResult getDocuments(
      * Converting search results to ApiSearchResult
      */
     ApiDocumentSearchResult searchResult = new ApiDocumentSearchResult();
-    List<ApiDocumentInfo> info = new ArrayList<ApiDocumentInfo>();
+    List<ApiDocumentInfo> info = new ArrayList<>();
```
rs-nicof, line 147: Do you think it would be more performing in this case if you create a `LinkedList` here?

src/main/java/com/researchspace/dao/hibernate/RecordDaoHibernate.java
```diff
@@ -299,7 +299,7 @@ public List<RSpaceDocView> getRecordViewsById(Set<Long> dbids) {
   }
 
   @Override
-  public Set<BaseRecord> getViewableRecordsForUser(User user) {
+  public Set<BaseRecord> getViewableRecordsForUsers(List<User> users) {
```
rs-nicof, line 302: Could we use a `Set<User>` as an actual parameter instead? I guess that would save uys to re-map it at the end?

src/main/java/com/researchspace/service/impl/RecordManagerImpl.java
```diff
@@ -1222,16 +1215,16 @@ private List<BaseRecord> getRecordsSharedWithUser(User u, boolean onlyTemplates,
         recordGroupSharingDao.getSharedRecordsWithUser(u), PermissionType.READ, subject);
   }
 
-  private Set<BaseRecord> getViewableRecordsByUser(User u, boolean onlyTemplates) {
+  private Set<BaseRecord> getViewableRecordsByUsers(List<User> users, boolean onlyTemplates) {
```
rs-nicof, line 1218: Same, can we use `Set<User>` in the parameters?

src/main/java/com/researchspace/service/impl/RecordManagerImpl.java
```diff
@@ -1158,25 +1148,19 @@ private Folder getFolderOrRecordParent(Long parentId, User user, Record original
    * Retrieve all viewable documents by specific user (Role).
    *
    * @param subject the user whose viewable records we are retrieving
-   * @param onlyTemplates
-   * @return
    */
   private List<BaseRecord> getViewableRecordsByRole(User subject, boolean onlyTemplates) {
-    List<BaseRecord> result = new ArrayList<>();
-
+    Set<BaseRecord> viewableRecords = new HashSet<>();
```
rs-nicof, line 1153: not sure what we do with this collection but it looks like that a `LinkedHashSet` scanning a set would be slightly more efficient here

src/main/java/com/researchspace/service/impl/RecordManagerImpl.java
```diff
@@ -1198,20 +1182,29 @@ private List<BaseRecord> getViewableRecordsByRole(User subject, boolean onlyTemp
...
+            getViewableRecordsByUsers(List.of(viewableUser), onlyTemplates));
+        if (viewableUser.hasRole(Role.PI_ROLE)) {
+          // if the subject is a PI of group A, which contains another PI as a group member, they
+          // can only view the docs of the other PI which has been shared with group A (whereas the
+          // subject PI can view ALL records created by a non-PI), so filter the other PIs records
+          // by their permissions to ensure correct visibility
+          subjectsViewableRecords =
+              subjectsViewableRecords.stream()
+                  .filter(br -> permissnUtils.isPermitted(br, PermissionType.READ, subject))
+                  .collect(toCollection(HashSet::new));
```
rs-nicof, line 1202: Here qwe are re-createing the set and reassign every loop.
is there any chance we could just remove from the original set the ones that are not permitted?

src/main/java/com/researchspace/service/impl/RecordManagerImpl.java
```diff
@@ -1222,16 +1215,16 @@ private List<BaseRecord> getRecordsSharedWithUser(User u, boolean onlyTemplates,
...
+      return recordDao.getViewableTemplatesForUsers(users);
     }
-    return recordDao.getViewableRecordsForUser(u);
+    return recordDao.getViewableRecordsForUsers(users);
   }
 
   @Override
-  public Set<BaseRecord> getViewableTemplates(User user) {
-    return recordDao.getViewableTemplatesForUser(user);
+  public Set<BaseRecord> getViewableTemplates(List<User> users) {
```
rs-nicof, line 1226: What do you think about having Set<User>`? 
I mean if we put all in the `Set` (for the `Users` that are always uniquely treated) all the retainAll and join operations will be very efficient rather tha using the `List`s

rs-nicof (changes requested):

src/main/java/com/researchspace/api/v1/controller/DocumentsApiController.java
rs-fraser: I didn't actually make any change here, only removed the pre-java 7 type declaration. This arraylist is also only used as a return from the api method so there won't be any performance benefit in using a `LinkedList` here.

src/main/java/com/researchspace/dao/hibernate/RecordDaoHibernate.java
rs-fraser: Yes good shout will update this

src/main/java/com/researchspace/service/impl/RecordManagerImpl.java
rs-fraser: Good point. The reason for this probably inefficient looping was to cover the case where the PI of
group A can see the records of their group member, who is a PI of group B. In that case, PI A should only be able to see the records shared with group A by PI B, and none of their other work in group B.

But by doing so we are individually fetching each users records from the db and not making use of the new method which takes the set of user ids. I'll add instead just a single loop at the end through all records, and if the owner of that record is a PI, check permissions and remove if not permitted. Should speed things up a bit, thanks!

src/main/java/com/researchspace/service/impl/RecordManagerImpl.java
```diff
@@ -1198,20 +1181,28 @@ private List<BaseRecord> getViewableRecordsByRole(User subject, boolean onlyTemp
...
+      // add records created by each of the subjects viewable users
+      subjectsViewableRecords.addAll(getViewableRecordsByUsers(viewableUserIds, onlyTemplates));
+
+      // Remove any records the subject is not permitted to see. Using PI role check as a
+      // short-circuit to avoid unnecessary permissions check.
+      // Handles case where the subject is a PI of group A, which contains another PI as a group
+      // member, they can only view the docs of the other PI which has been shared with group A
+      // (whereas the subject PI can view ALL records created by a non-PI), so filter the other PIs
+      // records by their permissions to ensure correct visibility.
+      subjectsViewableRecords.removeIf(
```
rs-fraser, line 1199: @rs-nicof now instead of looping through each user, getting their records, then checking permissions if they are a PI, we now get all records in one db call, and have 1 pass through the list, first checking if the user is a PI, then perform a permissions check. 

In most cases the user won't be a PI and the check will short-circuit and avoid checking permissions, and this way avoids multiple db calls to get the initial records.

src/main/java/com/researchspace/service/impl/RecordManagerImpl.java
rs-fraser: I don't think there will be any benefit of using a `LinkedHashSet` here. Any performance benefit in iteration is probably balanced out by the extra time taken for adding new elements and maintaining the order. Also we don't really iterate over this list any more either, and it gets converted back to an arraylist as the return.

rs-fraser:
@mKowalski256 @rs-nicof I am off this coming week. I've tested out the changes I made in response to the inefficiency @rs-nicof  pointed out yesterday on pangolin and all seems to be working fine. Please feel free to make any changes and merge this PR when you guys are happy with it.

rs-nicof (commented):
il looks ok to me!

src/main/java/com/researchspace/dao/hibernate/FullTextSearcherImpl.java
```diff
@@ -166,6 +191,19 @@ private List<BaseRecord> filterAndSortQueryList(
...
+      resultList.retainAll(sharedRecords);
+    }
+
+    if (srchConfig.getFilters().isFavoritesFilter()) {
+      List<BaseRecord> userFavourites =
+          recordUserFavoritesDao.getFavoriteRecordsByUser(
+              srchConfig.getAuthenticatedUser().getId());
+      resultList.retainAll(userFavourites);
+    }
+
```
mKowalski256, line 206: This is to re-apply filtering that was previously handled by SearchManager.searchWorkspaceRecords(), right? Looking through the filters that used to be handled there, one missing here is ontology filter. You can actually see that filtering by ontology files no longer works after UI search on the branch (i.e. after applying 'ontology files' filter the result list shows all documents, not just ontologies).

[Workspace-ResearchSpace.webm](https://github.com/user-attachments/assets/6fb8700a-5360-4d2d-ad79-ecc0c61ccc35)

mKowalski256 (approved):
Looks good - retested on https://rsdev-148-api-performance-16.researchspace.com/.

PR #59, opened 2024-07-15
RSDEV-280: Return the list of Subsamples when cannot delete a Sample

rs-nicof:
This change include changes to how the error message when deleting samples is shown. In the screenshot below, a sample with two subsamples in containers and a lone subsample from a different sample have been deleted. The former has failed as the subsamples are in containers (with the main part of this change being that those subsamples are now detailed and linked to), and the latter has been deleted correctly hence the green alert.

Screenshot 2024-07-16 at 11 08 16


rs-nicof:
✅ Build full java testds passed: https://jenkins.researchspace.com/job/rspace-web/job/RSDEV-280-backend/16/

mKowalski256:
I've looked through the code, and eventually decided to change the way how info about which subsamples are problematic is encoded into API response. @rs-nicof to avoid enlarging sample response object I've only kept `canBeDeleted` flag on ApiSample level, removed `subSamplesInContainer` array, and instead added `isStoredInContainer` flag to ApiSubSampleInfo model, which means it can be read from existing `ApiSample.subSamples` array.

@rlamacraft that needs adjusting frontend code to new logic, basically in `/ui/src/stores/models/Search.js` the alert generation code should check `canBeDeleted` flag on sample level (as before), then (if flag is false) iterate over subSamples array, and only list subSamples with `isStoredInContainer === true`.

rlamacraft:
Oh right yeah, each sample might contain a mixture of subsamples that can or cannot be deleted. Ok, made the change

PR #65, opened 2024-07-22
Rsdev 240 pdf page numbers

rs-fraser:
This PR fixes the page numbering where the user has selected not to restart page numbering for each doc in a multi-doc or notebook export, by passing the existing length of the pdf doc into the pdf writer i.e for a new doc or when page numbering should restart each doc, the length is 0, otherwise it's the length of the existing doc.

Also rename the config property `pageName` to `restartPageNumberingPerDoc` throughout the front and back-end code, as `pageName` was very confusing. 

And finally reduced the logging level for printing the pdf html input to debug since that's probably not something we need to log every time we export pdf as it could be very long.


rs-nicof:
The code looks good to me!

mKowalski256 (approved):
Code change looks good to me, and works fine on [test AWS build](https://rsdev-240-pdf-page-numbers-4.researchspace.com/).

PR #72, opened 2024-07-29
Rsdev 258 reactivate google drive

rs-fraser:
Reactivates the google drive integration by replacing deprecated google apis. 

This is part 1 of 2 of fully fixing Google Drive, as until our app is authorised by Google there is a user cap of 100 users, which we have maxed out. This _should_ re-activate the integration for any of those 100 users, although we don't know which accounts are part of the 100 users, only that the cap has been reached.


rlamacraft:
Looks fine to me

rs-fraser:
Seems to be an issue I wasn't seeing locally on AWS about a missing parameter, looking into it. 

https://rsdev-258-reactivate-google-drive-4.researchspace.com/

rs-fraser:
I've added a check for the required google keys being present, and if not the user is informed and asked to check with their rspace admin, so this is good to go again.

Build is up with the integration active at https://rsdev-258-reactivate-google-drive-7.researchspace.com/

mKowalski256 (approved):
Looks good & works fine for me on aws build.

PR #74, opened 2024-07-31
issue-73: reworking API key display logic and ways for sysadmin to access it

mKowalski256:
This PR consists of multiple changes that should make retrieval of API key more secure and traceable.

The changes in behaviour:
- for user who generated API key the value of their key is no longer sent to front-end whenever 'My Profile' page is displayed, but only after user explicitly clicks on 'Show Key' link
- when System Admin user 'operates as' they can no longer see user's API key ('Show Key' returns an error); System Admin can still generate new API key or revoke the old one, as there may be valid reasons for these actions (similarly how we do allow System Admin to reset user's password)
- the `/apiKeyInfo/all` endpoint within `UserDetailsSysAdminApi` is now only available when deployment property `sysadmin.apikey.access` is set to `true` (it's `false` by default)

### DIFF
--- a/src/main/webapp/WEB-INF/pages/userform.jsp
+++ b/src/main/webapp/WEB-INF/pages/userform.jsp
@@ -299,7 +299,7 @@
 <div class="bootstrap-custom-flat">
 	<div class="col-xs-9" style="padding: 0px">
 		<hr />
-		<c:if test="${canEdit and !sessionScope['rs.IS_RUN_AS']}">
+		<c:if test="${canEdit}">
 			<div class="api-menu__header col-xs-12">
 				<fmt:message key="user.apikey.label" />
 			</div>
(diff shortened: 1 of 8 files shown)


mKowalski256:
test AWS build running at https://issue-73-api-key-6.researchspace.com

src/main/webapp/WEB-INF/pages/userform.jsp
```diff
@@ -567,14 +561,14 @@
 <script type="text/template" id="apiKeyDetailsTemplate">
 	{{#enabled}}
     <div class="api-menu__key col-xs-8">
-      {{#key}}
-        <strong>Key</strong>: <span id="api-menu__keyValue">{{key}}</span>
+      {{#revokable}}
+        <strong>Key</strong>: <span id="api-menu__keyValue"> ... </span>
```
rs-fraser, line 565: should we display some text here instead like the message from the api 'API key details cannot be accessed when 'operating' as another user'? or maybe disable the button and add a tooltip explaining?

rs-fraser (approved):
Code looks good.

Tested on the AWS build and it's all working fine. Added 1 small suggestion but happy to approve as is too.

src/main/webapp/WEB-INF/pages/userform.jsp
mKowalski256: ok yes, I've changed "operating as" error to be a standard error, which UI can report the normal way (error popup).

PR #75, opened 2024-07-31
RSDEV-301 Add link to PyRAT server at top of generated table

rlamacraft:
This change adds a row to the top of the table that is inserted into the rich text fields of documents by the PyRAT integrations. This row includes a link back to the PyRAT server and the date of when the table was generated.

image


mKowalski256 (approved):
New link looks good and works fine for me (on localhost). 

I'm slightly confused that inserted pyrat table in view mode doesn't seem to have styling of the regular table (no cell/row border lines), but that's unrelated to the issue tackled in this PR.

PR #76, opened 2024-07-31
RSDEV-302 Shrink size of generated PyRAT table

rlamacraft:
This change reduces to the font size of the generated PyRAT table that is inserted into the rich text fields of documents by the PyRAT integration. This is done to reduced the probability of horizontal scrolling being required most of the time.

This image shows the new font-size at the top and the old size just below it.

image


mKowalski256 (approved):
Looks good to me - new table now fits the page width.

PR #85, opened 2024-08-13
RSDEV-224 Cosmetic tweaks to audit page

rlamacraft:
This change includes:
- Removing spaces between words and commas
- Consistently applying link to name of filter, and not the conjunctive words
- Formats time according ISO 8601


src/main/webapp/WEB-INF/pages/audit/auditing.jsp
```diff
@@ -46,24 +37,13 @@
...
-	  <a href="#" id="users" class="addRow">
-		  <spring:message code="audit.users.label"/>
-	  </a>
+		<spring:message code="audit.activity.filter.label"/>
+		by <a href="#" id="domains" class="addRow"><spring:message code="audit.domains.label"/></a>,
+		by <a href="#" id="actions" class="addRow"><spring:message code="audit.actions.label"/></a>,
+		by <a href="#" id="oids" class="addRow"><spring:message code="audit.identifiers.label"/></a>,
+		by <a href="#" id="dates" class="addRow"><spring:message code="audit.date.label"/></a>
+    and
+		by <a href="#" id="dates" class="addRow"><spring:message code="audit.users.label"/></a>
```
mKowalski256, line 46: should be `id="users"` in this line

mKowalski256 (commented):
Looks good, but I've noticed one incorrect `id` element after refactoring - makes 'users' filter not appearing for the PI.

Also, as the content of 'Time' column is now shorter, I believe the width of that column could be reduced from 25% to 20% [here](https://github.com/ResearchSpace-ELN/rspace-web/blob/ae98842118ee27530e6f0347c1c91c7529657571/src/main/webapp/WEB-INF/pages/audit/auditSearchResultTemplate.html#L5).

src/main/webapp/WEB-INF/pages/audit/auditing.jsp
rlamacraft: oops, yep, fixed

rlamacraft:
> Also, as the content of 'Time' column is now shorter, I believe the width of that column could be reduced from 25% to 20% [here](https://github.com/ResearchSpace-ELN/rspace-web/blob/ae98842118ee27530e6f0347c1c91c7529657571/src/main/webapp/WEB-INF/pages/audit/auditSearchResultTemplate.html#L5).

Only down to 22% as otherwise it wraps where there's a long timezone offset

mKowalski256 (approved):
Looks good to me!

PR #88, opened 2024-08-16
RSDEV-326 Update existing tests and add new tests covering time validation for sample fields

rs-fraser:
~~This PR uses the new "time" type from `rspace-core-model` so that we can validate times are in the correct format for custom template fields in inventory samples., and adds some mvc tests around that validation.~~

This PR updates existing tests now that the `complexSampleTemplate` used in many tests has a "time" field added, which previously was the only type of sample field missing from that template.

Also added some tests around the validation of time fields. Validation has been added in the related `rspace-web` PR: https://github.com/rspace-os/rspace-core-model/pull/7/files


src/test/java/com/researchspace/api/v1/controller/SamplesApiControllerMVCIT.java
```diff
@@ -1083,4 +1080,62 @@ public void changeSampleOwner() throws Exception {
...
+  private String makeJsonSampleWithIdAndTime(String id, String time) {
+    return "{ \"templateId\": \""
+        + id
+        + "\",\"fields\": [ { \"content\": \"3.14\" }, {}, {},  { \"type\": \"Text\","
+        + " \"content\": \"text content\" }, {}, {}, {}, {}, {}], \"extraFields\" : [   {"
+        + " \"name\": \"extraFieldName\", \"type\" : \"time\", \"content\": \""
+        + time
+        + "\" } ],"
+        + " \"expiryDate\":null, \"name\": \"sample1\" }";
+  }
```
mKowalski256, line 1140: as discussed on rspace-core-model PR, this test should actually try setting an invalid _field_ of "Time" type, rather than _extraField_ of such type. For that it must use a Sample Template that has a Time value, e.g. ComplexSampleTemplate created in dev/run environment....

... only that looking at [RecordFactory](https://github.com/ResearchSpace-ELN/rspace-core-model/blob/a08e8be5c1e7cf41887f2cb97c9055d165f0ac6e/src/main/java/com/researchspace/model/record/RecordFactory.java#L328) class I see that "CompexSampleTemplate" doesn't have Time field. That must be by accident, it has all the other field types. It would be good to extend the template to have Time field too, it'll make it easier to test the functionality in unit test but also on localhost UI.

pom.xml
```diff
@@ -1737,7 +1737,7 @@
     </dependency>
     <dependency>
         <groupId>com.github.rspace-os</groupId>
-        <artifactId>rspace-core-model</artifactId>
+        <artifactId>rspace-core-model-rsinv-930</artifactId>
```
rs-fraser, line 1740: Note to self: remove this after `rspace-core-model` changes have been merged

src/test/java/com/researchspace/api/v1/controller/SamplesApiControllerMVCIT.java
rs-fraser: Thanks, this makes sense. I've added the additional time field to the `complexSampleTemplate` and updated existing tests.

src/test/java/com/researchspace/api/v1/controller/SamplesApiControllerMVCIT.java
```diff
@@ -448,14 +449,11 @@ public void createEditSampleWithFields() throws Exception {
...
-            .filter(
-                t ->
-                    t.getName()
-                        .equals(ContentInitializerForDevRunManager.COMPLEX_SAMPLE_TEMPLATE_NAME))
-            .findAny();
-    return complexTemplate;
+    return sampleApiMgr.getAllTemplates(user).stream()
+        .filter(
+            t ->
+                t.getName().equals(ContentInitializerForDevRunManager.COMPLEX_SAMPLE_TEMPLATE_NAME))
```
rs-nicof, line 455: would it be null pointer safer to invert the parameters? like
```
                ContentInitializerForDevRunManager.COMPLEX_SAMPLE_TEMPLATE_NAME).equals(t.getName())
```

rs-nicof:
It looks good to me, I'ìve added just a minor comment

src/test/java/com/researchspace/api/v1/controller/SamplesApiControllerMVCIT.java
rs-fraser: I think `name` from `getName()` is NOT NULL in the DB, but no harm in being sure so I've swtiched them round like you suggest :+1:

mKowalski256 (approved):

PR #90, opened 2024-08-19
RSDEV-257 Add "Move all to trash" button to error toast when deleting samples

rlamacraft:
When deleting a sample, if some of the subsamples are in containers then the user is prevented from being able to delete the sample as it is likely that some of the substance is still in use/in freezers/etc. However, in the case where a sample only ever has one subsample, or there happens to only be one subsample left, it would be very useful to be able to delete the sample and subsample both in one operation. Furthermore, there may be edge cases where it is desirable to delete all of the subsamples of a sample in a storage and the system should be flexible enough to support that it even if its is an unusual request.

As such, this change adds a button to the error toast that explains that some of the subsamples are still in containers that allows the user to force the deletion of all of the subsamples and sample even though the subsamples remain in containers.


mKowalski256 (approved):

PR #95, opened 2024-08-27
Rsdev 275 transfer forms on user deletion

rs-fraser:
When a sysadmin deletes a user, and the user has created forms which have been used by other users to create documents, those forms will be transferred to the sysadmin so that the deletion can take place.

If this is the case, the sysadmin is show a warning when they try to delete the user that forms will be transferred under their ownership:
image

Also some code cleanup on the files I changed.

~~I'm still working on the format of the audit trail message (it should show type FORM and probably the form ID) but everything else is good to go so raising the PR now so we can begin review earlier with the release coming up.~~
 
The accompanying `rspace-core-model` [PR](https://github.com/rspace-os/rspace-core-model/pull/11) add some annotations to the Form classes, so that the relevant fields display properly in the audit trail: 
Screenshot from 2024-08-27 12-22-18


mKowalski256 (approved):
Code changes look good to me and work fine on AWS build.

Functionality-wise I think there should be more changes to UI dialog that is shown to sysadmin when user has forms - the transfer part is new and has non-intuitive consequences, but I feel the new text it's easy to miss if someone has used this dialog before an no longer goes through the text. 

E.g. the part explaining the forms could be more distinguished graphically, with more/shorter fragments in bold. The label of confirmation button could be changed to 'transfer forms and delete'. Something like that:
image

rs-fraser:
> Code changes look good to me and work fine on AWS build.
> 
> Functionality-wise I think there should be more changes to UI dialog that is shown to sysadmin when user has forms - the transfer part is new and has non-intuitive consequences, but I feel the new text it's easy to miss if someone has used this dialog before an no longer goes through the text.
> 
> E.g. the part explaining the forms could be more distinguished graphically, with more/shorter fragments in bold. The label of confirmation button could be changed to 'transfer forms and delete'. Something like that: image

Updated version:
image

PR #100, opened 2024-08-30
PRT-803: Fix wrong return value when connection is not alive for Diigital Commons Data

rs-nicof:


rs-fraser (approved):
All good.

Confirmed the issue on pangolin85 - after 5 mins of inactivity I could see the upload failed with an exception.

Tested these changes on an AWS build and after 10mins of inactivity the upload to Digital Commons worked as expected.

PR #109, opened 2024-09-06
RSDEV-196 Have dependency cruiser check for incompatible licenses

rlamacraft:
This dependency cruiser rule encodes a whitelist of strings of which at least one must appear in the license string of every third-party library we install from npm. As such, by running dependency cruiser, any violations will be reported.

### DIFF
--- a/src/main/webapp/ui/.dependency-cruiser.js
+++ b/src/main/webapp/ui/.dependency-cruiser.js
@@ -1,6 +1,15 @@
 /** @type {import('dependency-cruiser').IConfiguration} */
 module.exports = {
   forbidden: [
+    {
+      name: "compatible-licenses",
+      comment: "Third-party code should adhere to a license that is compatible with our own",
+      severity: "error",
+      from: {},
+      to: {
+        licenseNot: ["MIT", "BSD", "Apache", "Hippocratic"]
+      }
+    },
     // {
     //   name: 'no-circular',
     //   severity: 'warn',


tilorspace (commented):
@rlamacraft do we need to specify GPLv3 explicitly in this allowlist as well or is that accounted for elsewhere?

rlamacraft:
> @rlamacraft do we need to specify GPLv3 explicitly in this allowlist as well or is that accounted for elsewhere?

It's just doing a string matching, so anything with "GPL" will match. If there's a meaningful difference between the two versions then I can look at make the check more sophisticated

mKowalski256 (approved):
Looks good to me, but are we running dependency-cruiser as a part of any code building workflow? I've run `npx depcruise src` in `src/main/webapp/ui` dir to test the new rule, and license check was fine but the run has shown a pre-existing error for other rule (see [my log](https://github.com/user-attachments/files/16931992/dep-cruise-log-matthias.log)
). I haven't seen that other error reported on jenkins or during react rebuild, which suggest that either it happens just on my local, or it's not run/not failing the builds.

Probably unrelated, but I also cannot get Jest Tests pass on this branch [on jenkins] (https://jenkins.researchspace.com/job/rspace-web/job/rsdev-196-enforce-js-licenses/) (even though I've merged latest `main` branch).

rlamacraft:
> Looks good to me, but are we running dependency-cruiser as a part of any code building workflow?

It does a handle things, of various severity, that are mostly just hints for debugging but as we're using it more and more having it be run with flow is probably a good idea

> Probably unrelated, but I also cannot get Jest Tests pass on this branch

I can't figure out what's going on with those tests; sometimes they start failing consistently and other times it comes and goes, especially on localhost

PR #114, opened 2024-09-16
RSDEV-333 The UI tweaks to not show the API key

rlamacraft:


mKowalski256:
I've started AWS build at rsdev-333-api-key-ui-tweaks-9.researchspace.com.

The changes look good to me, but they don't seem to cover all points from [RSDEV-333](https://researchspace.atlassian.net/browse/RSDEV-333) ticket:
> - My Profile page should only tell user whether they have an active API key, and if they do, when it was generated 
> - when new API key is generated/regenerated it should be shown to the user with information they need to note it down, because there will be no way to retrieve it later (UI change)

rlamacraft:
> when it was generated

done

mKowalski256 (approved):
Tested on https://rsdev-333-api-key-ui-tweaks-13.researchspace.com, looks good to me!

Page before key generation:
image

Page after key generation:
image

Page for user with key generated some time ago:
image

rs-nicof (approved):

PR #115, opened 2024-09-17
Rsdev 292 inventory thumbnail performance

rs-fraser:
### Background
Loading an inventory record can be slowed down by the front-end requesting multiple of the same images. On loading an inventory record, various images can be loaded:
	- main container image
	- sample image(s)
	- sub-sample image(s)
	- thumbnail versions of each of the above

In many cases the image is the same (images can inherit from their parent e.g. all sub-samples use parent sample image unless a different one is specified). We load (and store) these separately though, keyed on `InventoryRecord` (Container, Sample or Sub-sample) ID. The images are retrieved via multiple endpoints e.g.
	- `/samples/{sampleId}/image/{timestamp}`
	- `/samples/{sampleId}/thumbnail/{timestamp}`
	- `/subsamples/{sampleId}/image/{timestamp}`
	- `/subsamples/{sampleId}/thumbnail/{timestamp}`
	- (`{timestamp}` in each of the above examples gets ignored by the back-end)

These multiple endpoints reduce the chance to cache what is often the same image for each of those items.

### Changes
`FileProperty` now contains a column `contentsHash`, which gets populated when a new image is added. Any `FileProperty`, keyed on `username` and `contentsHash`, will be re-used if the same image is being saved.

A new endpoint `/inventory/file/image/{contentHash}` has been added and is used where we generate HATEOAS links for an `InventoryRecord`, which tells the front-end the url to retrieve the images(main image, thumbnail image, and in the case of container location image) from, meaning any request from the front-end for the same image by the same user uses the same endpoint for improved cache-ability.

A liquibase migration has also been added to add the new `FileProperty.contentsHash` column, and populate that column with the hash of each the existing inventory files. 

Related `rspace-core-model` PR: https://github.com/rspace-os/rspace-core-model/pull/14
Jenkins build: https://jenkins.researchspace.com/job/rspace-web/job/rsdev-292-inventory-thumbnail-performance/37/


src/main/java/com/researchspace/service/inventory/impl/InventoryApiManagerImpl.java
rs-fraser, line 1: The only actual changes here are the renaming of the method to `createImagesForRecord`. The rest are formatting changes where the order of fields and methods has been changed.

src/main/java/com/researchspace/dao/customliquibaseupdates/StoreFileContentsHash_rsdev292.java
```diff
@@ -0,0 +1,60 @@
...
+  public String getConfirmationMessage() {
+    return String.format(
+        "Populated contentsHash column in the FileProperty table for %d entities.", fileCounter);
+  }
+
+  @Override
+  protected void doExecute(Database database) {
+    List<FileProperty> fileProperties = getAllFileProperties();
+    for (FileProperty fileProperty : fileProperties) {
+      if (fileProperty.getContentsHash() == null || fileProperty.getContentsHash().isEmpty()) {
```
rs-nicof, line 35: Could you use `StringsUtil.isBlank(...)`?

src/main/java/com/researchspace/service/inventory/impl/InventoryApiManagerImpl.java
```diff
@@ -236,77 +236,88 @@ public ApiInventorySearchResult convertToApiInventorySearchResult(
...
+  }
 
-    String fileName = imageName + "." + imageExtension;
-    return inventoryFileApiManager.generateInventoryFileProperty(user, fileName, imageIS);
+  /*
+  The same FileProperty (and therefore the same file on disk) can belong to many InventoryRecords.
+  Checks if a FileProperty already exists for the given user and hash of the contents of the image
+  and returns that if so. Otherwise, generates a new FileProperty.
+   */
+  private FileProperty saveOrRetrieveImage(
```
rs-nicof, line 268: Is this actually saving? would it be `retrieveOrCreateImage` a better method name?

src/main/java/com/researchspace/service/inventory/impl/InventoryApiManagerImpl.java
```diff
@@ -236,77 +236,88 @@ public ApiInventorySearchResult convertToApiInventorySearchResult(
...
+      User user, String fileName, InputStream imageIs, String contentsHash) throws IOException {
+    Optional<FileProperty> existingFile =
+        getExistingFilePropertyForImage(contentsHash, user.getUsername());
+    if (existingFile.isPresent()) {
+      return existingFile.get();
+    } else {
+      return inventoryFileApiManager.generateInventoryFileProperty(
```
rs-nicof, line 275: Is this actually saving? would it be `saveNewInventoryFileProperty` a better method name?

src/main/java/com/researchspace/dao/customliquibaseupdates/StoreFileContentsHash_rsdev292.java
rs-fraser: Yeah good call, have updated it

src/main/java/com/researchspace/service/inventory/impl/InventoryApiManagerImpl.java
rs-fraser: Yeah it also saves that actual file to the filestore eventually, as well as the `FileProperty` metadata. 

That's why i renamed the enclosing method from `generateInventoryFilePropertyFromBase64Image` to `saveOrRetrieveImage` because I was confused about where the file itself actually got saved.

I've renamed this method as well now, good shout.

mKowalski256 (approved):
Tested on localhost and AWS build, looks good for me!

I've added some minor code comments, the only more-significant one is about liquibase update, which I've noticed will be calculating hashes for all inventory files (i.e. including inventory attachments), while I think the idea was just to hash preview images & thumbnails? But the consequences won't be large, just that some extra files will get the checksum calculated.

src/main/java/com/researchspace/api/v1/InventoryFilesApi.java
```diff
@@ -65,4 +67,9 @@ ApiInventoryFile uploadFile(MultipartFile file, ApiInventoryFilePost settings, U
 
   @DeleteMapping(value = "/{id}")
   ApiInventoryFile deleteFile(Long id, User user);
+
+  @GetMapping("/image/{contentsHash}")
+  ResponseEntity<byte[]> getImageByContentsHash(
+      @PathVariable String contentsHash, @RequestAttribute(name = "user") User user)
```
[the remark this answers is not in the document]
rs-fraser: It turns out actually since spring 5 annotations on the method params on Controller interfaces are honoured https://www.baeldung.com/spring-interface-driven-controllers. 

I've removed `@RequestAttribute` to match the rest of the methods. If I remove `@PathVariable` though it generates a warning in IntelliJ. I think it makes sense to define the `@PathVariable` at the same place where the `@GetMapping` with the placeholder value is though i.e. on the interface.

src/main/java/com/researchspace/api/v1/model/ApiContainerInfo.java
```diff
@@ -258,11 +258,16 @@ protected String getSelfLinkEndpoint() {
...
 
-    if (isLocationsImage()) {
-      buildAndAddInventoryImageLink(ApiLinkItem.LOCATIONS_IMAGE_REL, inventoryApiBaseUrl);
+    if (getLocationsImageFileProperty() != null) {
+      addLocationImageLink(inventoryApiBaseUrl);
     }
   }
 
+  void addLocationImageLink(UriComponentsBuilder baseUrlBuilder) {
+    String imageType = "locationsImage";
```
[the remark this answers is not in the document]
rs-fraser: No reason, have fixed :+1:

src/main/java/com/researchspace/api/v1/model/ApiInventoryRecordInfo.java
```diff
@@ -541,24 +548,34 @@ protected boolean applyChangesToSharingMode(InventoryRecord invRec, User user) {
...
+  void addMainImageLink(UriComponentsBuilder baseUrlBuilder) {
+    String imageType = "image";
+    String contentsHash = getImageFileProperty().getContentsHash();
+    addImageLink(baseUrlBuilder, contentsHash, imageType);
+  }
+
+  void addThumbnailLink(UriComponentsBuilder baseUrlBuilder) {
+    String imageType = "thumbnail";
+    String contentsHash = getThumbnailFileProperty().getContentsHash();
+    addImageLink(baseUrlBuilder, contentsHash, imageType);
```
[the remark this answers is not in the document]
rs-fraser: fixed.

src/main/java/com/researchspace/dao/customliquibaseupdates/StoreFileContentsHash_rsdev292.java
```diff
@@ -0,0 +1,61 @@
...
+              e.getMessage());
+        }
+      }
+    }
+  }
+
+  private List<FileProperty> getAllFileProperties() {
+    return sessionFactory
+        .getCurrentSession()
+        .createQuery("from FileProperty where fileCategory = 'inventory'", FileProperty.class)
```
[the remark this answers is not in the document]
rs-fraser: Yeah good point. It's probably not a big issue, but it would be confusing why some inventory attachments have a hash which isn't used, and like you say it will take longer too. 

I've updated so that we go via each entity like you suggest.

PR #124, opened 2024-09-20
RSDEV-174 Attach Gallery files to Inventory records with new picker

rlamacraft:
This change takes the new Gallery page, which is not yet ready to replace the whole Gallery page, and puts it into a dialog for use as a means of a picking a file and attaching it to an Inventory record.


mKowalski256 (approved):
Looks good and works for me on localhost and [AWS build](https://rsdev-174-front-and-back-14.researchspace.com/).

Backend changes are written to additionally support Gallery files attached to sample's field of 'attachment' type, which doesn't seem to be supported by front-end yet, but that can be added later.

PR #130, opened 2024-09-24
Send notification to user when async export fails

rs-fraser:
When an export e.g. export all users work fails, the user wasn't being notified of the failure. 

I wrapped the export generation attempt in a try-catch, and when there is an exception it gets logged, and an export failure notification is generated. Other types of export get handled in this way too.

The failure case can be recreated by setting `src/main/resources/deployments/defaultDeployment.properties` property `archive.folder.location` to an non-writable location.


mKowalski256 (commented):
Hmm, the code looks good, but doesn't seem to solve the issue on [a test AWS build](https://rsdev-145-notify-on-export-failure-3.researchspace.com). i.e. after setting `archive.folder.location` to unwritable location I still don't get notification after export attempt.

rs-fraser:
@mKowalski256 this is ready for review again, and now covers each type of export: all users work, selected documents, and all group work.

The exceptions get caught in each case, the failure notification is generated, then the exception is re-thrown as there is client code in the controller which performs some further checks based on the type of the exception.

mKowalski256 (approved):
Looks good - after setting unwrittable archive dir location I got expected notification for every export scenario I tried.

PR #132, opened 2024-09-25
PRT-805: Make sure we hash the apiKey with `SHA256` (instead of `BCrypt`)

rs-nicof:
The `BCrypt` is not ideal to be used crypting apiKeys since it is not indexable on the database column. So in order to retrieve the user we were scanning all the list of users (O(n) complexity) instead of querying an indexed column of the database that would have had a constant complexity.


mKowalski256:
Please also modify the `rsdev-310-perform-hash-apiKeys.xml` - it has a comment mentioning bcrypt, also `id` of changeset should be changed so it executes again on pangolin. While on it, add `context="run"` to the changeset, we only want it to run once, and without that context it'd run on every restart.

mKowalski256 (approved):
Looks good and works fine for me on test AWS build.

Don't forget to merge core-model first, as version `2.2.3`, then update core-model reference in pom.xml here.

rs-fraser (approved):

rs-nicof:
✅ Successfully tested on AWS ✅ 


Uploading Screen sharing - 2024-09-25 1_21_41 PM.mp4…

rs-nicof:
✅  Successfully tested in AWS ✅ 


https://github.com/user-attachments/assets/3979b3d7-da01-4845-8e00-537a3dba8616

PR #142, opened 2024-10-04
Glossary

rlamacraft:
Here's a glossary of terms that I often use in code documentation, as well as various terms that we use in and around the project that would be helpful for a new contributor. Additions/alterations welcome!


tilorspace (approved):
looks good to me. I'll keep an eye out for further additions.

PR #148, opened 2024-10-11
RSDEV-284: support for OpenID Connect as SSO provider

mKowalski256:
This PR implements [RSDEV-284](https://researchspace.atlassian.net/browse/RSDEV-284) ticket, and it can be functionally tested at https://openid-connect.researchspace.com (maybe contact me/Ramon before trying, as we're testing various configurations right now).

Code-wise the new functionality is implemented through new `OpenIdRemoteUserPolicy.java` class, but I also did some refactoring around classes implementing `RemoteUserRetrievalPolicy`, as the code was quite tied to SAML (shibboleth) integration, which was the main SSO system we've been using until now.

### DIFF
--- a/src/main/java/com/axiope/service/cfg/ProductionConfig.java
+++ b/src/main/java/com/axiope/service/cfg/ProductionConfig.java
@@ -33,6 +33,7 @@ import com.researchspace.service.impl.license.NoCheckLicenseService;
 import com.researchspace.spring.taskexecutors.ShiroThreadBindingSubjectThreadPoolExecutor;
 import com.researchspace.webapp.filter.EASERemoteUserPolicy;
 import com.researchspace.webapp.filter.MockRemoteUserPolicy;
+import com.researchspace.webapp.filter.OpenIdRemoteUserPolicy;
 import com.researchspace.webapp.filter.RemoteUserRetrievalPolicy;
 import com.researchspace.webapp.filter.SAMLRemoteUserPolicy;
 import java.nio.file.Files;
@@ -282,6 +283,8 @@ public class ProductionConfig extends BaseConfig {
   protected RemoteUserRetrievalPolicy remoteUserRetrievalPolicy() {
     if ("SAML".equals(deploymentSsoType)) {
       return new SAMLRemoteUserPolicy();
+    } else if ("openid".equals(deploymentSsoType)) {
+      return new OpenIdRemoteUserPolicy();
     } else if ("TEST".equals(deploymentSsoType)) {
       return new MockRemoteUserPolicy();
     } else {
--- a/src/main/java/com/axiope/service/cfg/ProductionConfig.java
+++ b/src/main/java/com/axiope/service/cfg/ProductionConfig.java
@@ -281,11 +281,11 @@ public class ProductionConfig extends BaseConfig {
 
   @Bean
   protected RemoteUserRetrievalPolicy remoteUserRetrievalPolicy() {
-    if ("SAML".equals(deploymentSsoType)) {
+    if ("SAML".equalsIgnoreCase(deploymentSsoType)) {
       return new SAMLRemoteUserPolicy();
-    } else if ("openid".equals(deploymentSsoType)) {
+    } else if ("openid".equalsIgnoreCase(deploymentSsoType)) {
       return new OpenIdRemoteUserPolicy();
-    } else if ("TEST".equals(deploymentSsoType)) {
+    } else if ("TEST".equalsIgnoreCase(deploymentSsoType)) {
       return new MockRemoteUserPolicy();
     } else {
       return new EASERemoteUserPolicy();
--- a/src/main/java/com/researchspace/webapp/filter/OpenIdRemoteUserPolicy.java
+++ b/src/main/java/com/researchspace/webapp/filter/OpenIdRemoteUserPolicy.java
@@ -1,0 +1,102 @@
+package com.researchspace.webapp.filter;
+
+import com.researchspace.core.util.CryptoUtils;
+import java.util.Map;
+import java.util.TreeMap;
+import javax.servlet.http.HttpServletRequest;
+import lombok.AccessLevel;
+import lombok.Setter;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.beans.factory.annotation.Value;
+
+/** Translates OpenId claims form httpRequest headers into user information */
+@Slf4j
+@Setter(AccessLevel.PACKAGE)
+public class OpenIdRemoteUserPolicy extends AbstractSsoRemoteUserPolicy {
+
+  private static final String ADDITIONAL_CLAIMS_SEPARATOR = ".";
+
+  @Value("${deployment.sso.openid.usernameClaim}")
+  private String usernameClaim;
+
+  @Value("${deployment.sso.openid.additionalUsernameClaim}")
+  private String additionalUsernameClaim;
+
+  @Value("${deployment.sso.openid.additionalHashedUsernameClaim}")
+  private String additionalHashedUsernameClaim;
+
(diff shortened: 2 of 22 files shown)


src/main/java/com/axiope/service/cfg/ProductionConfig.java
```diff
@@ -282,6 +283,8 @@ public TimeLimitedMemoryCache<License> licenseCache() {
   protected RemoteUserRetrievalPolicy remoteUserRetrievalPolicy() {
     if ("SAML".equals(deploymentSsoType)) {
       return new SAMLRemoteUserPolicy();
+    } else if ("openid".equals(deploymentSsoType)) {
```
rs-nicof, line 286: can we use `OPENID` capitalised as per `SAML` and `TEST`?

src/main/java/com/researchspace/webapp/filter/OpenIdRemoteUserPolicy.java
```diff
@@ -0,0 +1,102 @@
...
+import javax.servlet.http.HttpServletRequest;
+import lombok.AccessLevel;
+import lombok.Setter;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.beans.factory.annotation.Value;
+
+/** Translates OpenId claims form httpRequest headers into user information */
+@Slf4j
+@Setter(AccessLevel.PACKAGE)
```
rs-nicof, line 15: would it make sense to be PROTECTED in case someone extends it?

src/main/java/com/researchspace/webapp/filter/SAMLRemoteUserPolicy.java
```diff
@@ -40,71 +34,57 @@ public class SAMLRemoteUserPolicy implements RemoteUserRetrievalPolicy {
...
-    log.debug("Req headers: {}", logHeaders(httpRequest).toString());
-    log.debug("Environment variables: {}", logEnv().toString());
-    log.debug("Attributes: {}", logReqAttributes(httpRequest).toString());
-    log.debug("SAML attributes: {}", logSamlAttributes(httpRequest).toString());
-    log.debug("Remote user:  {}", httpRequest.getRemoteUser());
+    SECURITY_LOG.debug("Req headers: {}", logHeaders(httpRequest));
+    SECURITY_LOG.debug("Environment variables: {}", logEnv());
+    SECURITY_LOG.debug("Attributes: {}", logReqAttributes(httpRequest));
+    SECURITY_LOG.debug("SAML attributes: {}", logSamlAttributes(httpRequest));
+    SECURITY_LOG.debug("Remote user:  {}", httpRequest.getRemoteUser());
```
rs-nicof, line 41: Do we still need all those debug lines? Are they possibly printing PI data?

src/main/java/com/researchspace/webapp/filter/SAMLRemoteUserPolicy.java
```diff
@@ -1,24 +1,18 @@
...
   protected static final String SHIBBOLETH_ID_ATTRIBUTE = "eduPersonPrincipalName";
 
-  private static final Logger log = LoggerFactory.getLogger(SecurityLogger.class);
-
-  // don't alter, edit or remove these as they may be used in production deployments
-  // only add new values if needed.
-  static final String[] shib_attributes = {
+  /* don't alter, edit or remove these as they may be used in production deployments;
+  only add new values if needed. */
+  private static final String[] shib_attributes = {
```
rs-nicof, line 15: We could add a test to make sure that the number of values and the values themselves are not mistakely changed

src/test/java/com/researchspace/webapp/filter/OpenIdRemoteUserPolicyTest.java
```diff
@@ -0,0 +1,77 @@
...
+    assertEquals(3, openIdPolicy.getOtherRemoteAttributes(req).size());
+    assertEquals(
+        "someone@somewhere.com",
+        openIdPolicy.getOtherRemoteAttributes(req).get(RemoteUserAttribute.EMAIL));
+    assertEquals(
+        "Mark", openIdPolicy.getOtherRemoteAttributes(req).get(RemoteUserAttribute.FIRST_NAME));
+    assertEquals(
+        "Smith", openIdPolicy.getOtherRemoteAttributes(req).get(RemoteUserAttribute.LAST_NAME));
+  }
+}
```
rs-nicof, line 77: <maybe we should check the order if it is important? so that we can avoid the TreeMap is changed without the tests fail

src/main/java/com/axiope/service/cfg/ProductionConfig.java
mKowalski256: `SAML` is capitalized as it's actually an acronym, `openid` doesn't have to be. But yeah I can switch to `equalsIgnoreCase`.

src/main/java/com/researchspace/webapp/filter/OpenIdRemoteUserPolicy.java
mKowalski256: ok.

src/main/java/com/researchspace/webapp/filter/SAMLRemoteUserPolicy.java
mKowalski256: Yes these are occasionally useful and are not printed by default.

src/main/java/com/researchspace/webapp/filter/SAMLRemoteUserPolicy.java
```diff
@@ -40,71 +34,57 @@ public class SAMLRemoteUserPolicy implements RemoteUserRetrievalPolicy {
...
-    StringBuilder sb = new StringBuilder();
-    while (attrNames.hasMoreElements()) {
-      String attr = attrNames.nextElement();
-      sb.append(attr + ":" + httpRequest.getAttribute(attr).toString() + ", ");
-    }
-    return sb;
-  }
+  @Override
+  public Map<RemoteUserAttribute, String> getOtherRemoteAttributes(HttpServletRequest httpRequest) {
+    Map<RemoteUserAttribute, String> rc = new TreeMap<>();
```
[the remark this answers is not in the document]
mKowalski256: No, I think I just copied `TreeMap` from `SAMLRemoteUserPolicy`, but any `Map` would do.

src/main/java/com/researchspace/webapp/filter/SAMLRemoteUserPolicy.java
mKowalski256: Yes good point, but let's maybe leave it for next person changing that SAML attributes list ;) 
(I was only reformatting the comment with my commit here).

rs-fraser (commented):
Code looks good.

Tested various logins and logouts on https://openid-connect.researchspace.com/ and signing up new users from Okta.

Only comment I have is that after clicking 'Sign out' on rspace the user is still logged in on page refresh. 

image

We show this message. Can we direct the user back to the SSO provider to fully logout? Or somehow invalidate their login on the rspace side?

mKowalski256:
> We show this message. Can we direct the user back to the SSO provider to fully logout? Or somehow invalidate their login on the rspace side?

Yeah, logging out of external SSO is a bit problematic as often there is no specific logout, and also user may not necessarily want to complete their global SSO session just because they finished working with RSpace. But now when thinking about it, in many cases we could provide a direct SSO logout link on RSpace logout page, so while it doesn't happen automatically, it is just one click away. I'll add it now.

mKowalski256:
Conditional messages on SSO logout page:
1. if logout link for SSO IDP **is not** configured:
image
2. if logout link for SSO ID **is** configured:
image

rs-fraser (approved):

PR #152, opened 2024-10-14
RSDEV-345 Change TinyMCE in Inventory to not be floating

rlamacraft:


mKowalski256 (approved):

PR #163, opened 2024-10-23
Fix for ldapsearch command (which RSpace calls on LDAP setup) no longer working on ubuntu22.04 and newer

tuxRamon:
In previous versions of Ubuntu the ldapsearch command required the -h flag to specify the host. However on ubuntu 22.04 and newer the -h flag breaks the command and you can just specify the IP with the -h flag and it works.

As a result, this breaks RSpace from being able to be setup with LDAP SSO when running on Ubuntu 22.04

I've edited the Java code to remove the -h flag when it calls the ldapsearch command.

This would fix LDAP for ubuntu 22.04 servers. 

Just creating this as a draft request at the moment, because I still need to check some things over from the DevOps side. When I have I'll change to a full pull request

### DIFF
--- a/src/main/java/com/researchspace/ldap/impl/LdapSearchCmdLineExecutor.java
+++ b/src/main/java/com/researchspace/ldap/impl/LdapSearchCmdLineExecutor.java
@@ -17,7 +17,7 @@ public class LdapSearchCmdLineExecutor {
 
   public LdapSearchCmdLineExecutor(String ldapBaseSuffix, String ldapUrl, String dnField) {
     searchBase = ldapBaseSuffix;
-    host = ldapUrl.split("//")[1];
+    host = ldapUrl;
     this.dnField = dnField;
   }
 
@@ -29,8 +29,7 @@ public class LdapSearchCmdLineExecutor {
    */
   public String findDnForUid(String ldapUsername) {
 
-    String command =
-        String.format("ldapsearch -x %s -b \"%s\" uid=%s", host, searchBase, ldapUsername);
+    String command = getLdapSearchCommandString(ldapUsername);
     log.debug("running: " + command);
 
     String foundDn = null;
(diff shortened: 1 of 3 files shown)


tuxRamon:
ldapsearch command without the -h flag, does also work on older versions of ubuntu, so we wouldn't be breaking backwards compatibility with ubuntu 20.04 for example.

mKowalski256:
I've manage to test it on p8088, but the new format of command doesn't seem to work with RSpace, nor actually when run on pangolin from command line. Does it need a different format apart from removing `-h` flag?

Previous command, which works fine on p8088, was:
`ldapsearch -x -h howler.researchspace.com -b "dc=test,dc=howler,dc=researchspace,dc=com" uid=ldapUser1`

New command no longer works on p8088, and looks like this:
`ldapsearch -x howler.researchspace.com -b "dc=test,dc=howler,dc=researchspace,dc=com" uid=ldapUser1`

And trying to run on bash, I get:
```
> ldapsearch -x howler.researchspace.com -b "dc=test,dc=howler,dc=researchspace,dc=com" uid=ldapUser1
ldap_sasl_bind(SIMPLE): Can't contact LDAP server (-1)
```

peko-hki:
Did you configured /etc/ldap/ldap.conf ?
If it is setup correctly ldapsearch should work (in our environment it is working ; ubuntu 22.04 LTS)

root@elab:/var/lib/tomcat9/rspace# ldapsearch -x 10.XX.XX.XX -b "o=YY" uid=pjkoch | grep ^dn: | grep -i pjk
dn: cn=PJKoch,ou=ADM,o=YY

tuxRamon:
The new command works on howler, I think it might be the way communication is between servers. Will take a look

mKowalski256:
Is it that deprecated `-h` flag was removed in ubuntu 22 and later, but the `-H` which was not deprecated is still working? If so, the command with -H flag does work for me on pangolin, although requires prefixing hostname with `ldap://`.

```
> ldapsearch -x -H ldap://howler.researchspace.com -b "dc=test,dc=howler,dc=researchspace,dc=com" uid=ldapUser1
# extended LDIF
```

If that solution works for both old and new version the PR could just change flag to `-H` and prefix the url with `ldap://`.

peko-hki:
Here it is working that way (ubuntu 18 & 22 LTS). So changing it to "-H ldap://$HOSTNAME" would be a solution!

mKowalski256 (approved):
Tested new build internally, can connect to test howler ldap fine.

PR #164, opened 2024-10-28
RSDEV-373 New Gallery pagination

rlamacraft:
Adds a "Load more" button to grid and tree view of the new Gallery. Crucially, this change also includes a new search parameter for the `getUploadedFiles` endpoint that allows for filtering the results based on whether they are folders for populating the move dialog.


mKowalski256:
Checked at https://rsdev-373-paginating-gallery-11.researchspace.com. 

The `load more` button seems to work fine, but I see the following problem: after using the button, then navigating away and back to the folder, the button seems to remember it was clicked, but the gallery listing only includes initial page of items. That means I can no longer 'load more' items even though I only see the first page.

rlamacraft:
> after using the button, then navigating away and back to the folder, the button seems to remember it was clicked, but the gallery listing only includes initial page of items. That means I can no longer 'load more' items even though I only see the first page.

Good catch, should be fixed now

mKowalski256:
Also, there seems to be git conficts with latest `main` branch.

rlamacraft:
> Also, there seems to be git conficts with latest `main` branch.

Resolved now

mKowalski256 (approved):
Tested at https://rsdev-373-paginating-gallery-14.researchspace.com/, looks good to me!

PR #167, opened 2024-10-28
RSDEV-314 Create subsamples from existing samples

rlamacraft:


mKowalski256:
Did some testing on https://rsdev-314-final-6.researchspace.com, looks good to me.

I noted just one issue - it seems that 'create subsample' option is enabled even if user has no permission to edit the sample. In this  case the back-end won't allow subsample creation, and I think the option should be disabled in the menu.

See the recording reproducing the issue:
[RSpace-Inventory (2).webm](https://github.com/user-attachments/assets/bbefac3a-cf3a-401c-ac39-40c23f6d7c91)

mKowalski256 (approved):

PR #173, opened 2024-10-31
RSDEV-346: fix repoCfg object passed on DCD export

mKowalski256:
Quick fix for an export to DCD issue found by selenium test, `repoCfg` was no longer initialized correctly as the label has changed, and the code was conditional on the label text.


mKowalski256:
I deployed the PR to p8085 just now, and it seems to fix the problem with DCD export.
@rlamacraft is it fine to merge as a quick fix?

rlamacraft:
> I deployed the PR to p8085 just now, and it seems to fix the problem with DCD export. @rlamacraft is it fine to merge as a quick fix?

yeah, not sure why that the code is using the display name though

PR #174, opened 2024-11-01
Refactor setup of repoCfg

rlamacraft:
Most importantly this replaces the use of `displayName` with `repoName`, which is what the logic of the code should exclusively rely on. Nobody would expect that changing the `displayName` would have a functional change to the logic.

Secondly, this change makes is explicit that it is dataverse that is doing the exceptional thing by supporting multiple destinations and not all of the other repository services by inverting the conditional logic and adding explanatory comments.


mKowalski256 (approved):
Code looks good, although I couldn't really retest on AWS, we can check on pangolins after merge.

PR #179, opened 2024-11-05
RSDEV-352: remove old baseline db script

mKowalski256:
The `ecat5dev-baseline2014-1-24.sql` script was effectively replaced by `rs-dbbaseline-utf8.sql`, and after looking into changes between the two they are basically the same, apart from charset encoding for which I believe we want utf8 as default anyway (that matches prod db setup).


tuxRamon:
Looks good. We've been using `rs-dbbaseline-utf8.sql` for a while now (it's what's been included with our packages the last few years and also forms the base of the db-template.sql file for the open source docker rspace).

I don't see any problem removing `ecat5dev-baseline2014-1-24.sql` and only keeping `rs-dbbaseline-utf8.sql`

PR #182, opened 2024-11-08
RSDEV-352: maven 'environment=drop-recreate-db' param now builds local db from baseline sql

mKowalski256:
The `environment=drop-recreate-db` maven param, used by devs to setup localhost environment, was previously triggering re-creation of local `rspace` database out of hibernate/JPA mappings, then during RSpace server startup the liquibase changesets marked with `dev-test` context were executed. The main change of this PR is that  `environment=drop-recreate-db` now re-creates local database out of baseline sql script, upon which the liquibase updates from context="run" are applied too. As a result database schema created by maven is much closer to schema of production database.

When running code in this branch the `mvn` command should no longer specify `liquibase.context` param, i.e. `-Dliquibase.context=dev-test` should not be used. Correct context for localhost dev setup is now `run,dev-test`, but that doesn't need to be explicitly passed to `mvn` command, as is picked up through `deployment.properties` or specified in `pom.xml`. Keeping the `-Dliquibase.context=dev-test` param will cause db setup script to fail on a first changset that assumes changes from `run` context are present, with an error like: `MySQLSyntaxErrorException: Unknown column 'tempAccount' in 'field list'`. 

The main visible change when starting up a localhost server with code from this branch is that logs will show much more entries about changests being applied. Also the database schema will have some differences, even some orphaned tables, e.g. in hibernate-generated schema there are unused `customercontact/customerinfo` tables, while in liquibase-generated schema there is an orphaned `br_aud_temp` table.

Other changes in this PR:
- old way of creating local database is still available through `environment=drop-recreate-hibernate-db` maven param, in case it's needed (e.g. for `generateDiff` task); `drop-recreate-hibernate-db` option operates on a database named `rspace_hib` (needs to be created manually), rather than on main `rspace` database
- "testLiquibase" maven profile is removed, as it's now obsolete; it used to create baseline-liquibase db schema for comparision, but now it's main drop-recreate-db profile that does that; the `generateDiff` maven task is now 
- the extra `changeLog-rsdev-351.xml` file applies a few schema changes that were always present in hibernate-created db, but not in liquibase-created one
- documentation updates


src/main/resources/sqlUpdates/DatabaseChangeGuidelines.md
```diff
@@ -39,13 +42,13 @@ There are various naming conventions, please [IntegrationAndAppNotes](integratio
...
 - when adding/modifying audit table columns, remember that they don't generally have not-null constraints (apart from id/REV columns).
--  If you are adding such constraint make sure you know the consequences, and double check if constraint is actually applied on your local database when testing.
+- If you are adding such constraint make sure you know the consequences, and double check if constraint is actually applied on your local database when testing.
 
 ### Have you set a context? Values are:
 - no value. Will run on all databases in all environments. E.g. static  reference data
 - Suitable for adding lookup static data to all test/dev/production databases. E.g., permissions.
   - `run` for general schema changes - adding/removing columns/ tables /keys
-  - `dev-test` change should only be applied during unit test runs - e.g., loading up test data.
+  - `dev-test` change should only be applied on local dev environment, or during unit test runs - e.g., loading up test data.
```
rs-fraser, line 51: So going forward, the vast majority of changes will only have the `run` context as it runs all of the time, as oppose to now where many have both `run,dev-test`? And `dev-test` will only be for the specific case you mention here?

rs-fraser (approved):
Looks good. 

I removed the `liquibase.context` param and could run rspace as normal. 

I was also able to generate the diff file. Should we create a ticket to look into these differences in case any are important?

src/main/resources/sqlUpdates/DatabaseChangeGuidelines.md
```diff
@@ -154,28 +117,46 @@ E.g.,
...
 
-This will attempt to apply new liquibase changes to the testLiquibaseUpdate and run JUnit tests.
-All the updates are applied via Spring at startup - since all database access in both tests and the application is through Spring beans, this should work fine.
+You can compare manually, or run another maven command:
+
+    mvn -e process-resources -Denvironment=keepdbintact -PgenerateDiff
 
-4. After they pass, it's OK to commit. If the tests fail, read the maven output to see what went wrong, edit the liquibase changesets that didn't run and try again. Don't edit changesets that completed successfully - Liquibase might try to run them again. 
+which is based on `liquibase-maven-plugin` and will compare your `rspace` database with `rspace_hib`
+database, and generate a time-stamped changeset file in folder *generatedDBChangeSets/* in your
```
rs-fraser, line 150: On linux I had to create `generatedDBChangeSets` before running the command otherwise it failed, which will probably be the case for others. 

Could be worth mentioning here, or just create the file in the root of the repo.

src/main/resources/sqlUpdates/DatabaseChangeGuidelines.md
mKowalski256: yes, going forward we shouldn't need combined `run,dev-test` context at all (i.e. can be replaced with just `run`), and only occasional `dev-test` for change that should be run for unit tests or localhost.

mKowalski256:
> I was also able to generate the diff file. Should we create a ticket to look into these differences in case any are important?

There is a pre-existing [RSDEV-351](https://researchspace.atlassian.net/browse/RSDEV-351) ticket for that already, although I think it'll be less urgent now as we will naturally stop accumulating these differences. In the past the differences were increasing because when someone made a change on model level and saw it on localhost db, it was easy to miss to put them into liquibase.

I've also raised a separate issue [RSDEV-411](https://researchspace.atlassian.net/browse/RSDEV-411) for clearing the changelogs from obsolete changesets/contexts, as discussed in other comment.

PR #184, opened 2024-11-12
RSDEV-411: a few fixes for pdf export, most importantly allow docs to have & (ampersand) in the name

mKowalski256:
Turns out that introducing a new pdf-printing library (`flying-saucer-pdf`) caused a problem with exporting some ELN documents to PDF. The problem was caused by & (ampersand) character that RSpace may try to put into PDF page header (or footer) as a part of document name, or (less likely) as a part of full name of the exporting user. Attempt to export a document that has `&` in the name was resulting in an error like:
```
Can't load the XML resource (using TrAX transformer). org.xml.sax.SAXParseException; lineNumber: 19; 
columnNumber: 67; The reference to entity "in..." must end with the ';' delimiter..
```

After investigating, it seems that `flying-saucer-pdf` has a problem with interpreting html entities/chars passed through CSS 'content' attribute, which was how the header/footer generation was done. The fix applied in this PR is to use [running CSS headers/footers](https://printcss.net/articles/running-headers-and-footers), which allows to define content of header/footer through html fragment in '<body>' section, rather than CSS attribute in '<style>' section. When html entities/chars are set in body, the lib can print them to pdf fine. 

This PR also fixes a couple of other PDF export issues found during the investigation:
- when document name contained non-ascii symbols, these symbols were generally not printed to PDF in a document name that is put into page header (they were printed fine if inside part of document content); this was because non-ascii symbols require defining a different font family, which was not done for header part
- the PDF export dialog toggle 'Insert date footer at file end only' was not working properly, i.e. the PDF output always had the footer at file end only; this was due to front-end passing the parameter under a different name than expected by back-end (`includeFooter` rather than `includeFooterAtEndOnly`)

--- 

### Test scenario 

Test AWS build running at [https://rsdev-411-pdf-special-chars-5.researchspace.com](https://rsdev-411-pdf-special-chars-5.researchspace.com), compare to any 1.104 build (e.g. pangolin).

1. Go to My RSpace -> My Profile, update user's first/last name to one with non-ascii chars, e.g. `Mæck Schødt-Żębski`
2. Create a new basic document named with ampersand and non-ascii chars, e.g. `my special chars &™∞§`; put non-ascii content into text field e.g. `my special content &™∞§`. Save the doc;
3. Try exporting doc to PDF, with `Insert date footer at file end only` toggle disabled in pdf dialog
- on 1.104 server the export will fail - notification count will be up by 1, but there will be no notification
- on AWS build with the fix, the PDF will generate fine
4. On 1.104 server, rename the doc to not have `&` in the name, then try exporting again, with  `Insert date footer at file end only` toggle disabled
- the PDF will generate fine, but compare to PDF generated from AWS instance build with the fix, and note: 1) not all non-ascii chars printed in header 2) no footer on every page

### DIFF
--- a/src/main/java/com/researchspace/export/pdf/ExportToFileConfig.java
+++ b/src/main/java/com/researchspace/export/pdf/ExportToFileConfig.java
@@ -112,7 +112,7 @@ public class ExportToFileConfig implements StructuredDocumentHTMLViewConfig, IEx
         + pageSize
         + ", dateType="
         + dateType
-        + ", includeFooter="
+        + ", includeFooterAtEndOnly="
         + includeFooterAtEndOnly
         + ", exportName="
         + exportName
--- a/src/main/java/com/researchspace/export/pdf/PdfHtmlGenerator.java
+++ b/src/main/java/com/researchspace/export/pdf/PdfHtmlGenerator.java
@@ -36,30 +36,33 @@ public class PdfHtmlGenerator {
 
   private final VelocityEngine velocityEngine;
 
+  private HTMLUnicodeFontProcesser htmlUnicodeFontProcesser;
+
   public static final int MAX_TITLE_WIDTH = 50;
 
   public final SimpleDateFormat simpleDateFmt = new SimpleDateFormat("yyyy-MM-dd");
 
   @Autowired
-  public PdfHtmlGenerator(VelocityEngine velocityEngine) {
+  public PdfHtmlGenerator(VelocityEngine velocityEngine, HTMLUnicodeFontProcesser fontProcessor) {
     this.velocityEngine = velocityEngine;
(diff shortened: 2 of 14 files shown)


mKowalski256:
@rs-fraser I've added you as a reviewer, as you're most familiar with pdf-generation code, hopefully all looks good!
@rlamacraft just one change in front-end here, renaming param that is passed to server with PDF export dialog

rlamacraft (approved):
Seems fine; if flow, jest, and the e2e tests are okay with it then the variable name change must be fine

rs-fraser (approved):
Looks good. Followed the testing scenarios and can see both the issues with special characters and the footer issue on an older build. On the AWS feature branch I can see these issue fixed and the pdf is generated correctly.

PR #186, opened 2024-11-13
RSDEV-418: more robust notifications for pdf export errors

mKowalski256:
In current codebase the PDF export fails for documents that have html entity hash code in their name (e.g. a doc named `test &#x3 doc`), but such error doesn't produce any notification, just bumps the notification count.

The main change of this PR is that async PDF exports now calls `postArchiveExportFailure()` in `catch` blocks on `ExportImportImpl` class level, which is the same pattern as used in async archive exports, and that pattern does result in error notification delivered to the user. The call to produce error notification in `PdfWordExportManagerImpl.processDocsForDocExport()`, which was resulting only in bumped notification count, is removed.

As a bit of refactoring, I also renamed most of the methods in `ExportImport` interface so they follow the same naming pattern. 

### Test scenario

Compare behaviour of three problematic export PDF scenarios executed on [AWS build](https://rsdev-418-pdf-export-error-handling-5.researchspace.com/) vs 1.104.x server (or pangolins).

1. login to the account of user in a group (e.g. `user3c` on AWS build).
2. create a new ELN document named `test &#x3 doc`.
3. try exporting the document to PDF
- on AWS build you should get an error notification
- on 1.104.x builds (or pangolins) you should see just a bumped notification count, but no notification
4. try exporting whole user's account to PDF (`My RSpace -> Export/Import -> Export all my work`), and whole group content to PDF (`My RSpace -> My LabGroups -> Export LabGroup's Work`)
- again, on AWS build you should get error notification
- on 1.104.x builds (or pangolins) there should be no notification

### DIFF
--- a/src/main/java/com/researchspace/api/v1/service/impl/ExportTasklet.java
+++ b/src/main/java/com/researchspace/api/v1/service/impl/ExportTasklet.java
@@ -87,7 +87,7 @@ public class ExportTasklet implements Tasklet {
       }
       try {
         ArchiveResult result =
-            exportService.exportArchiveSyncUserWork(
+            exportService.syncExportUserWorkToArchive(
                 config, toExport, serverUrl(), user, postArchiveCompleter, exportIdSupplier);
 
         updateContext(chunkContext, result);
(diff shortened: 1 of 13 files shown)


rs-fraser (approved):
Looks good.

Tested the AWS build and can see notifications are now properly generated when the PDF export fails.

PR #198, opened 2024-12-02
RSDEV-262: remove remnants of mendeley-linking integration

mKowalski256:
Remove front-end & back-end code related to old mendeley integration that allowed browsing mendeley account and inserting links into ELN documents, through tinyMCE plugin.


mKowalski256:
@rlamacraft I marked you as a reviewer but just for 5 files from `src/main/webapp/ui/src` that I've modified/removed as a part of this PR, could you check them? I didn't look into details of how this code was fitting the apps page, and was it really safe to remove it.

PR #205, opened 2024-12-12
RSDEV-367: fix jove and youtube embeds

mKowalski256:
When saving content of tinymce field to the back-end, the html is sanitized and potentially unsafe tags/attributes are removed. In particular we do remove `iframe` tags by default, but make exception for [youtube](https://documentation.researchspace.com/article/cdzdub1ykw-embedding-video-content-from-youtube) & [jove](https://documentation.researchspace.com/article/mopbqzzdf5-setting-up-jove-integration), as these are supported integrations.

The code in the PR adjusts whitelist of iframe `src` and other iframe attributes, so the latest embed fragments generated by Youtube & Jove are accepted.

Note that Jove integration doesn't work (Jove app on Apps page is scheduled for removal, see RSDEV-335), but we do want to allow directly embedded fragments to be pasted into RSpace tinyMCE.

### DIFF
--- a/src/main/java/com/researchspace/linkedelements/TextFieldEmbedIframeHandler.java
+++ b/src/main/java/com/researchspace/linkedelements/TextFieldEmbedIframeHandler.java
@@ -26,7 +26,7 @@ public class TextFieldEmbedIframeHandler {
 
   // youtube src url last fragment, e.g. 'YkRldqVfTJo?start=17'
   private static final String YOUTUBE_SRC_SUFFIX_PATTERN =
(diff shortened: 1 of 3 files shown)


mKowalski256:
AWS build at https://rsdev-367-fix-jove-youtube-embed-5.researchspace.com. 
Example embed fragments in new format are in summary of [RSDEV-367](https://researchspace.atlassian.net/browse/RSDEV-367).

rs-nicof (approved):

PR #209, opened 2024-12-18
RSDEV-294: public API for accessing Gallery Filestores

mKowalski256:
The PR adds multiple endpoints to ELN API in Gallery Filestores section, this is to expose all the operations available in the UI of the old Filestores subgallery.

image

The usage workflow is similar to one in old UI. Filesystems are defined/activated by the System Admin (through the UI on System page, no public API yet). To browse the content on external filesystem (or filestore saved on that filesystem) user needs to be connected to that filesystem (`/filesystem/login` endpoint). User can save the remote filesystem path as their named filestore, and later browse/download/link to resources on that filestore path.

Summary of changes in this PR:
* new `GalleryFilestoresApiController` handles `/gallery/filestores` endpoints that manage standard gallery filestores actions
* `GalleryApiController` renamed to `GalleryIrodsApiController`, as that controller is handling `/gallery/irods` endpoints, which are more specialised filestore actions (limited to iRODS filestores)
* encapsulated filestore authentication/caching mechanism from `GalleryIrodsApiController` into separate `GalleryFilestoresCredentialsStore` class, so that code can be reused in `GalleryFilestoresApiController`
* extended filestore model used by previous UI (e.g. `NfsFileTreeNode` class), so the old jsp UI still works, but public API can return more useful properties (e.g. file-size in bytes, rather than as readable string)
* guarding both `/gallery/filestores` and `gallery/irods` endpoints with pre-existing deployment property `netfilestores.enabled`, which was always used block filestores access in the UI. If the property is set to `false`, only sysadmin account can use filesystem/filestores API (that's useful for stage of sysadmin setup), while a regular users get an error message


src/main/java/com/researchspace/api/v1/controller/GalleryFilestoresApiController.java
```diff
@@ -0,0 +1,217 @@
...
+  public void downloadFromFilestore(
+      @PathVariable Long filestoreId,
+      @RequestParam(value = "remotePath", required = false) String remotePath,
+      @RequestParam(name = "remoteId", required = false) Long remoteId,
+      @RequestAttribute(name = "user") User user,
+      HttpServletResponse response)
+      throws IOException {
+
+    assertFilestoresApiEnabled(user);
+    if (StringUtils.isEmpty(remotePath) && remoteId == null) {
```
rs-nicof, line 97: if you use `isBlank` you can check also if there are only special character like TAB of NewLine

src/main/java/com/researchspace/api/v1/controller/GalleryFilestoresApiController.java
```diff
@@ -0,0 +1,217 @@
...
+    File downloadedFile = nfsFileDetails.getLocalFile();
+    log.info("downloaded to: " + downloadedFile.getCanonicalPath());
+
+    response.setContentType("application/octet-stream");
+    response.setContentLength((int) downloadedFile.length());
+    response.setHeader(
+        "Content-Disposition", "attachment; filename=\"" + downloadedFile.getName() + "\"");
+    try (InputStream is = new FileInputStream(downloadedFile);
+        ServletOutputStream out = response.getOutputStream()) {
+      IOUtils.copy(is, out);
```
rs-nicof, line 117: Is the exception management (i.e.: disk full, file name already existing, or any other issue that can happen in a filesystem) taken care somewhere else?

src/main/java/com/researchspace/api/v1/controller/GalleryFilestoresApiController.java
```diff
@@ -0,0 +1,217 @@
...
+@NoArgsConstructor
+@ApiController
+public class GalleryFilestoresApiController extends GalleryFilestoresBaseApiController
+    implements GalleryFilestoresApi {
+
+  @Autowired private NfsManager nfsManager;
+  @Autowired private NfsFileHandler nfsFileHandler;
+
+  @Override
+  public List<NfsFileStoreInfo> getUserFilestores(@RequestAttribute(name = "user") User user) {
```
rs-nicof, line 50: if you return a `Set` then you can avoid to scan the list to verify if the name is already existing

src/main/java/com/researchspace/service/impl/NfsManagerImpl.java
```diff
@@ -52,15 +55,60 @@ public void saveNfsFileStore(NfsFileStore fileStore) {
...
   @Override
   public void markFileStoreAsDeleted(NfsFileStore fileStore) {
     fileStore.setDeleted(true);
     nfsDao.saveNfsFileStore(fileStore);
   }
 
   @Override
-  public List<NfsFileStore> getFileStoresForUser(Long userId) {
-    return nfsDao.getUserFileStores(userId);
+  public List<NfsFileStoreInfo> getFileStoreInfosForUser(User user) {
```
rs-nicof, line 88: You could return a `Set` here so it will be quicker to check if the file file store name is already exsisting in `verifyFileStoreNameUniqueForUser` 

But you need to change equals/hashcode on `NfsFileStoreInfo` to be based the `name` rather than `id`

rs-nicof (approved):
The PR looks good to me!
I've added few minor comments regarding changes that can be taken care or left as is.

mKowalski256:
Test AWS instance running at https://rsdev-294-filestores-api-24.researchspace.com/.

src/main/java/com/researchspace/api/v1/controller/GalleryFilestoresApiController.java
mKowalski256: The download code is actually part of pre-existing `NfsFileHandler` that was used in old UI, here the API code just assumes the file is there and streams it to servlet response, that should always be fine. 

Looking into implementation in `NfsFileHandler.downloadNfsFileToRSpace()` - it downloads file into a new temp folder each time, so that should cover some potential problems like name conflicts. It doesn't check the remaining free disk space though.

src/main/java/com/researchspace/service/impl/NfsManagerImpl.java
mKowalski256: Yeah the filestore name validation currently takes a bit of extra code, but it's old code so I'd leave it the way it is.

PR #222, opened 2025-01-03
RSDEV-445: improved share/unshare performance, especially when sharing with large group

mKowalski256:
The PR changes some internals of share/unshare action, to fix performance problem noticed when sharing with group of tens of users. The found bottleneck was in a fragment that was trying to establish into which shared folder the newly shared record should be placed. 

When sharing with a group, the old code was getting list of all members of the group and then was checking a) if the sharing user is allowed to share with every member of the group and b) what is the shared group folder of that group member. The loop there seems unnecessary, as a) when we establish that user is allowed to share with a group, we don't need to check it for every group member b) the same group's shared folder is referenced by individual shared folders of every group member, so it's enough to find one shared group folder.

Note - there is currently RSDEV-447 raised about a problem with sharing permissions, I'll tackle it separately. For this PR, let's just look into the rest of the logic.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1710,7 +1710,7 @@
     </dependency>
     <dependency>
         <groupId>com.github.rspace-os</groupId>
-        <artifactId>rspace-core-model</artifactId>
+        <artifactId>rspace-core-model-rsdev-441</artifactId>
         <version>${rspace-core-model.version}</version>
         <exclusions>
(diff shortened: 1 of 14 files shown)


rs-fraser (approved):
Changes look good.

I've test various sharing and un-sharing scenarios locally and didn't find any issues.

PR #224, opened 2025-01-07
RSDEV-447: additional checks before sharing

mKowalski256:
Previous checks in sharing code were not verifying whether sharing target (user or group that user tries to share with) is a valid target, but only whether the user has a permission to the record they try to share. Permissions to the record were actually checked three times - in the new code it's checked only in `assertCanDoSharingWithGroup/assertCanDoSharingWithUser` methods at the start of `doSharing()` method, and in top-level `shareRecord()` method.

The extra validation added by the PR ensures that, outside RSpace Community, user can only share with the groups they are member of, and with the users who are members of these groups.


rs-fraser (approved):
Code changes look good.

I tested the UI and API methods for sharing, and with these changes userA was correctly blocked from sharing with userB when they had no groups in common.  When they are members of the same group, sharing works as expected.

PR #236, opened 2025-01-20
RSDEV-455: GitHub integration now connects to default branch of the repo

mKowalski256:
* when browsing connected github repositories, query for name of default branch rather than assuming it's "master"
* minor refactoring & fixing scrollbar on "Github link" dialog


mKowalski256:
Test instance running at https://rsdev-455-github-fix-5.researchspace.com/.

rs-nicof (approved):
The code looks good to me. 

Is there any new test we can add? otherwise it is approved for me

mKowalski256:
> The code looks good to me.
> 
> Is there any new test we can add? otherwise it is approved for me

Thanks, yes you're right the test would be good - I've updated mocked mvcit one to cover the different default branches.

PR #241, opened 2025-01-27
RSDEV-297 Place current gallery folder id in URL

rlamacraft:
As the user navigates around the new gallery, this changes updates the URL so that it includes the id of the current folder. This does two things
1. It maintains backwards compatibility with all of the existing links to folders embedded in documents, keeping them all working once we swap `/oldGallery` to `/gallery`
2. It makes the browsers back and forward buttons useful for navigating around the Gallery, as well as allowing users to bookmark the current page and later jump back to a specific folder.


rlamacraft:
@mKowalski256 do you want to merge the backend changes separately, or shall we just merge this PR?

PR #245, opened 2025-01-28
RSDEV-458 UI work for supporting multiple PyRAT servers

rlamacraft:


rlamacraft:
With alias and time 
image

PR #251, opened 2025-01-29
RSDEV-451 Switch newGallery to gallery, and gallery to oldGallery

rlamacraft:


src/main/webapp/WEB-INF/decorators.xml
```diff
@@ -31,6 +31,9 @@
         <pattern>/newGallery</pattern>
         <pattern>/newGallery*</pattern>
         <pattern>/newGallery/**</pattern>
```
mKowalski256, line 33: should be fine to remove /newGallery patterns now, right?

mKowalski256 (approved):
Looks good & works fine for me at https://rsdev-451-repoint-gallery-10.researchspace.com/.

PR #253, opened 2025-01-31
RSDEV-347: various fixes around displaying document with historical attachments

mKowalski256:
This PR consists of multiple fixes for displaying documents that contain historical versions of attachments. Historical attachment versions cannot be added to documents in normal way, but will appear as a result of 'restoring' a historical version of a document. 

What is added by this PR:
- historical version of the attachment is now labelled with 'time' icon next to the attachment name
- after opening 'info' panel of the historical attachment, there is clear message explaining this is a historical attachment
- after switching to edit mode, the context menu is now correctly recognizing the historical attachment and links the proper view/download/info actions (previously it was assuming the attachment is at latest version)
- in document view mode, the UI now correctly handles scenario when multiple different versions of the same attachment are inserted in the document, and correctly updates the details in preview panel


mKowalski256:
Test scenario in [RSDEV-347](https://researchspace.atlassian.net/browse/RSDEV-347), test AWS build running at https://rsdev-347-historical-attachments-3.researchspace.com/.

src/main/java/com/researchspace/service/impl/BaseRecordManagerImpl.java
```diff
@@ -80,17 +80,17 @@ public BaseRecord load(Long id) {
   }
 
   @Override
-  public Map<Long, EcatMediaFile> retrieveMediaFiles(
+  public Map<String, EcatMediaFile> retrieveMediaFiles(
       User subject, Long[] mediaFileId, Long[] revisionId, DocumentInitializationPolicy policy) {
-    Map<Long, EcatMediaFile> mediaFiles = new HashMap<>();
+    Map<String, EcatMediaFile> mediaFiles = new HashMap<>();
```
rs-nicof, line 85: What do you think if you were using a structure like:
`Map<Long, Map<String, EcatMediaFile>>` that is `Map<id, Map<revision, file>>`?

mKowalski256: That's true, but this is a quick fix for a method that is only called by legacy javascript code, so I went with the least impactful change that'll get the work done...

PR #267, opened 2025-02-19
RSDEV-533 Refactor PyRAT tinymce plugin to avoid need for iframe

rlamacraft:
It seems excessive to render a react component inside of an iframe inside of a tinymce dialog, when we could just render the react component. Rendering the dialog directly also gives us full control over the styling, allowing us to using the accented theme based on each organisation's branding.


rlamacraft:
@mKowalski256 any thoughts? I feel like this simplifies things, especially with the ultimate goal of reactifying the document editor. It means that any new plugins going forward will be easier to migrate when the time comes. The content of each dialog could be implemented in a separate node module by external collaborators if they would prefer, with the plugin set up code trivial even if we had to do it every time.

rlamacraft:
Visually, it makes the PyRAT dialog look like this 
image

PR #269, opened 2025-02-20
New tinymce plugin docs

rlamacraft:


rlamacraft:
@mKowalski256 Still working on much of it, but figured you might want to get ahead of reviewing as I go + looking at adding to the section on authentication

DevDocs/DeveloperNotes/GettingStarted/NewEditorPlugin.md
```diff
@@ -0,0 +1,329 @@
...
+then simply use that colour, if it does not then pick the colour that is most
+associated with the branding of the organisation.
+
+
+## Step 2: Define a new TinyMCE plugin
+
+### Step 2.1: Registering a plugin
+Lets define a script in `src/main/webapp/ui/tinyMCE/plugins/newPlugin/index.js`,
+inside of which we will call `tinymce.PluginManager.add("newPlugin", NewPlugin);`,
+where `NewPlugin` is a class. This class doesn't need to have any particular
```
mKowalski256, line 75: This describes the way of writing new React-based plugin, which I think is a way preferred by us, but would it be worth mentioning that it's possible to write pure javascript-based plugin, and that various implementations can be found in `src/main/webapp/scripts/externalTinymcePlugins`?

mKowalski256 (approved):

DevDocs/DeveloperNotes/GettingStarted/NewEditorPlugin.md
rlamacraft: No, because we'll only have to migrate them ourselves when we migrate the document editor to react. The logic specific to each plugin can still be implemented using any tech stack because the dialog content can be an iframe (as I mention), but the dialog itself should be a react component so that it is consistent with the rest of the our product in terms of design, access to the help system, etc.

PR #274, opened 2025-02-27
RSDEV-527: Fix repository export with DMP fails if DMPonline app is enabled

rs-nicof:
✅  Successfully tested on AWS instance https://rsdev-527-6.researchspace.com/ (mapped to https://dmponline-test.researchspace.com/) ✅ 

Videos:

1. https://github.com/user-attachments/assets/7a4181bd-3f4f-4def-ae9f-0bd4d6843b02
2. https://github.com/user-attachments/assets/e2101d6a-10a7-42af-9762-334a88c6734b

### DIFF
--- a/src/main/java/com/researchspace/service/impl/AsyncDepositorImpl.java
+++ b/src/main/java/com/researchspace/service/impl/AsyncDepositorImpl.java
@@ -42,6 +42,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
@@ -118,8 +119,8 @@ public class AsyncDepositorImpl implements IAsyncArchiveDepositor {
   /**
    * Updates the DMPs after a deposit has been made.
    *
-   * @param result the result of the deposit to the repository
-   * @param subject the user who made the deposit
+   * @param result            the result of the deposit to the repository
+   * @param subject           the user who made the deposit
    * @param repoDepositConfig the configuration of the deposit
    */
   void updateDMPS(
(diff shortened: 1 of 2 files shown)


src/main/java/com/researchspace/service/impl/AsyncDepositorImpl.java
```diff
@@ -237,60 +238,83 @@ public Future<RepositoryOperationResult> depositArchive(
...
+
+      // read json from file
+      ObjectMapper objectMapper = new ObjectMapper();
+      JsonNode dmpFile = objectMapper.readValue(file, JsonNode.class);
+      Iterator<JsonNode> nodeIt = dmpFile.get("items").iterator();
+      String url = null;
+      while (nodeIt.hasNext() && url == null) {
+        JsonNode dmpNode = nodeIt.next();
+        if (dmpNode.get("dmp") != null) {
+          url = dmpNode.get("dmp").get("dmp_id").get("identifier").asText().replace("/api/v1", "");
```
mKowalski256, line 275: Is the core of bugfix that code now iterates over "items"->"dmp" array in json file that comes from DMP? And the bug was that previous code couldn't find the proper `url` and was setting some incorrect/empty value into `metadata.setDmpDoi()`? 

This method is called `setDmpOnlineDmpUrlOnSubmissionMetadata()` and so it is supposed to update metadata with a dmp url coming from DMPonline. The json imported from DMPonline doesn't have "items"-> "dmp" property though. Which means that while this change fixes the export that includes DMP from dmptool, it will actually break the export that includes DMP from DMPonline:
image

It seems to me that root problem is actually in the `AsyncDepositorImpl.generateSubmissionMetaData()` method, which calls:
* `setDmpOnlineDmpUrlOnSubmissionMetadata()` to handle DMPOnline scenario
* `setArgosDmpDoiOnSubmissionMetadata()`, to handle Argos scenario
... but doesn't call method for handling dmptool scenario. 

I don't know how this code could ever worked, but I think the proper way to fix it would be to add an additional method that handles json in dmptool variant. I see we have a problem that DMPUser table doesn't store the information about the source of DMP - and so the code tries to guess whether it's dmptool or dmponline or argos json based on which RSpace App is enabled. That stops working correctly as soon as user has more than one App enabled. I think the method that parses the JSON should be careful to only set `metadata.setDmpDoi()` if it's absolutely sure it found the valid url.

mKowalski256 (changes requested):
I can export with attached DMP coming from dmptool fine now on test instance, but the way the fix works is that I now get the same export error if I try to export with attached DMP coming from DMPOnline.

More details in the comment on the code.

It doesn't help that this code seems to have no unit test coverage - it'd be worth to add some, i.e. to make sure that `generateSubmissionMetaData()` puts the correct DMP url on submission metadata in scenario where export attaches DMP from dmptool, and in scenario when DMP from DMPOnline is attached.

src/main/java/com/researchspace/service/impl/AsyncDepositorImpl.java
rs-nicof: good point, amended now

mKowalski256 (commented):
It's better now, I confirmed that if I enable DMPOnline app the "export with DMP" scenario is working and correctly including DMP from either dmptool or dmponline.

1. dmptool:
image
2. dmponline:
image

But! As I've commented in the code, to go through the code path that attaches the DMP into export metatada the DMPOnline app needs to be enabled. That means (confirmed on AWS build) that if I disabled DMPOnline on Apps page, the export submission no longer includes details of DMP from dmptool. 

This is not part of the original ticket (which was about export failure), but let's get it working too. Basically the new code should be triggered when either of DMPOnline/dmptool apps is enabled, not just for DMPOnline.

src/main/java/com/researchspace/service/impl/AsyncDepositorImpl.java
```diff
@@ -234,75 +237,107 @@ public Future<RepositoryOperationResult> depositArchive(
...
-    ObjectMapper objectMapper = new ObjectMapper();
-    JsonNode dmp = objectMapper.readValue(file, JsonNode.class);
-    String url = dmp.get("dmp_id").get("identifier").asText().replace("/api/v1", "");
-    metadata.setDmpDoi(Optional.of(url));
-
+    if (archiveConfig.hasDMPs()) {
+      IntegrationInfo dmpOnlineInfo =
+          integrationsHandler.getIntegration(subject, IntegrationsHandler.DMPONLINE_APP_NAME);
+
+      if (dmpOnlineInfo.isEnabled() && dmpOnlineInfo.isAvailable()) {
```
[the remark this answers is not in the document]
rs-nicof: change done

mKowalski256 (approved):
Looks good & fixes the exports for me on test AWS build.

PR #280, opened 2025-03-04
RSDEV-557 Always set "X-Requested-With" header when making API calls

rlamacraft:
This change also introduces "@" as a an alias for the root of the project, as that was less work than manually adding the relative path in all of the places that axios is used. Is perhaps something we should use in future to making moving modules around easier.


mKowalski256 (approved):
Looks good to me - on branch build I can now see 'X-Requested-With' header on `/livechatProperties` request from new Gallery page, while it wasn't present before.

image

PR #287, opened 2025-03-14
[RSDEV-438 - RSDEV-579]: Allow direct creation of document/notebook in shared folders from UI and API

rs-nicof:
✅  Successfully tested in AWS instance https://rsdev-438-create-doc-shared-folder-35.researchspace.com/ ✅

### DIFF
--- a/src/main/java/com/researchspace/dao/GroupDao.java
+++ b/src/main/java/com/researchspace/dao/GroupDao.java
@@ -38,6 +38,8 @@ public interface GroupDao extends GenericDao<Group, Long> {
    */
   Group getByUniqueName(String uniqueName);
 
+  Group getByCommunalGroupFolderId(Long folderId);
+
   List<Group> getForOwner(User owner);
 
   /**
--- a/src/main/java/com/researchspace/service/impl/RecordManagerImpl.java
+++ b/src/main/java/com/researchspace/service/impl/RecordManagerImpl.java
@@ -18,6 +18,7 @@ import com.researchspace.dao.FieldDao;
 import com.researchspace.dao.FolderDao;
 import com.researchspace.dao.FormDao;
 import com.researchspace.dao.FormUsageDao;
+import com.researchspace.dao.GroupDao;
 import com.researchspace.dao.NameDateFilter;
 import com.researchspace.dao.RecordDao;
 import com.researchspace.dao.RecordGroupSharingDao;
@@ -132,6 +133,7 @@ public class RecordManagerImpl implements RecordManager {
   private @Autowired MovePermissionChecker movePermissionChecker;
   private @Autowired IPermissionUtils permissnUtils;
   private @Autowired RecordDao recordDao;
+  private @Autowired GroupDao groupDao;
   private @Autowired UserManager userManager;
   private @Autowired UserDao userDao;
   private @Autowired AuditDao auditDao;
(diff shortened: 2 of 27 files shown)


src/main/java/com/researchspace/dao/GroupDao.java
```diff
@@ -38,6 +38,14 @@ public interface GroupDao extends GenericDao<Group, Long> {
    */
   Group getByUniqueName(String uniqueName);
 
+  /**
+   * Gets group by folder id, or <code>null</code> if not found.
+   *
+   * @param folderId
```
rs-fraser, line 44: the `param` and `return` could be removed if there's nothing else too add

rs-fraser (approved):
Code looks good. Left 1 minor comment.

Tested running locally and creating various documents, notebooks and forms in shared lab and project group folders and didn't find any issues.

src/main/java/com/researchspace/service/impl/RecordManagerImpl.java
```diff
@@ -359,7 +363,18 @@ private StructuredDocument doCreateDocument(
...
-      parentFolder.addChild(rc, user, skipAddingToChildren);
+      if (parentFolder.isSharedFolder()) {
+        // if is a shared folder then add the document ONLY on the root folder
+        log.warn(
+            "The record with globalId[{}] will be created inside the root folder since the"
+                + " specified folder [{}] was a shared one",
+            rc.getGlobalIdentifier(),
+            parentFolder.getName());
+        Folder rootFolder = folderDao.get(user.getRootFolder().getId());
+        rootFolder.addChild(rc, user, skipAddingToChildren);
```
mKowalski256, line 374: This feels a bit hacky to replace the target in last moment, also note that later in line 379 the call to `folderDao` saves `parentFolder` rather than `rootFolder`, as it doesn't realize the folder to save was changed.

Wouldn't it be simpler/more robust to switch `parentFolder` reference to point user's root folder earlier, e.g. in line 334, as soon as `parentFolder` is retrieved and is checked to be a shared folder.

src/main/java/com/researchspace/webapp/controller/StructuredDocumentController.java
```diff
@@ -372,23 +390,36 @@ public String createSDFromTemplate(
...
+  private String redirectToDocumentEditorInEditMode(
+      StructuredDocument newRecord, List<RecordGroupSharing> sharedWithGroup) {
     String redirectUrl = "redirect:" + STRUCTURED_DOCUMENT_EDITOR_URL + "/" + newRecord.getId();
     if (newRecord.isNotebookEntry()) {
       redirectUrl += "?fromNotebook=" + newRecord.getParentNotebook().getId();
     } else {
       redirectUrl += "?editMode=true";
     }
+    if (!(sharedWithGroup == null || sharedWithGroup.isEmpty())) {
+      redirectUrl += "&sharedWithGroup=" + sharedWithGroup.get(0).getSharee().getDisplayName();
```
mKowalski256, line 421: the group's display name should be encoded for url-safety, otherwise it may become interpreted incorrectly. E.g. try a group named 'Matt & Nico Lab Group', and see the `&` char being interpreted as a start of new parameter :

image

src/main/java/com/researchspace/webapp/controller/WorkspaceController.java
```diff
@@ -398,9 +401,34 @@ public String createNotebookAndRedirect(
...
+  }
 
-    Long newNotebookId = createNotebook(parentRecordId, notebookName, principal);
-    return "redirect:/notebookEditor/" + newNotebookId;
+  @NotNull
+  private static String getNotebookRedirectUrl(
+      Long newNotebookId, List<RecordGroupSharing> sharedWithGroup) {
+    String redirectUrl = "redirect:/notebookEditor/" + newNotebookId;
+    if (!(sharedWithGroup == null || sharedWithGroup.isEmpty())) {
+      redirectUrl += "?sharedWithGroup=" + sharedWithGroup.get(0).getSharee().getDisplayName();
```
mKowalski256, line 429: similarly, display name needs to be url-encoded.

mKowalski256 (commented):
Looks good & works fine for me, but I've added a few comments.

src/main/java/com/researchspace/webapp/controller/WorkspaceController.java
rs-nicof: now encoded

src/main/java/com/researchspace/webapp/controller/StructuredDocumentController.java
rs-nicof: now encoded

mKowalski256 (approved):
Looks good and reported issues are fixed now!

I did one more round of functional testing, and found another problem. 'Create' menu is available not just within group-share folders, but also within individual-share folders. These are folders that are initialized after user shares some content with an individual group member (rather than a group). Attempt to create content in that folder fails with an error. But I'm thinking I'll raise that separately as a PRT issue, so the main PR can be merged, and individual share case is treated separately.

PR #291, opened 2025-03-20
RSDEV-550: disable chemistry on startup

mKowalski256:
The main functional change in this PR is adding `SystemConfigurationInitialisor` class (and test), which runs on RSpace startup and checks if deployment properties connected to chemistry are configured. If these are not configured, the initializor disables 'chemistry.available' system setting (visible on System -> Configuration -> Settings page). Disabling that setting means users will not be able to use chemistry features, which is what we want as the features won't work without proper configuratino - that's different from old chemaxon engine, where even misconfigured setup still allowed chemical calculations being run, as most of them was based on chemaxon libraries bundled within RSpace.

But then, when working on the PR, I've noticed that backend code that tries to read/update system settings does operate on string names of the properties (rather than strong-typing) and these string names are generally copied across the code. That does seem to be error-prone and not a great practise, so I've added an enum class `SystemPropertyName` that collects system settings used by back-end code, and I've added strong-typed `findByName` and `save` methods to `SystemPropertyManager`. 
There is a bit of extra complexity coming from the fact that some historical settings are hardcoded in `Preference` class, and some code does actually want to operate on weakly-typed setting names, but hopefully the PR is a step forward. The code also tries to use `HierarchicalPermission` enum rather than `ALLOWED`/`DENIED` string values.

### DIFF
--- a/src/main/java/com/axiope/service/cfg/BaseConfig.java
+++ b/src/main/java/com/axiope/service/cfg/BaseConfig.java
@@ -207,6 +207,7 @@ import com.researchspace.service.impl.SanityChecker;
 import com.researchspace.service.impl.SharingHandlerImpl;
 import com.researchspace.service.impl.StrictEmailContentGenerator;
 import com.researchspace.service.impl.SysadminUserCreationHandlerImpl;
+import com.researchspace.service.impl.SystemConfigurationInitialisor;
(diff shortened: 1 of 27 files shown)


rs-fraser (approved):
Code changes look good.

Tested running locally and setting the system property to ALLOWED but not supplying one of the deployment properties for chemistry would update the system property to DENIED.

PR #292, opened 2025-03-21
PRT-902: Disable the create menu ONLY for individual shared folder

rs-nicof:


src/main/java/com/researchspace/webapp/controller/WorkspacePermissionsDTOBuilder.java
```diff
@@ -41,7 +41,7 @@ public ActionPermissionsDTO addCreateAndOptionsMenuPermissions(
       boolean isSearch) {
 
     boolean createRecord =
-        parentFolder.isSharedFolder()
+        parentFolder.hasType(RecordType.SHARED_GROUP_FOLDER_ROOT)
```
mKowalski256, line 44: That only checks whether parent is a top-level group-share folder, so it's not going to work in subfolders of group-share folder. You can see it on test instance, that after creating a subfolder the 'CREATE' options disappear:

image

Saying that, maybe a condition like that would work:
```
    boolean createRecord =
        (parentFolder.isSharedFolder() && !parentFolder.hasType(RecordType.INDIVIDUAL_SHARED_FOLDER_ROOT))
            || parentFolder.getSharingACL().isPermitted(usr, PermissionType.CREATE);
```

That's based on the fact that you currently cannot create subfolders inside individual-share folder.

mKowalski256 (changes requested):

src/main/java/com/researchspace/webapp/controller/WorkspacePermissionsDTOBuilder.java
rs-nicof: amended

mKowalski256 (approved):
Looks good & works for me!

PR #301, opened 2025-04-08
RSDEV-627: Make sure to create IGSNs in bulk ONLY if Datacite is enabled

rs-nicof:


src/main/java/com/researchspace/service/inventory/impl/InventoryIdentifierApiManagerImpl.java
```diff
@@ -105,11 +105,13 @@ public List<ApiInventoryDOI> registerBulkIdentifiers(Integer igsnsToAllocate, Us
...
         currentDoi = createNewDoi(user);
         DigitalObjectIdentifier dbObj = apiIdentifiersHelper.createDoiToSave(currentDoi, user);
-        dbObj = doiDao.save(dbObj);
+        log.info("New IGSN allocated: {}", dbObj.getIdentifier());
 
+        dbObj = doiDao.save(dbObj);
         result.add(new ApiInventoryDOI(dbObj));
       } catch (Exception ex) {
         log.warn("It was not possible to allocate IGSN: {}", ex.getMessage(), ex);
+        throw ex;
```
mKowalski256, line 114: I think throwing an exception here would rollback the transaction that is started when controller calls the manager method. If the exception happens at the start of `for` loop that's fine, but if it happens later, there will be draft IGSNs left registered in DataCite, but without of any trace of them in RSpace. 

As the registration of draft IGSNs is not easily revokable (we could attempt to delete the IGSN, but that can fail too) I think better way would be if bulk creation endpoint counted the number of generated IGSNs, and returned an error if the number doesn't match the requested number. The error message could be something like "Requested registration of X draft IGSNs, but only managed to register Y".

src/main/java/com/researchspace/webapp/integrations/datacite/DataCiteConnectorDummyError.java
mKowalski256, line 1: Looks good, but I've noticed this class `DataCiteConnectorDummyError` is only used in tests, same goes for pre-existing class `DataCiteConnectorDummy`. Should these two classes maybe be moved from `/src/main/java` into `/src/test/java`?

rs-nicof: good point, moved

src/main/java/com/researchspace/service/inventory/impl/InventoryIdentifierApiManagerImpl.java
rs-nicof: added the created number check

mKowalski256: But you've just extended the message in the logs, didn't change the error returned by the endpoint, right? 

The problem I'm seeing in this code is that `registerBulkIdentifiers()` method can encounter an exception in the middle of `for` loop, after some IGSNs were already created in Datacite. The code in this PR makes encountered exception to be re-thrown, that means that RSpace database transaction will rollback, and there will be orphaned draft IGSNs on Datacite side, that RSpace will no longer know about. 

In other words, the code will work fine if Datacite exception is encountered on first iteration of `for` loop, in that case probably nothing was created in Datacite. The problem is if exception happens in later iterations.

As I was saying in previous comment, the code shouldn't leave the Manager method with exception, to not trigger transaction rollback. The method needs to end normally, so the db transaction completes, and data about successfully registered DOIs is committed into RSpace database. Then it should be `InventoryIdentifiersApiController` that recognize if action wasn't a full success, and either returns an error to the client (with the details of the problem), or does some other stuff.

mKowalski256 (changes requested):

src/main/java/com/researchspace/service/inventory/impl/InventoryIdentifierApiManagerImpl.java
rs-nicof: As discussed on the call, I've raised the exception on the Controller but left the exception ithrowing in the for loop in case it is a dataciteException and it is on the first iteration (that basically means server misconfiguration/credential errors)

mKowalski256 (approved):
Looks good and works fine for me!

mKowalski256 (commented):
Actually, I didn't check the latest version of the code, as my `git pull` failed due to force-push... give me a moment to review again.

mKowalski256 (approved):
Ok, works for me with the latest code too, so good to merge!

PR #315, opened 2025-04-25
RSDEV-630: changing behaviour of api.available setting

mKowalski256:
This PR changes behaviour of `api.available` setting that sysadmin can modify on 'System -> Configuration -> System Settings page'. Previously, the setting was checked by `APIAvailableInterceptor` that was intercepting all requests to public api endpoints, (i.e. requests under `/api/**` path), and if the setting was disabled, the request was rejected.

New behaviour doesn't use `APIAvailableInterceptor` anymore, as the public API code is used also by RSpace UI, and we're slowly migrating UIto use more of it. Therefore the setting shouldn't block access to the particular endpoints, but rather should stop connection attempts outside of RSpace UI. The RSpace UI is authenticating with public API using oauth tokens generated during http session, and so disabling `api.available` setting is no longer stopping such calls. Instead, the setting is stopping the calls that use API key (new check added into `ApiAuthenticationInterceptor`), and it also stops user from generating/refreshing oauth token through `OAuthClientController` password/refresh flows.

Additionally, as a partial solution to RSDEV-181, the analytics event `apiUsed` was replaced with two new events: `apiKeyUsed` and `apiOAuthTokenGenerated`. First is triggered for public API calls that use API key, second is triggered on usage of oauth authentication workflow.

The PR also add some refactoring (I'm sorry, that complicates the changeset, should be done separately):
- rename `ApiAuthenticatorImpl` class, one used in interceptor registry, to `MainRSpaceApiAuthenticator`
- replacing usage of `inventory.available` String with enum `SystemPropertyName.INVENTORY_AVAILABLE` 
- extending `OAuthClientControllerMVCIT` to specifically cover each failure scenario


mKowalski256:
There is AWS build running at: https://rsdev-630-api-avaialble-setting-14.researchspace.com. I've configured dev segment analytics there, so the new events are passed to posthog - assuming the API is enabled:
image

After logging as sysadmin1 and disabling the API access, the [API docs](https://rsdev-630-api-avaialble-setting-14.researchspace.com/public/apiDocs) should no longer allow to run queries using API key, and oauth authentication workflow shouldn't work either.

rs-fraser (approved):
Code looks good. Made 1 minor comment about naming. 

Also occurred to me we should update any internal jira ticket references within the code to be more useful to open-source contributors. I'll create a ticket for that.

Tested running locally and toggling `api.available` enables and disables the use of api tokens as expected.

src/main/java/com/researchspace/api/v1/auth/MainRSpaceApiAuthenticator.java
```diff
@@ -0,0 +1,59 @@
...
+
+    this.apiKeyAuthenticator = apiKeyAuthenticator;
+    this.oAuthAuthenticator = oAuthAuthenticator;
+    this.analyticsMgr = analyticsManager;
+    this.apiHandler = apiAvailabilityHandler;
+  }
+
+  @Override
+  public User authenticate(HttpServletRequest request) {
+    if (request.getHeader("apiKey") != null) {
```
rs-nicof, line 34: Would `StringUtils.isNotBlank()` work better in theis case?

src/main/java/com/researchspace/api/v1/auth/MainRSpaceApiAuthenticator.java
```diff
@@ -0,0 +1,59 @@
...
+        if (!apiHandler.isApiAvailableForUser(user)) {
+          throw new ApiAuthenticationException(
+              String.format("Access to API has been disabled for user '%s'", user.getUsername()));
+        }
+        analyticsMgr.apiAccessed(user, true, request);
+      }
+      return user;
+    }
+
+    if (request.getHeader("Authorization") != null) {
```
rs-nicof, line 50: Have you considered externalising "`Authorization`" and "`apiKey`" as costants?

rs-nicof (approved):

src/main/java/com/researchspace/api/v1/auth/MainRSpaceApiAuthenticator.java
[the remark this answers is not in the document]
mKowalski256: Yeah I wanted to rename the class so it's obvious this is the entry point for authentication, to not confuse with other ApiAuthenticator implementations which are for specific authentication methods. But RSpace is redundant here, I'll rename to more descriptive `CombinedApiAuthenticator`.

src/main/java/com/researchspace/service/impl/ApiAvailabilityHandlerImpl.java
```diff
@@ -26,29 +27,26 @@ void setSystemPropertyManager(SystemPropertyPermissionManager systemPropertyMana
   @Override
   public ServiceOperationResult<String> isAvailable(User user, HttpServletRequest request) {
     // rsinv-365
```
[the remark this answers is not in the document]
mKowalski256: Sounds like a good idea to at least update pre-existing comments this way, if we work with the code nearby. I'll do it for this case.

src/main/java/com/researchspace/api/v1/auth/MainRSpaceApiAuthenticator.java
mKowalski256: We could, but they are used quite a lot across codebase, so I'll leave it for next time.

PR #327, opened 2025-05-06
PRT-913: Fix info panels when move document into sharedNotebooks

rs-nicof:
This PR is depending from: https://github.com/rspace-os/rspace-core-model/pull/25

✅  Successfully tested on AWS instance: https://prt-913-fix-permissions-20.researchspace.com/ ✅

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3009,7 +3009,7 @@
     <jstl.version>1.1.2</jstl.version>
     <lucene.version>5.5.5</lucene.version>
     <netty.version>4.1.63.Final</netty.version>
-    <rspace-core-model.version>PRT-913-MODEL-SNAPSHOT</rspace-core-model.version>
+    <rspace-core-model.version>2.10.0</rspace-core-model.version>
     <servlet.version>3.1.0</servlet.version>
     <sitemesh.version>2.4.2</sitemesh.version>
     <urlrewrite.version>4.0.3</urlrewrite.version>
--- a/pom.xml
+++ b/pom.xml
@@ -3009,7 +3009,7 @@
     <jstl.version>1.1.2</jstl.version>
     <lucene.version>5.5.5</lucene.version>
     <netty.version>4.1.63.Final</netty.version>
-    <rspace-core-model.version>2.9.0</rspace-core-model.version>
+    <rspace-core-model.version>PRT-913-MODEL-SNAPSHOT</rspace-core-model.version>
     <servlet.version>3.1.0</servlet.version>
     <sitemesh.version>2.4.2</sitemesh.version>
     <urlrewrite.version>4.0.3</urlrewrite.version>
--- a/src/main/java/com/researchspace/service/impl/SharingHandlerImpl.java
+++ b/src/main/java/com/researchspace/service/impl/SharingHandlerImpl.java
@@ -1,7 +1,5 @@
 package com.researchspace.service.impl;
 
-import static com.researchspace.service.RecordDeletionManager.MIN_PATH_LENGTH_TOSHARED_ROOT_FOLDER;
-
 import com.researchspace.model.AbstractUserOrGroupImpl;
 import com.researchspace.model.Group;
 import com.researchspace.model.RecordGroupSharing;
@@ -33,7 +31,6 @@ import com.researchspace.service.UserManager;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.stream.Collectors;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.Validate;
@@ -98,20 +95,7 @@ public class SharingHandlerImpl implements SharingHandler {
   @Override
   public ServiceOperationResultCollection<RecordGroupSharing, RecordGroupSharing>
       shareIntoSharedFolder(User user, Folder sharedFolder, Long recordId) {
-    Optional<Folder> sharedFolderRoot =
-        folderManager.getGroupOrIndividualShrdFolderRootFromSharedSubfolder(
-            sharedFolder.getId(), user);
-    RSPath path = folderManager.getShortestPathToSharedRootFolder(sharedFolder.getId(), user);
-    if (path.isEmpty() || path.size() <= MIN_PATH_LENGTH_TOSHARED_ROOT_FOLDER) {
-      String msg =
-          String.format(
-              "The folder '%s' is not a shared subfolder - id [%d] is not in a shared folder!",
-              sharedFolder.getName(), sharedFolder.getId());
-      throw new IllegalArgumentException(msg);
-    }
-
-    Group sharedGroup =
-        groupManager.getGroupByCommunalGroupFolderId(sharedFolderRoot.get().getId());
+    Group sharedGroup = groupManager.getGroupFromAnyLevelOfSharedFolder(user, sharedFolder);
     ShareConfigElement shareConfigElement = new ShareConfigElement(sharedGroup.getId(), "write");
     shareConfigElement.setGroupFolderId(sharedFolder.getId());
     ShareConfigCommand shareConfig =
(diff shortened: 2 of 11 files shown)


src/main/java/com/researchspace/webapp/controller/WorkspaceController.java
```diff
@@ -597,20 +597,39 @@ public ModelAndView move(
...
+          User recordOwner = recordManager.get(recordIdToMove).getOwner();
+          if (baseRecordToMove.isShared()) {
+            RSPath pathToRootSharedFolder =
+                folderManager.getShortestPathToSharedRootFolder(target.getId(), recordOwner);
+            sharingManager.unshareFromSharedFolder(
+                recordOwner, baseRecordToMove, pathToRootSharedFolder);
+          }
+          ShareConfigCommand shareConfig = buildShareCommandForTarget(recordIdToMove, target);
+          SharingResult sharingResult =
+              this.shareRecord(shareConfig, getPrincipal(recordOwner)).getData();
```
mKowalski256, line 623: This change is good, but now when a 1-step move action is redesigned into 2-step process, it should be encoded into one manager call, which is fully reverted on error/exception. Otherwise an action can end up half-done. 

i.e. currently if there is an error/exception in 2nd call (to `sharingManager.shareRecord()`), the 1st call (to `sharingManager.unshareFromSharedFolder()`) is not reverted, as transactions are only started when moving from Controller code to Manager code, so the code above makes two separate transactions (1st for unshare action, 2nd for re-share). 

You can reproduce it the following way:
1. as user A, share a notebook N, and (separately) document D
2. go to shared folder, try moving document D into notebook N
-- there is an error popup, as sharing into own notebook is blocked
image
-- but after closing the popup the document D is no longer shared, as the "unshare" part of the action completed/was committed fine

I think the better error-handling for multi-step processes is ensured if a method on Manager level throws an exception if there is a problem, and Controller converts this exception in error message. When exception is thrown from Manager, the database is rollbacked i.e. the action is fully reverted. 

Alternatively, when WorkspaceController realizes the `shareRecord()` failed, it could try to re-share document into old location, as a part of error handling.

rs-nicof: yes I see what you mean. Let me refactor it to be atomic

rs-nicof: changes done and successfully tested

pom.xml
```diff
@@ -3009,7 +3009,7 @@
     <jstl.version>1.1.2</jstl.version>
     <lucene.version>5.5.5</lucene.version>
     <netty.version>4.1.63.Final</netty.version>
-    <rspace-core-model.version>2.9.0</rspace-core-model.version>
+    <rspace-core-model.version>PRT-913-MODEL-SNAPSHOT</rspace-core-model.version>
```
rs-nicof, line 3012: this will be `2.10.0`

src/main/java/com/researchspace/service/impl/SharingHandlerImpl.java
```diff
@@ -162,4 +171,88 @@ private ServiceOperationResult<RecordGroupSharing> doUnshare(
...
+    return sharingResult;
+  }
+
+  @NotNull
+  private static ShareConfigCommand buildShareCommandForTarget(Long idToMove, Folder target) {
+    ShareConfigElement shareElement = new ShareConfigElement();
+    Group group = target.getOwner().getPrimaryLabGroupWithPIRole();
+    shareElement.setGroupid(group.getId());
+    shareElement.setGroupFolderId(target.getId());
+    shareElement.setOperation("write");
```
mKowalski256, line 218: hardcoding permission type to 'write' is also too simplistic, we should keep the original permission that sharing user put on a document, which may be 'read' permission. The details how record was shared with various groups, including permission level, can be retrieved by calling

```
recordSharingManager.getRecordSharingInfo(recordID)
```

src/main/java/com/researchspace/service/impl/SharingHandlerImpl.java
```diff
@@ -162,4 +171,88 @@ private ServiceOperationResult<RecordGroupSharing> doUnshare(
...
+    shareElement.setAllowedOps(List.of("read", "write"));
```
mKowalski256, line 219: is call to `setAllowedOps()` actually needed? I don't see it being called or used anywhere.

mKowalski256 (commented):
Rollback-on-error works fine for my tests, thanks. I've noted other issue though: in re-sharing step, the new share configuration (used for share into notebook) doesn't necessarily match the original share configuration (selected by user when sharing the original document). See the comments inline.

src/main/java/com/researchspace/service/impl/SharingHandlerImpl.java
rs-nicof: I have reoplicated the call the the UI makes when share inrto a SharedNotebook and that is basically the request

src/main/java/com/researchspace/service/impl/SharingHandlerImpl.java
mKowalski256: I've re-checked, and from my testing it seems that when sharing directly into notebook the UI indeed hardcodes the permission type, but it is hardcoded to 'read' rather than 'write'.

image

So you're right it's fine to hardcode, but it should be hardcoded as 'read'.

src/main/java/com/researchspace/service/impl/SharingHandlerImpl.java
mKowalski256: I see, the thing is that `allowedOps` field in ShareConfigElement is set during construction, so this line is redundant. 

See https://github.com/rspace-os/rspace-web/blob/b60f3225814b66c35b991c957c7ed91c63176cc8/src/main/java/com/researchspace/model/dtos/ShareConfigElement.java#L22

src/main/java/com/researchspace/service/impl/SharingHandlerImpl.java
rs-nicof: changed

src/main/java/com/researchspace/service/impl/SharingHandlerImpl.java
rs-nicof: removed the line that was redundand

src/main/java/com/researchspace/service/impl/SharingHandlerImpl.java
```diff
@@ -162,4 +171,88 @@ private ServiceOperationResult<RecordGroupSharing> doUnshare(
...
+              + sharingResult.getError().getAllErrorMessagesAsStringsSeparatedBy(",")
+              + "]");
+    }
+    return sharingResult;
+  }
+
+  @NotNull
+  private static ShareConfigCommand buildShareCommandForTarget(Long idToMove, Folder target) {
+    ShareConfigElement shareElement = new ShareConfigElement();
+    Group group = target.getOwner().getPrimaryLabGroupWithPIRole();
```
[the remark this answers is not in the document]
rs-nicof: amended

src/main/java/com/researchspace/webapp/controller/WorkspaceController.java
```diff
@@ -597,20 +596,42 @@ public ModelAndView move(
...
       ServiceOperationResult<? extends BaseRecord> moveResult = null;
       if (isFolder) {
-        moveResult = folderManager.move(id, target.getId(), sourceFolder.getId(), user);
+        moveResult = folderManager.move(recordIdToMove, target.getId(), sourceFolder.getId(), user);
       } else {
-        moveResult = recordManager.move(id, target.getId(), sourceFolder.getId(), user);
+        BaseRecord baseRecordToMove = recordManager.get(recordIdToMove);
+        if (target.isNotebook() && target.isShared()) {
+          try {
+            Group group = groupManager.getGroupByCommunalGroupFolderId(sourceFolder.getId());
```
mKowalski256, line 614: this will only resolve the `group` if `sourceFolder` equals shared folder, i.e. if user is moving document that is on top-level of shared folder. If `sourceFolder` is a subfolder, i.e. if the user is moving document from a subfolder of shared folder, the `group` resolves to `null`, and move action fails.

image

rs-nicof: mmm good point. I've created a new method in the group manager that will work better. 
The current PR then covers also this scenarion now

mKowalski256 (approved):
I've retested on https://prt-913-fix-permissions-24.researchspace.com/, looks good & works for me!

PR #329, opened 2025-05-08
RSDEV-665: chemical search performance fixes 

mKowalski256:
## Description ##

This PR makes a few changes to RSpace-side of collecting chemical search results:
* main change is that `RSChemElementDaoHibernate` now calls `getSession().clear();` after collecting a set of chemistry results, which stops possible out-of-memory exception, which could happen when method was called multiple times during processing single search request
* another minor performance change on DAO level is using `READ_ONLY` query hint, that is fine as we only read the results (i.e. show them on search results page)
* on manager level, in `RSChemElementManagerImpl.getFilteredChemSearchedItems()` method, the order of checks is reverted when checking if given document, which contains a matching chemical hit, should be displayed to the user.

## Design decisions

The first two changes shouldn't be controversial, but on the reversed order of checks during filtering the results:

Previously the first check was to parse the field content to ensure the linked chemical element is surely there, then the second check was done to ensure the user has permission to see the document. But in real searches, the field almost always contains the linked chem item, while searching user quite often doesn't have permission to see the given document. Reversing the check should therefore speed-up search for significant number of users.

### DIFF
--- a/src/main/java/com/researchspace/dao/RSChemElementDao.java
+++ b/src/main/java/com/researchspace/dao/RSChemElementDao.java
@@ -8,7 +8,7 @@ public interface RSChemElementDao extends GenericDao<RSChemElement, Long> {
 
   List<RSChemElement> getAllChemElementsFromField(Long fieldId);
 
-  List<RSChemElement> getChemElementsForChemIds(List<Long> chemIds);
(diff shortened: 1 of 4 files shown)


rs-fraser (approved):
Code changes look good. I've done some basic testing on the feature branch build and search still works as expected.

PR #333, opened 2025-05-13
Add warning text to API key password confirmation dialog

rlamacraft:
## Description ##
When generating or regenerating an API key, the user is now presented with a warning about the security implications of using API keys.

## Design decisions
N/A

## Testing notes
To test, just generate or regenerate an API key from My RSpace.


mKowalski256 (approved):
Looks good & works fine for me!

PR #335, opened 2025-05-14
Rsdev 611 fix ketcher preview inventory

rs-fraser:
## Description ##
When a chemistry file attachment to an Inventory item was opened in Ketcher, the contents of the file were first being converted to the "default" format before being read by Ketcher. This causes some information to be lost, and is needless. 

The fix is to load the file in it's original format, using `ChemistryProvider.convert(File)` which reads the file contents, handling conversion to base64 if the file is a `.cdx`. This same fix was applied in other cases where files are loaded into Ketcher.

before:
Screenshot 2025-05-14 at 14 39 00

after:
Screenshot 2025-05-14 at 14 39 24


mKowalski256 (approved):
Looks good and fixes the issue on test AWS build.

PR #336, opened 2025-05-15
RSDEV-644 Add calendar button to messaging page

rlamacraft:
## Description ##
This is part of a wider change to make the messaging system more readily available across the product, rather than just being centered on the workspace. The ability to create new calendar entries was, until now, functionality only reachable from the workspace and not the messaging page itself, which is more appropriate given that it is not specific to the ELN.

## Design decisions
None

## Testing notes
To test, just go to the messaging page and use the new button


mKowalski256 (commented):
The dialog for creating calendar event has a checkbox to 'include links to resources', which upon clicking on Workspace page opens filetree, but in messaging page shows a console error:

image

rlamacraft:
> The dialog for creating calendar event has a checkbox to 'include links to resources', which upon clicking on Workspace page opens filetree, but in messaging page shows a console error:

Should be fixed now

mKowalski256 (approved):
The code looks good and calendar creation on Messaging page works fine for me.

PR #348, opened 2025-05-28
RSDEV-664: oauth tokens

mKowalski256:
## Description ##

Main change of this PR is added handling of new field `tokenType` in `OAuthToken` entity coming from [rspace-core-model#PR27](https://github.com/rspace-os/rspace-core-model/pull/27) . The field distinguishes tokens generated during UI session by shiro-authenticated call to `/userform/ajax/inventoryOauthToken` endpoint (handled by `UserProfileController.getInventoryOauthToken()` method), from tokens generated by calling public `/oauth/token` endpoint as a part of username/password grant flow (handled by `OAuthClientController.getToken()` method).

There is a liquibase changelog applying `OAuthToken` table changes and relaxing the db table constraint to allow tokens of various types saved for the same user and oauthapp.

The secondary change is to api-related analytics events, and I've added it to this PR mainly to showcase that new `tokenType` field allows to effectively distinguish the internal/external API interactions. The pre-existing events `apiKeyUsed` and `apiOAuthTokenGenerated` are replaced by single event `publicApiUsed`, as discussed in [RSDEV-181](https://researchspace.atlassian.net/browse/RSDEV-181?focusedCommentId=62961). This new event is triggered either when authenticating request with API key, or with non-internal oauth token.

## Testing notes ##

Note that to enable external OAuth authentication flow the sysadmin must enable system setting `api.oauth.authentication` (on System -> Configuration -> System Settings page).


mKowalski256:
The test server is running at https://rsdev-664-oauth-tokens-16.researchspace.com, it has analytics enabled and so the new `publicApiUsed` events should be appearing in posthog.

PR #352, opened 2025-05-29
Dynamically assign playwright port

rlamacraft:
## Description ##
@nebay-abraha's great suggestion to use `getPort` to prevent issues with running multiple instances of playwright component testing in parallel.

As background, Playwright components testing spins up a webserver to serve simple HTML pages including just the React component under test, which are then visited by the headless browser. By hardcoding the port number, we were preventing multiple instances of the Playwright component test runner from executing in parallel because they would all attempt to acquire the same port.

### DIFF
--- a/src/main/webapp/ui/package-lock.json
+++ b/src/main/webapp/ui/package-lock.json
@@ -116,6 +116,7 @@
         "fast-check": "^3.6.2",
         "flow-bin": "=0.239.1",
         "flow-typed": "=3.9.0",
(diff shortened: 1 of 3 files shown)


rlamacraft:
@rs-fraser @mKowalski256 I think this should fix the immediate issue of the Jenkins run completely freezing up, although of course it does nothing for total performance.

mKowalski256 (approved):
Looks good & worth trying.

PR #353, opened 2025-05-29
Rsdev 649 mark all notifications

rs-fraser:
## Description ##
**Mark all as read**
When a user tried to mark lots of notifications as read at once, the query would hang and eventually require intervention to kill the hanging db processes.

I was able to recreate the issue using the [attached sql script](https://github.com/user-attachments/files/20507396/sql.txt) to recreate the scenario for the user, where they had 10k notifications. (10K total, unsure whether they were unread or not, but the way the query originally worked all would need to be loaded into memory anyway).

The original query was inefficient as we loaded all `Notification` entities then looped through each `CommunicationTarget` (recipient), and if the recipient matches the current user then the status is updated. 

Also, statuses were updated and `lastStatusUpdate` was set each time the user marked all as read, even for `CommunicationTarget` entities which already had the COMPLETE status. This mean that the batch delete job detailed below wouldn't clear-up those notifications, if the user frequently marked all as read, as the batch job deletes 14 days after `lastStatusUpdate`. That could explain why the user had 10K notifications.

Instead, now there is 1 query to perform the same action of updating all of the statuses to COMPLETE, and the update excludes `CommunicationTarget` entities which are already COMPLETE to avoid incorrectly updating `lastStatusUpdate`.

**Delete all read notifications**
There's a batch job which runs by default nightly which deletes all `Notification` entities from the db, if all of the recipients have read it, and `lastStatusUpdate`. When I tried to manually run this job with 10k notifications to delete, it also caused a lock on the db.

Instead, now the ids of all `Notification` entities in scope for deletion are found, then used to delete `CommunicationTarget` and `Notification` entities. There was a comment in the code explaining the Notification entities had to be deleted one-by-one via the hibernate session so the deletion was registered with envers, but those entities aren't audited (i.e. there's no `_AUD` table for either) so I don't think it needs to be done like that, and have updated it to delete in 3 queries.

## Testing notes
I added a button to the front-end (which will be removed before merging) to test the sequence manually after running the sql script to generate 10k entities.

1. run rspace-web locally with a clean db
2. login as sysadmin1, and create a lab group with user1a as pi and user2b as a group member
3. run the attached sql script to generate 10k notifications for user2b [sql.txt](https://github.com/user-attachments/files/20507396/sql.txt)
4. login as user2b, check notifications and click 'Delete All' (which marks all as read in the db)
5. click 'Delete all read'

Previously, steps 4 and 5 would hang.


src/main/java/com/researchspace/webapp/controller/DashboardController.java
rs-fraser, line 1: only for manual testing - will be removed

PR #357, opened 2025-05-30
Rsdev 682 websocket backend

rs-fraser:
## Description ##
This change adds a simple message broker where a client can subscribe to a topic at `/topic/notifications/{userId}` and receive an updated notification count when the user has a new message.

The format of the message is the same as the existing polled message:
```
{
 "notificationCount": 15,
 "messageCount": 42,
 "specialMessageCount": 7
}
```

There's example code for how to connect in this poc branch: https://github.com/ResearchSpace-ELN/rspace-web/tree/rsdev-679-backend-websockets-poc

A message is triggered each time a message or notification count changes inside `MessageAndNotificationTracker`.

## Testing notes
This change can be seen in action in the POC branch, which has these changes plus the front-end modified to use websockets to update notificaitions.


rlamacraft:
What authentication mechanism is it using? I'm not seeing an auth headers in the POC branch, unless I'm missing something

rlamacraft:
Also, what triggers a special message?

rs-fraser:
> Also, what triggers a special message?

Special messages are defined on the backend as 
REQUEST_CREATE_LAB_GROUP
REQUEST_JOIN_LAB_GROUP
REQUEST_SHARE_RECORD
GLOBAL_MESSAGE
REQUEST_JOIN_PROJECT_GROUP

I'm not sure how a user would actually trigger those requests though - e.g. where could they see groups or records they don't have access to to be able to generate a request.

rlamacraft:
> > What authentication mechanism is it using? I'm not seeing an auth headers in the POC branch, unless I'm missing something
> 
> It's using shiro (session) auth, which is the fallback when nothing else is specified. I assume on the front-end the `JSESSIONSID` token is added via some middleware? Is that what happens with other requests which aren't api key or token based?

If an API key isn't specified then we fallback on the browser cookie, which may well store that token -- I don't remember. I'm just wondering whether it should be API key driven, I guess it probably doesn't matter too much as I don't really see a third-party use-case

rs-fraser:
I’ve updated this PR as follows:

- added origin checking for the websockets endpoint, applying the same allowed list as is allowed for http web requests using `OriginReferrerChecker`
- fixed issues with existing tests by excluding the websocket from being configured in the test context, and supplying mock beans

rs-fraser:
> [Left a comment on the prototype UI code](https://github.com/ResearchSpace-ELN/rspace-web/commit/a0e10fa5187c84dde55992b9700690b046e1ce35#r160126344), highlighting a bug resulting from the client not knowing what the initial values are when connecting.

Should be straightforward enough - we could either
1. call the existing endpoint (the one which we currently poll) when the page loads to get the current notification count
2. hook into some websocket connection event on the back-end, and publish a message with the current count

I think 2 is the better solution as then the front-end code doesn't need to manage 2 ways of getting notification counts. I'll look into it shortly

rs-fraser:
@rlamacraft I've added an event listener, so that when the front-end subscribes to a topic, a message is immediately published showing the users notification counts.

rlamacraft:
> @rlamacraft I've added an event listener, so that when the front-end subscribes to a topic, a message is immediately published showing the users notification counts.

That's great, thanks -- got everything to implement a badge counter in the app bar now

rs-fraser:
> @mKowalski256 @rs-fraser Just worth checking that this isn't potentially a privacy issue? Any users will be able to query the API for how many notifications and messages that any other user currently has. Even if we don't think that is an issue, perhaps worth making very clear in the code so that we don't end up extending support for content of messages, which definitely would be.

Good point, I'll handle it now, we can grab the user making the request from the auth header then ensure their ID matches the topic they're subscribing to. As you say it's probably not a major concern now as there's nothing sensitive in a notification count, but it could easily be forgotten about if/when we expand websocket usage.

PR #359, opened 2025-06-02
RSDEV-686: Check if Inventory is enabled before to allow `identifiers`end point

rs-nicof:
## Description ##
- Adding the `inventory enabled`check when receiving a call to the `/api/inventory/v1/identifiers/*` endopoints


src/main/java/com/researchspace/api/v1/controller/InventoryIdentifiersApiController.java
```diff
@@ -150,13 +159,25 @@ public boolean testDataCiteConnection(User user) {
     return dataCiteConnector.testDataCiteConnection();
   }
 
+  private void assertInventoryAndDataciteEnabled(User user){
+    assertDataCiteConnectorEnabled();
+    assertInventoryAvailable(user);
```
rlamacraft, line 164: Can these be flipped around, so that if neither are configured then it is Inventory that is reported in the error message? In order to enable DataCite, a sysadmin has to first enable Inventory so that should be the error reported to them by users

rlamacraft (changes requested):
Thanks for fixing, just one comment about which assertion triggers the error first

PR #367, opened 2025-06-03
Only run changed since js tests

rlamacraft:
## Description ##
In feature branch builds, only run the jest and playwright component tests that could actually be impacted by the changes since the branch diverged from `main`.


rlamacraft:
Not sure if this is the most efficient way to encode this? Perhaps someone else has a better idea

PR #373, opened 2025-06-17
Rsdev 684 limit search input to a single molecule

rs-fraser:
## Description ##
Chemistry search currently only works with a single molecule as input, although there was nothing stopping users adding multiple elements. This change adds some helper text at the top of the search dialog, and also validates that there's only 1 chemical element on the canvas when the user submits the search.

Other minor changes to get this to work:
- Moved the code to reset the canvas `window.ketcher.setMolecule("")`, which was handled in the child Ketcher component and tied to clicking the submit button, to each of the parents, as in the case of search we don't always want to clear the canvas, as the search input might not be valid. This way the user has a chance to fix the issue without having to recreate their search input
- Wrapped `KetcherTinyMCE` in a `<ThemeProvider>` as there was an error around themes after adding the `ValidatingSubmitButton` to the child component

There's also an auto-formatting change in some unrelated java code.

After these changes, this is what the user sees when they try to search with more than 1 element:

Screenshot 2025-06-17 at 11 32 30

### DIFF
--- a/src/main/webapp/ui/src/components/Ketcher/KetcherDialog.tsx
+++ b/src/main/webapp/ui/src/components/Ketcher/KetcherDialog.tsx
@@ -12,7 +12,11 @@ import { styled } from "@mui/material/styles";
 import Stack from "@mui/material/Stack";
 import AnalyticsContext from "../../stores/contexts/Analytics";
 import { Ketcher } from "ketcher-core";
-import ValidatingSubmitButton, { IsValid, IsInvalid, ValidationResult } from "../ValidatingSubmitButton";
+import ValidatingSubmitButton, {
+  IsValid,
+  IsInvalid,
+  ValidationResult,
+} from "../ValidatingSubmitButton";
 
 declare global {
   interface Window {
@@ -32,6 +36,7 @@ type KetcherDialogArgs = {
   readOnly?: boolean;
   additionalControls?: React.ReactNode;
   validationResult?: ValidationResult;
+  onChange: () => void;
 };
 
 const StyledDialog = styled(Dialog)(() => ({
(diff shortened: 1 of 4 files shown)


src/main/webapp/ui/src/components/Ketcher/KetcherDialog.tsx
```diff
@@ -30,6 +35,9 @@ type KetcherDialogArgs = {
   actionBtnText?: string;
   readOnly?: boolean;
   additionalControls?: React.ReactNode;
+  validationResult?: ValidationResult;
+  onChange: () => void;
```
rlamacraft, line 39: This should be `onChange?: () => void;` so that its not required in all of the other places that KetcherDialog is used. Could you then update `KetcherDialog.js.flow whilst we're still in the middle of the TypeScript migration with these new props, please: should just be a copy-and-paste once ValidationResult has been imported in the flow declaration file. That way, the JavaScript places that are using KetcherDialog will continue to type check correctly (there isn't a bug because all three new props are not required, but there would be if someone makes them required in the future)

src/main/webapp/ui/src/components/Ketcher/KetcherDialog.tsx
```diff
@@ -72,6 +82,9 @@ const KetcherDialog = ({
         <DialogTitle>{title}</DialogTitle>
         <DialogContent style={{ minHeight: "0" }}>
           <Stack sx={{ height: "100%" }}>
+            {instructionText && (
+                <Typography variant="body2" style={{ paddingBottom: "1em"}}>{instructionText}</Typography>
```
rlamacraft, line 86: Instead of using "em"s as units for padding, we generally just use MUI's built-in mechanism so you can replace the style prop with the simpler `sx={{ pb: 1 }}` (or 2)

rlamacraft (changes requested):
Just a couple of minor things concerning type correctness and style consistency

src/main/webapp/ui/src/components/Ketcher/KetcherDialog.tsx
rs-fraser: Updated `onChange` to be optional, and added these 3 new props to the flow file.

PR #376, opened 2025-06-19
Rsdev 575 add ascenscia

rs-fraser:
## Description ##
This PR adds an Ascenscia app controller, with a `/connect` endpoint.

The user enters their Ascenscia username, password and organization on the apps page entry, and via the back-end an Ascenscia api token is generated and stored in the db. 

This will allow the Ascenscia integration to get the api key by calling `/integration/integrationInfo?name=ASCENSCIA`.

Where there's an error at the Ascenscia end, we try to relay the error as when the password or org is incorrect, Ascenscia includes that information. In other cases, including when the username is wrong, we return a general 500 error as there's no error message to relay from Ascenscia.

There's also validation on the back-end to ensure each field is not blank with appropriate error messages.

In the previous commit [commit](https://github.com/rspace-os/rspace-web/pull/376/commits/2ae626ebc6133ae2a9198f5d9fcd73def7dcd051) there's a button on the apps page UI I used during testing to show the token, which might be useful if testing locally.

I've removed the refresh token code for now as I realised earlier we need to store the Ascenscia username and org to be able to use the refresh endpoint, which I'd overlooked as I was using hardcoded values, so I'll pick that up via a new ticket.   If when using the Ascenscia integration the token stops working, the user should be directed back to the apps page to generate a new one.

## Testing
Flow:
- sysadmin enables deployment property `ascenscia.enabled`
- user enables Ascnescia app on apps page
- user enters Ascenscia username, password and organisation on the apps page
    - these credentials aren't stored by rspace
- `rspace-web` backend calls Ascenscia api and swaps credentials for JWT auth and refresh tokens and stores both in the db
- user is notified auth token was successful, or relays error from Ascenscia api if there's a problem
- the Ascenscia integration in the editor gets the users Ascenscia auth token via `/integration/integrationInfo?name=ASCENSCIA`
- ~~if there's an auth error when using the token, Ascenscia integration calls `rspace-web` `/refresh` endpoint to refresh the token~~ (refresh not implemented yet)
- ~~if the refresh is unsuccessful,~~ Ascenscia integration should tell the user to enter their credentials again and generate a new token via the rspace-web apps page


src/main/webapp/ui/src/eln/apps/CardListing.tsx
```diff
@@ -73,6 +73,16 @@ function CardListing({
...
   );
 
+  const ascensciaUpdate = React.useCallback(
+    (newState: IntegrationStates["ASCENSCIA"]) => {
+      void runInAction(async () => {
+        integrationStates.ASCENSCIA = await update("ASCENSCIA", newState);
+      });
+    },
+    //eslint-disable-next-line react-hooks/exhaustive-deps
+    [update]
```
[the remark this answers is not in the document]
rlamacraft: This should be ignored; applying this suggestion would result in an infinite loop

src/main/webapp/ui/src/eln/apps/integrations/Ascenscia.tsx
```diff
@@ -1,48 +1,242 @@
...
+            message: await response.text(),
+          }),
+        );
+      }
+    } catch (e) {
+      if (e instanceof Error)
+        addAlert(
+          mkAlert({
+            variant: "error",
+            title: "Unable to save Ascenscia API key",
```
rlamacraft, line 145: @rs-fraser These titles should say "Unable to get Ascenscia API key", no? And in any case, just wanted to check that we do need to show the API key and usually we don't

src/main/webapp/ui/src/eln/apps/integrations/Ascenscia.tsx
```diff
@@ -1,48 +1,242 @@
...
+                            }}
+                          />
+                        </Grid>
+                      </Grid>
+                    </CardContent>
+                    <CardActions>
+                      <Button type="submit">Save</Button>
+                      <Button onClick={handleRefresh}>Refresh Token</Button>
+                      <Button onClick={showToken}>Show Token</Button>
+                    </CardActions>
```
rlamacraft, line 233: All these textfields and buttons look fine to me; glad the pattern is simple-enough to modify and extend

src/main/webapp/ui/src/eln/apps/integrations/Ascenscia.tsx
rs-fraser: Sorry I forgot there was already a PR created for this. I've still got some tidying to do. The buttons for refresh and show token here are purely for demo/testing since this will be handled by the actual Ascenscia integration. I'll move this PR to draft and let you know when it's ready to review again.

rs-fraser:
This is ready to review again. @rlamacraft I have 1 failing jest test that I'm unsure of around the apps page - I removed one that was checking Ascenscia was under the 3rd party heading as it no longer was, but there's one failing about the number of headings if you've any ideas? I tried removing the 3rd party one since it might not render if there's nothing under it, but it still fails
```
Summary of all failing tests
 FAIL  src/eln/apps/__tests__/App.test.tsx (11.234 s)
  ● Apps page › Has all of the correct headings.

    There are more actual headings than expected headings.

      72 |
      73 |     // @ts-expect-error assertHeadings comes from assertSemanticHeadings
    > 74 |     expect(container).assertHeadings([
         |                       ^
      75 |       { level: 1, content: "Apps" },
      76 |       { level: 2, content: "Enabled" },
      77 |       { level: 2, content: "Disabled" },

      at Object.<anonymous> (src/eln/apps/__tests__/App.test.tsx:74:23)


```

I'll check back on the java tests later as well.

rlamacraft:
> This is ready to review again. @rlamacraft I have 1 failing jest test that I'm unsure of around the apps page - I removed one that was checking Ascenscia was under the 3rd party heading as it no longer was, but there's one failing about the number of headings if you've any ideas? I tried removing the 3rd party one since it might not render if there's nothing under it, but it still fails
> 
> ```
> Summary of all failing tests
>  FAIL  src/eln/apps/__tests__/App.test.tsx (11.234 s)
>   ● Apps page › Has all of the correct headings.
> 
>     There are more actual headings than expected headings.
> 
>       72 |
>       73 |     // @ts-expect-error assertHeadings comes from assertSemanticHeadings
>     > 74 |     expect(container).assertHeadings([
>          |                       ^
>       75 |       { level: 1, content: "Apps" },
>       76 |       { level: 2, content: "Enabled" },
>       77 |       { level: 2, content: "Disabled" },
> 
>       at Object.<anonymous> (src/eln/apps/__tests__/App.test.tsx:74:23)
> ```
> 
> I'll check back on the java tests later as well.

That test was being skipped before (I don't remember why), and by removing the Ascenscia specific one you effectively added back the one being skipped. I'll investigate why it was being skipped and do something about it

rlamacraft:
> > This is ready to review again. @rlamacraft I have 1 failing jest test that I'm unsure of around the apps page - I removed one that was checking Ascenscia was under the 3rd party heading as it no longer was, but there's one failing about the number of headings if you've any ideas? I tried removing the 3rd party one since it might not render if there's nothing under it, but it still fails
> > ```
> > Summary of all failing tests
> >  FAIL  src/eln/apps/__tests__/App.test.tsx (11.234 s)
> >   ● Apps page › Has all of the correct headings.
> > 
> >     There are more actual headings than expected headings.
> > 
> >       72 |
> >       73 |     // @ts-expect-error assertHeadings comes from assertSemanticHeadings
> >     > 74 |     expect(container).assertHeadings([
> >          |                       ^
> >       75 |       { level: 1, content: "Apps" },
> >       76 |       { level: 2, content: "Enabled" },
> >       77 |       { level: 2, content: "Disabled" },
> > 
> >       at Object.<anonymous> (src/eln/apps/__tests__/App.test.tsx:74:23)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'll check back on the java tests later as well.
> 
> That test was being skipped before (I don't remember why), and by removing the Ascenscia specific one you effectively added back the one being skipped. I'll investigate why it was being skipped and do something about it

I've fixed in the issue in separate branch (PR #412) that should be merged first, but after the release is done

PR #383, opened 2025-06-25
RSDEV-691: Created doc in shared Notebook

rs-nicof:
## Description ##
- Allows to creae a document inside a shared notebook when the user that creates it is not the owner of the notebook

## Testing notes
- AWS instance: https://rsdev-691-create-doc-into-notebook-29.researchspace.com/

### DIFF
--- a/src/main/java/com/researchspace/service/SharingHandler.java
+++ b/src/main/java/com/researchspace/service/SharingHandler.java
@@ -10,6 +10,7 @@ import com.researchspace.model.record.Folder;
 import com.researchspace.model.record.Notebook;
 import com.researchspace.model.views.ServiceOperationResult;
 import com.researchspace.model.views.ServiceOperationResultCollection;
+import java.util.List;
 
 /** Facade to share documents/notebooks */
 public interface SharingHandler {
@@ -19,8 +20,8 @@ public interface SharingHandler {
 
   SharingResult shareRecordsWithResult(ShareConfigCommand shareConfig, User sharer);
 
-  ServiceOperationResultCollection<RecordGroupSharing, RecordGroupSharing> shareIntoSharedFolder(
-      User user, Folder sharedFolder, Long recordId);
+  List<RecordGroupSharing> shareIntoSharedFolderOrNotebook(
+      User user, Folder sharedFolderOrNotebook, Long recordId);
 
   SharingResult moveIntoSharedNotebook(
       Group group, BaseRecord baseRecordToMove, Notebook targetSharedNotebook);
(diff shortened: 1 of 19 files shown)


src/main/java/com/researchspace/service/impl/SharingHandlerImpl.java
```diff
@@ -92,8 +93,25 @@ public ServiceOperationResultCollection<RecordGroupSharing, RecordGroupSharing>
...
+      User user, Folder sharedFolderOrNotebook, Long recordId) {
+    if (folderManager.isSharedFolderOrSharedNotebookWithoutCreatePermssison(
+        user, sharedFolderOrNotebook)) {
+      ServiceOperationResultCollection<RecordGroupSharing, RecordGroupSharing> sharingResult;
+      if (sharedFolderOrNotebook.isNotebook()) {
+        // shareDocument into the current parent sharedNotebook
+        sharingResult =
+            this.shareIntoSharedNotebook(user, (Notebook) sharedFolderOrNotebook, recordId);
+      } else {
+        // shareDocument into the current parentFolder
```
mKowalski256, line 106: again, a bit redundant comment

src/main/java/com/researchspace/service/impl/RecordManagerImpl.java
```diff
@@ -335,7 +335,9 @@ private StructuredDocument doCreateDocument(
       parentFolder = folderDao.get(parentId);
     }
 
-    if (!parentFolder.isSharedFolder()) {
+    if (!parentFolder.isShared()
+        || (parentFolder.isNotebook()
+            && permissnUtils.isPermitted(parentFolder, PermissionType.CREATE, user))) {
```
mKowalski256, line 340: I believe this could be replaced with a call to newly added method `folderManager.isSharedFolderOrSharedNotebookWithoutCreatePermssison()`
(and if/else could be reversed here, so the code is even clearer)

src/main/java/com/researchspace/service/SharingHandler.java
```diff
@@ -19,8 +20,8 @@ ServiceOperationResultCollection<RecordGroupSharing, RecordGroupSharing> shareRe
 
   SharingResult shareRecordsWithResult(ShareConfigCommand shareConfig, User sharer);
 
-  ServiceOperationResultCollection<RecordGroupSharing, RecordGroupSharing> shareIntoSharedFolder(
-      User user, Folder sharedFolder, Long recordId);
+  List<RecordGroupSharing> shareIntoSharedFolderOrNotebook(
+      User user, Folder sharedNotebook, Long recordId);
```
mKowalski256, line 24: argument should rather be named `sharedFolderOrNotebook`

mKowalski256 (approved):
Looks good to me & works fine on [AWS test buid](rsdev-691-create-doc-into-notebook-29.researchspace.com).

src/main/java/com/researchspace/service/impl/RecordManagerImpl.java
rs-nicof: they seemed to be 2 different condiotions, but after or call/brain storming we figured out that could be unified.

I've moved the condition into the `RecordManager` so that can be reached by the `SharingHandler` as well 😎

PR #384, opened 2025-06-25
RSDEV-706: when moving/copying to irods, ensure the absolute path is used

mKowalski256:
## Description ##
The PR fixes problem with 'move to iRODS' action executed on filestores created through a new Gallery UI. 

The problem is that iRODS jargon library action for uploading files requires absolute file path to the destination location on target iRODS server. The other jargon library actions, e.g. ones used for browsing the content of iRODS filestore, are fine with relative file paths ('relative' here means relative to irods home directory, saved on filesystem details page). 

It turns out that iRODS filestores saved by the new Gallery UI are saving relative filestore paths (this seems like a change of the behaviour from old Gallery, which we didn't notice). The relative filestore paths should be fine though, we do use them for samba/sftp filestores, it's just that when the absolute path is needed, it needs to be calculated. The change in the PR is therefore to make sure the destination path used for 'move to iRODS' action is always an absolute path, i.e. it starts with irods home directory fragment.

### DIFF
--- a/src/main/java/com/researchspace/api/v1/controller/GalleryIrodsApiController.java
+++ b/src/main/java/com/researchspace/api/v1/controller/GalleryIrodsApiController.java
@@ -220,11 +220,11 @@ public class GalleryIrodsApiController extends GalleryFilestoresBaseApiControlle
     String irodsHomePath =
         nfsFileStore.getFileSystem().getClientOption(NfsFileSystemOption.IRODS_HOME_DIR);
     String filestorePath = nfsFileStore.getPath();
-    if (StringUtils.isEmpty(irodsHomePath)) {
-      return filestorePath;
+    if (StringUtils.isBlank(irodsHomePath)) {
+      irodsHomePath = "";
     }
-    if (StringUtils.isEmpty(filestorePath)) {
-      return irodsHomePath;
+    if (StringUtils.isBlank(filestorePath)) {
+      filestorePath = "";
     }
     if (filestorePath.startsWith(irodsHomePath)) {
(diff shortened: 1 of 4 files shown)


src/main/java/com/researchspace/api/v1/controller/GalleryIrodsApiController.java
```diff
@@ -213,6 +216,25 @@ private ApiExternalStorageOperationResult performCopyToIRODS(
...
+  protected String getAbsoluteFilestorePathForIrods(NfsFileStore nfsFileStore) {
+    String irodsHomePath =
+        nfsFileStore.getFileSystem().getClientOption(NfsFileSystemOption.IRODS_HOME_DIR);
+    String filestorePath = nfsFileStore.getPath();
+    if (StringUtils.isEmpty(irodsHomePath)) {
+      return filestorePath;
+    }
+    if (StringUtils.isEmpty(filestorePath)) {
+      return irodsHomePath;
+    }
```
rs-nicof, line 228: I think you could avoid those lines caue if they are empty then they would fall into the case `irodsHomePath + filestorePath`

rs-nicof, line 228: I think you can delete those lines since it should fall into the return case `irodsHomePath + filestorePath;`

mKowalski256: the separate cases were to handle potential `null` values, but I've updated the code to explicitly convert nulls into empty strings.

PR #391, opened 2025-06-30
RSDEV-644 messaging in app bar

rlamacraft:
## Description ##
Adds a badged icon to the app bar to show the sum of notifications and messages, allowing users to view the unread account across the entire product. The icon button, when pressed, navigates the user to the messaging page.

This change also adds badge counts to the existing buttons on the messaging page so that it is clear that there are unread messages even if there are no unread notifications

### DIFF
--- a/src/main/webapp/WEB-INF/pages/dashboard/dashboard.jsp
+++ b/src/main/webapp/WEB-INF/pages/dashboard/dashboard.jsp
@@ -38,13 +38,13 @@
 		<a class="dashboardOptions" id="mynotifications" href="#">
 		    <spring:message code="messaging.newNotificationsTitle"/>
            	<c:if test="${not empty notificationList}">
-                <span class="badge">${fn:length(notificationList)}</span>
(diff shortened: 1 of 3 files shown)


rlamacraft:
> The only minor issue I found is that the badge counts displayed on Dashboard page seem to be applied on initial page load, and are not updated afterwards. So it's easy for them to become out of sync when user does something on that page (e.g. mark notification as deleted). But I understand we may want to wait with fix until whole page is reworked. image

[Vaida said the same](https://researchspace.atlassian.net/browse/RSDEV-644?focusedCommentId=63528), and whilst we could look at trying to keep that badge up to date there is a question of whether it would be worth the effort. Perhaps just having it clear when the users marks as read would suffice, given that would be 90% of the time that the bug shows itself?

mKowalski256:
> Perhaps just having it clear when the users marks as read would suffice, given that would be 90% of the time that the bug shows itself?

Yes, sounds good to me!

PR #392, opened 2025-06-30
RSDEV-729: store temporary notebook data on page as text, rather than html

mKowalski256:
## Description ##
The PR changes the way how temporary notebook data is stored on a page during loading of the notebook view. Previously this data was stored within `<div>` element as html, and as a consequence various js/react code selectors were matching the temporary data fragment. That was causing front-end to make duplicated calls e.g. to `/listOfMaterials/forField/` and `/chemical/ajax/getInfo` endpoints.

Now that temporary notebook data is stored as text, it no longer triggers any js/react code.


mKowalski256:
Can be tested at https://rsdev-729-notebook-calls-3.researchspace.com/.

Issue #232, opened 2025-01-15
[IDEA] Align API with RSpace UI for retrieving audit trail


tilorspace:
### Summary

The Rspace ELN API has a GET /activities endpoint that allows users to query the audit trail, but activities of USER type are not included, causing a discrepancy between results obtained through UI and API.

This may be caused by /activities endpoint showing a limited ‘domains’ list that really does display event types (not domains), but without including USER type. This ‘domains’ are also not aligned with UI ‘domains’.

(related JIRA ticket: RSPAC-2873)

### Motivation

Enable efficient automation of audit trail related queries via the API

### User Story

As an API user I should be able to retrieve the same audit report through API as I see in the UI.

### Must Have

- ELN API /activities endpoint must return USER type events by default

### Should Have

- ELN API /activities endpoint should allow returning events of just USER type
- ELN API /activities endpoint should support same domains as UI (i.e. ELN and Inventory); if filtering by types is already implemented that should stay as a useful additional option;

tilorspace:
thanks for checking this one out @richarda23 . Feel free to raise a PR for the documentation. For the UI, I'd suggest to hold off for now, since we're in the process of revising the frontend code across RSpace.

PR #398, opened 2025-07-03
ISSUE-232: review audit 'domains' filter on Auditing page

mKowalski256:
## Description ##
My RSpace -> Auditing page in the UI currently allows filtering results by two "domain" checkboxes: `Inventory` and `ELN` . The label "domain" is used in common-sense way here, to describe the general areas of RSpace where the audit events come from.

image

At the same time "audit domain" is also a technical term within RSpace, as there is a granular list of possible audit domains used by back-end, which are also exposed through RSpace API. The `ELN` and `Inventory` checkboxes in the UI are really just groupings over these more granular audit domains, e.g. `Inventory` checkbox in the UI limits results to ones from `INV_CONTAINER`,`INV_SAMPLE`,`INV_SUBSAMPLE` domains. 

I think the confusion between how Audit UI and RSpace API are using the term "domain" was a cause behind #232. Therefore this PR changes the label used on RSpace -> Auditing UI page from "domains" to "activity areas". Additionally the PR regroups "activity areas" into three checkboxes `ELN`, `Inventory` and `Other`. All the tree checkboxes of are selected by default, so user gets all the events by default.

image

Finally, [per other comment](https://github.com/rspace-os/rspace-web/issues/232#issuecomment-3028455956), a few audit domains (`COMMUNITY`, `FORM`, `GROUP`, `MEDIA`, `MESSAGING`) were actually missing from the report generated by the UI, that's due to `Inventory`/`ELN` checkboxes not including these domains. The PR fixes that, so the audit results retrieved in the UI will be in line with ones retrievable by API.


mKowalski256:
@tilorspace marked you as a reviewer, but just on functional side ;) 
The test build is running at https://issue-232-audit-ui-6.researchspace.com/.

PR #399, opened 2025-07-03
RSDEV-681 Import from PubChem

rlamacraft:
## Description ##
Chemistry files can be inserted into documents, importing them from PubChem. The new files are read-only (just like drag-and-drop) and are added to the Gallery.

image


rs-fraser:
> > @rlamacraft let me know when you're happy to merge, as I need to update the version number in `rspace-core-model` and align in this repo since we are using a feature branch build
> 
> I don't have any further changes, though it's probably worth spinning up one last build with the merge conflicts resolved. The Ketcher one is simple enough but I can't figure out what's up with the other one

Sounds good, I've fixed the conflicts and will run locally then spin up an AWS build that we can check before merging.

PR #405, opened 2025-07-04
RSDEV-191: switch from ROR API v1 to v2

mKowalski256:
## Description ##
The code handling ROR API v2 responses was already there, the PR changes the default deployment property and removes handling for ROR API v1 responses, as these will no longer be necessary.

Additionally I've noticed that we're using ROR ID of some random (?) university for unit tests, and also as a placeholder in the UI. I've updated it to ROR ID of Research Space (`https://ror.org/038xqyz77`), my guess is we didn't have official ROR when developing the original implementation?


mKowalski256:
The test build is configured & running at https://rsdev-191-ror-update-5.researchspace.com.

PR #406, opened 2025-07-07
Rpd 191 av portal embed

mKowalski256:
## Description ##
- Updated TextFieldEmbedIframeHandler.java to recognise embed code from TIB's AV-portal as allowed content in RSpace documents; plus unit test
- removed an obsolete jove-test URL


tilorspace:
thanks! definitely room for improvement, but basic functionality is ok.

PR #410, opened 2025-07-09
RSDEV-957: Fix Permission issue when creating from Word (or creating from protocols.io) in a shared folder/notebook

rs-nicof:
## Description ##
When creatng from Word/Evernote (or protocols.io) the process was failing for a ermisisons issue
In this PR we have extended the `create` permissions in those cases to allow the user to create also if is a SharedFolder or a SharedNotebook

## Testing notes
To test you can create from Word/Evernote or protocols.io from inside a shared notebook made by another user of the same group
- AWS instance: https://prt-957-create-from-word-16.researchspace.com
-  - you need to configure `protocolsio.client.id` and `protocolsio.secret` in the deployment.properties

### DIFF
--- a/src/main/java/com/researchspace/linkedelements/TextFieldEmbedIframeHandler.java
+++ b/src/main/java/com/researchspace/linkedelements/TextFieldEmbedIframeHandler.java
@@ -32,7 +32,7 @@ public class TextFieldEmbedIframeHandler {
   private static final String JOVE_SRC_SUFFIX_PATTERN =
       "\\?id=\\d+((&\\w+=1)|(&language=\\w+)|(&access=\\w+)|(&utm_source=\\w+))*";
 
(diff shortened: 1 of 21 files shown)


src/main/resources/deployments/dev/deployment.properties
```diff
@@ -183,3 +183,6 @@ irods.realConnection.server.password=
 #chemistry.service.url=http://localhost:8090
 #chemistry.service.indexOnStartup=true
 #chemistry.service.reGenerateOpenSourceImages=true
+
+protocolsio.client.id=pr_live_id_6850816a7182b4064ea49f9049dc664f
+protocolsio.secret=pr_live_sc_07f1732db7dd2b14c5dd52082dd26fa5
```
rs-nicof, line 188: Remove those!

PR #418, opened 2025-07-15
RSDEV-520: Extend Fieldmark import with IGSN ID

rs-nicof:
## Description ##
- This PR extends the functionality "Import from Fieldmark"
- - From now when importing from fieldmark the system will analyse all the rows and returns a list of candidate fields that can be used as DOI identifiers. Finally the user will be able select one of the candidate fields to be treated as IGSN DOI Identifier. Finally, when importing the fieldmark notebook into the inventory, the Samples will already have the DOI linked as specified on the Fieldmark UI.

## Pre requisites to merge ##
 - merge the fieldmark repo PR: https://github.com/rspace-os/fieldmark-java-client/pull/4

## Testing notes
- I have alreadsy prefilled the "`Description`" field into Fieldmark UI to have DOIs.
- If you want to test on an instance you need to be sure that those DOIs are already generated on the RSpace instance and are not assigned to any inventory object already.
- This PR is **ONLY BACKEND** so in order to test it you need to go to the swagger page of the Inventory: https://your_instance/public/apiDocs?urls.primaryName=RSpace%20Inventory
- - **The UI for this feature** will be implemented on another PR referencing the folliowing JIRA: https://researchspace.atlassian.net/browse/RSDEV-746
- **AWS instance** spinned: https://rsdev-520-backend-25.researchspace.com 

### List Fieldmark fields (of a given notebook) that could be imported as IGSN ###
**WHEN**
1.  Enable fieldmark, auth access from the configuration
2. Add Fieldmark bearer token and enable Fieldmark banner from the App page
3. Enable DataCite from the Inventory setting section
4. Create a bunch of new IGSN to allocate (5 would be more than enough)
5. Go to Fieldmark UI and fill the "`Description`" field of all the 3 observations with the IGSNs created on fieldmark
6. go to the swagger UI and login
7. Run the end point `GET /fieldmark/notebooks/igsn/{notebookId}` giving your notebookId as parameter

**THEN** 
 - You need to have as result a list of fieldName (in this case should only be "`Description`"):
```
[ "Description" ]
```

### Import a notebook from Fieldmark into the RSpace Inventory (including IGSN DOIs)  ###
**WHEN**
1. go to the swagger UI and login
2. Run the end point `POST /import​/fieldmark​/notebook` giving your notebookId and the IGSN identifier field as parameters in the JSON body:
```
{
  "notebookId": "1726126204618-rspace-igsn-demo",
  "identifier": "Description"
}
```
**THEN** 
 - You need to have as result the importResult JSON with the IDs of the container, sample template and samples global IDs
 - On the inventory UI you can see all the samples just created, and each one of them has already the Identifier DOI corectly linked

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1861,7 +1861,13 @@
     <dependency>
         <groupId>com.github.rspace-os</groupId>
         <artifactId>fieldmark-java-client</artifactId>
-        <version>1.1.0</version>
+        <version>RSDEV-520-client-SNAPSHOT</version>
+      <exclusions>
+        <exclusion>
+          <groupId>commons-validator</groupId>
+          <artifactId>commons-validator</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
         <groupId>com.github.rspace-os</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -1861,7 +1861,7 @@
     <dependency>
         <groupId>com.github.rspace-os</groupId>
         <artifactId>fieldmark-java-client</artifactId>
-        <version>RSDEV-520-client-SNAPSHOT</version>
+        <version>2.0.0</version>
       <exclusions>
         <exclusion>
           <groupId>commons-validator</groupId>
--- a/src/main/java/com/researchspace/api/v1/FieldmarkApi.java
+++ b/src/main/java/com/researchspace/api/v1/FieldmarkApi.java
@@ -13,10 +13,10 @@ import org.springframework.http.MediaType;
 import org.springframework.validation.BindException;
 import org.springframework.validation.BindingResult;
 import org.springframework.web.bind.annotation.GetMapping;
-import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.RequestAttribute;
 import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.ResponseStatus;
 
 @RequestMapping("/api/inventory/v1")
@@ -47,10 +47,11 @@ public interface FieldmarkApi {
    * @throws MalformedURLException
    * @throws URISyntaxException
    */
-  @GetMapping(value = "/fieldmark/notebooks/igsn/{notebookId}")
+  @GetMapping(value = "/fieldmark/notebooks/igsnCandidateFields")
   @ResponseStatus(HttpStatus.OK)
   List<String> getIgsnCandidateFields(
-      @PathVariable("notebookId") String notebookId, @RequestAttribute(name = "user") User user)
+      @RequestParam(name = "notebookId") String notebookId,
+      @RequestAttribute(name = "user") User user)
       throws BindException;
 
   /***
--- a/src/main/java/com/researchspace/api/v1/FieldmarkApi.java
+++ b/src/main/java/com/researchspace/api/v1/FieldmarkApi.java
@@ -7,12 +7,15 @@ import com.researchspace.webapp.integrations.fieldmark.FieldmarkApiImportResult;
 import java.net.MalformedURLException;
 import java.net.URISyntaxException;
 import java.util.List;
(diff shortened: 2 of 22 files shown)


pom.xml
```diff
@@ -1861,7 +1861,13 @@
     <dependency>
         <groupId>com.github.rspace-os</groupId>
         <artifactId>fieldmark-java-client</artifactId>
-        <version>1.1.0</version>
+        <version>RSDEV-520-client-SNAPSHOT</version>
```
rs-nicof, line 1864: this will be `2.0.0`

src/main/java/com/researchspace/webapp/integrations/fieldmark/FieldmarkApiController.java
```diff
@@ -96,73 +117,96 @@ public FieldmarkApiImportResult importNotebook(
...
-              + "\" due to the following error: "
-              + clientEx.getMessage());
+          "", "errors.fieldmark.import", new Object[] {}, "ERROR: " + e.getMessage());
     } finally {
       throwBindExceptionIfErrors(errors);
     }
+    return importResult;
+  }
 
+  //TODO[nik]: remove this
```
rs-nicof, line 166: reminder:  revmore this method (it was the old controller code before I moved it to the Fieldmark Service)

src/main/java/com/researchspace/api/v1/FieldmarkApi.java
```diff
@@ -34,6 +37,22 @@ public interface FieldmarkApi {
...
+   * End point that gives the list of fields of the notebook that can be imported as identifiers
+   *
+   * @param notebookId
+   * @param user
+   * @return a list of fields that can be imported as Igsn
+   * @throws BindException
+   * @throws MalformedURLException
+   * @throws URISyntaxException
+   */
+  @GetMapping(value = "/fieldmark/notebooks/igsn/{notebookId}")
```
mKowalski256, line 50: I'd change the order of path fragments for the endpoint url configured by this `@GetMapping`.

The convention in RSpace API is that the endpoint paths are constructed in format: `/<entity>/{entityIdParam}/<action>`. So in this case, the endpoint should rather be:
`/fieldmark/notebooks/{notebookId}/igsn` 
... or maybe even
`/fieldmark/notebooks/{notebookId}/igsnCandidateFields`
...  which is more descriptive

----

Also, can we rely that `notebookId` is always url-friendly, i.e. encodable into url path fragment? If not, maybe it should be passed as url-encoded `GET` parameter.

src/main/java/com/researchspace/service/fieldmark/impl/FieldmarkServiceManagerImpl.java
```diff
@@ -0,0 +1,329 @@
...
+              + importRequest.getNotebookId()
+              + "\" has not being imported from Fieldmark "
+              + "cause to the following error: "
+              + ioEx.getMessage());
+      throw new FieldmarkImportException(
+          "The notebookID \""
+              + importRequest.getNotebookId()
+              + "\" has not being imported from Fieldmark "
+              + "cause to the following error: "
+              + ioEx.getMessage());
```
mKowalski256, line 314: the code would be a bit cleaner if error message was constructed once, and then just used twice.
(same with lines below).

mKowalski256 (approved):
Added a couple of comments, but looks good & works for me on test AWS build.

src/main/java/com/researchspace/api/v1/FieldmarkApi.java
rs-nicof: Sure that makes sense.

I'll change it to the `GET /fieldmark/notebooks/igsnCandidateFields?notebookId=paramEncodedValue`

src/main/java/com/researchspace/service/fieldmark/impl/FieldmarkServiceManagerImpl.java
rs-nicof: good point

src/main/java/com/researchspace/api/v1/FieldmarkApi.java
rs-nicof: done

PR #420, opened 2025-07-16
Fix stale UI state bug on apps page

rlamacraft:
## Description ##
If the user updated the state associated with an integration (e.g. the API key field) and then triggered a second call to `update` (such as by enabling or disabling the UI state) then the original value could be overwritten because the UI was caching the state and not flushing that state after the first update.

## Testing notes
The easiest integration to test is Zenodo as it just has an API key as state, and any string value can be stored in the database


nhanlon2:
tested by doing the same change with Galaxy and its fixed

nhanlon2 (approved):

PR #430, opened 2025-07-21
RSDEV-761 Remove useEffect dependencies that trigger excess network calls

rlamacraft:
## Description ##
Regression introduced by the TypeScript migration that caused some network calls to be made twice. Doesn't appear that there was any bug visible to the user, just inefficient network usage


mKowalski256 (approved):

PR #433, opened 2025-07-24
Rsdev 639 galaxy services and UI

nhanlon2:
## Description ##
***REQUIRED***

## Design decisions
***OPTIONAL***

## Testing notes
***OPTIONAL***


src/main/webapp/ui/src/assets/branding/galaxy/index.js.flow
```diff
@@ -0,0 +1,8 @@
+// @flow strict
```
rlamacraft, line 1: No need for this file any more, it can go

src/main/webapp/ui/src/tinyMCE/galaxy/Galaxy.tsx
```diff
@@ -0,0 +1,253 @@
...
+              {errorReason !== ErrorReason.None && (
+                  <ErrorView
+                      errorReason={errorReason}
+                      errorMessage={errorMessage}
+                      WorkFlowIcon={WorkFlowIcon}
+                  />
+              )}
+              <Modal
+                  open={uploading}
+                  aria-labelledby="uploading data to galaxy is in progress"
```
rlamacraft, line 189: Shouldn't this be `aria-label`, since `aria-labelledby` takes an id? Also, is the `aria-describedby` needed, I'm not sure when that information is actually used by the screen reader

src/main/webapp/ui/src/tinyMCE/galaxy/Galaxy.tsx
```diff
@@ -0,0 +1,253 @@
...
+        window.parent.postMessage(
+            {
+              mceAction:  "no-data-selected"
+            },
+            "*"
+        );
+      }
+    }
+    doTheUpload();
+  }, [doUpload]);
```
rlamacraft, line 109: I don't see what the point of the `doUpload` flag, being set by the tinymce event handler and then triggering this call is -- why not just have the event handle invoke this logic directly?

src/main/webapp/ui/src/eln/eln-external-workflows/ExternalWorkflowInvocations.tsx
```diff
@@ -0,0 +1,229 @@
...
+      setShouldCheckForGalaxyData(false);
+      //this is a workaround because saving the data is async we cant use
+      // a simple check for the existence of the data
+    }
+  });
+
+  return (
+      <>
+        {errorReason !== ErrorReason.None && (
+            // <div className={classes.launcherWrapper}>
```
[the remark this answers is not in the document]
nhanlon2: done

PR #434, opened 2025-07-25
RSDEV-746 UI for importing IGSN IDs with fieldmark

rlamacraft:


src/main/java/com/researchspace/api/v1/FieldmarkApi.java
```diff
@@ -7,13 +7,16 @@
 import java.net.MalformedURLException;
```
rs-nicof, line 7: you need to rebase form main casie it lokos like the PR has my commits

rs-nicof (changes requested):

PR #438, opened 2025-07-30
RSDEV-701 Add About RSpace dialog and page

rlamacraft:
## Description ##
This change adds a new "About RSpace" section that is both added to the account menu in the app bar as a dialog and in its own standalone "/about" page (which isn't linked anywhere but may at times be helpful for providing support).

image

## Testing notes
Either go to "/about", or click on the "About RSpace" menu item in the app bar's account menu.


nhanlon2:
> > You may want to also add the new properties to the PropertyHolder class, I think its optional if they will only ever be used in pure React components with no JSP involvement.
> 
> [Like so?](https://github.com/rspace-os/rspace-web/pull/438/commits/cafc3d606b90572f5c3e7714a958d8162ef6add5)

Yes!

PR #443, opened 2025-08-08
RSDEV-775: allow direct move of docs into owned shared notebook 

mKowalski256:
Previously the check in code was assuming that if notebook is shared, the move action should try sharing into notebook. But that's not true if user has permission to directly add things into notebook (e.g. they are a notebook owner), in that case we should do normal move.

The actual code fix is just one line in `WorkspaceController` (well, plus a new method in `RecordManager` and a unit test). The rest of changeset for this PR comes from fixed typo in adjacent method name.

### DIFF
--- a/src/main/java/com/researchspace/service/RecordManager.java
+++ b/src/main/java/com/researchspace/service/RecordManager.java
@@ -646,5 +646,7 @@ public interface RecordManager {
    */
   boolean forceMoveDocumentToOwnerWorkspace(StructuredDocument userDoc);
 
-  boolean isSharedFolderOrSharedNotebookWithoutCreatePermssion(User user, Folder parentFolder);
+  boolean isSharedNotebookWithoutCreatePermission(User user, Folder folder);
+
+  boolean isSharedFolderOrSharedNotebookWithoutCreatePermission(User user, Folder folder);
 }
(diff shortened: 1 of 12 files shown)


src/main/java/com/researchspace/service/RecordManager.java
```diff
@@ -646,5 +646,7 @@ List<StructuredDocument> getontologyDocumentsCreatedInPastThirtyMinutesByCurrent
    */
   boolean forceMoveDocumentToOwnerWorkspace(StructuredDocument userDoc);
 
-  boolean isSharedFolderOrSharedNotebookWithoutCreatePermssion(User user, Folder parentFolder);
+  boolean isSharedNotebookWithoutCreatePermission(User user, Folder folder);
+
+  boolean isSharedFolderOrSharedNotebookWithoutCreatePermission(User user, Folder folder);
```
mKowalski256, line 651: fixed a typo here (last word in method name is `Permission` rather than `Permssion`), which triggered some other changes in the PR.

src/main/java/com/researchspace/webapp/controller/WorkspaceController.java
```diff
@@ -609,7 +609,7 @@ public ModelAndView move(
         moveResult = folderManager.move(recordIdToMove, target.getId(), sourceFolder.getId(), user);
       } else {
         BaseRecord baseRecordToMove = recordManager.get(recordIdToMove);
-        if (target.isNotebook() && target.isShared()) {
+        if (recordManager.isSharedNotebookWithoutCreatePermission(user, target)) {
```
mKowalski256, line 612: this is the actual fix

PR #444, opened 2025-08-08
Organise custom react hooks

rlamacraft:
## Description ##
This change organises general purposes custom react hooks that were scattered about the codebase into a more organised structure
- /hooks/api -- for hooks that wrap calls to the JSON APIs
- /hooks/auth -- for hooks that are used in authentication with any APIs
- /hooks/browser -- for hooks that wrap browser APIs
- /hooks/ui -- for hooks that aid with implementing UI behaviour
- /hooks/websockets -- for hooks that wrap calls to websocket APIs

There are, unfortunately, a whole bunch of eslint fixes that have been picked up by these changes


rlamacraft:
> Would it be worth running eslint against all front-end files as a separate task, then future changesets will be cleaner?

Probably, I think they came about when the LLM tools did a lot of the Flow->TS migration without running prettier

PR #446, opened 2025-08-13
Rsdev 639 user configures galaxy server

nhanlon2:
## Description ##
***REQUIRED***
Galaxy Integration can now connect to multiple servers configured per user using their API key(s)

## Design decisions
***OPTIONAL***

## Testing notes
***OPTIONAL***


src/main/java/com/researchspace/webapp/integrations/galaxy/GalaxyController.java
```diff
@@ -27,7 +39,45 @@
...
+      this.alias = alias;
+      this.url = url;
+    }
+
+    @JsonInclude(value = Include.NON_EMPTY)
+    private String token;
+  }
+
   private final UserManager userManager;
+  @Getter private List<GalaxyAliasToServer> aliasServerPairs;
```
rs-fraser, line 69: This field and the code to populate it might be better moved to the GalaxyService, then the IntegrationsHandler wouldn't need to depend on the GalaxyController, which breaks separation of concerns a bit.

src/main/resources/sqlUpdates/changeLog-rsdev-639b.xml
```diff
@@ -0,0 +1,33 @@
...
+            <column name="type" type="NUMERIC" value="2" />
+        </insert>
+        <insert tableName="PropertyDescriptor">
+            <column name="id" type="NUMERIC" value="NULL"/>
+            <column name="defaultValue" type="STRING" value="" />
+            <column name="name" type="STRING" value="GALAXY_ALIAS" />
+            <column name="type" type="NUMERIC" value="2" />
+        </insert>
+    </changeSet>
+    <changeSet id="16-01-2025b" author="nico">
```
rs-fraser, line 23: date (and author) looks wrong here, not sure if there are any repercussions in terms of ordering but I'm assuming it should be the same date as the other change set.

src/main/resources/sqlUpdates/liquibase-master.xml
```diff
@@ -147,6 +147,7 @@
     <include relativeToChangelogFile="true" file="changeLog-rsdev-664.xml"/>
     <include relativeToChangelogFile="true" file="changeLog-rsdev-695.xml"/>
     <include relativeToChangelogFile="true" file="changeLog-rsdev-639.xml"/>
+    <include relativeToChangelogFile="true" file="changeLog-rsdev-639b.xml"/>
```
rs-fraser, line 150: I'm not sure how liquibase ordering works, but is there any reason not to have this at the end of the file as the most recent migration? The 575 one below will already have been applied to servers with the most recent release.

src/main/webapp/ui/src/eln/apps/integrations/Galaxy.tsx
```diff
@@ -40,47 +62,192 @@ function Galaxy({ integrationState, update }: GalaxyArgs): React.ReactNode {
           <>
             <ol>
               <li>
-                Obtain an API Key from Galaxy by going into User → Preferences → Manage API key.
+                Request a user access token by going to username → Preferences →
+                'Manage API key' in Galaxy.
+              </li>
+              <li>Chose the corresponding server from the add menu below.</li>
```
rs-fraser, line 68: Should this be `Choose`?

rs-fraser (approved):

PR #447, opened 2025-08-13
RSDEV-669: allow signup of SSO users with known suffix with RSpace usernames having replacement suffix

mKowalski256:
## Description ##

This PR adds two new deployment properties that allow modifying RSpace username used during signup process for SSO user. If incoming SSO username has a configured suffix, then RSpace username has that suffix replaced with other configured suffix, while original SSO username is saved as `usernameAlias`. 

That change is specifically to allow registering of RSpace users in SSO environment where SSO username adds a fixed known suffix to each username. By configuring `suffixToReplace` and `suffixReplacement` properties, RSpace allows SSO users register with shorter usernames. The original SSO usernames are saved in `usernameAlias` field, which means the login flow is still operating on SSO usernames.

The PR also extends `usernameAlias` database column from 50 -> 255 chars, long SSO usernames can be registered as aliases.

Finally, the PR slightly modifies the 'signup' page, to display an explanatory message when user who signs up in SSO mode is not allowed to register as a PI.

## Testing notes
Test steps added at the end of [RSTEST-396](https://researchspace.atlassian.net/browse/RSTEST-396) scenario.

### DIFF
--- a/DevDocs/DeveloperNotes/GettingStarted/GettingStarted.md
+++ b/DevDocs/DeveloperNotes/GettingStarted/GettingStarted.md
@@ -250,8 +250,8 @@ following to your mvn launch command:
 
 Enterprise Standalone: this is the default, no extra configuration is needed.
 
-Enterprise Institutional SSO: `-Ddeployment.sso.type=TEST -Dmock.remote.username=user -Ddeployment.standalone=false`
-(where the username is for a user already in the database)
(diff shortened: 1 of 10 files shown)


rs-fraser (approved):
Looks good. Tested following https://researchspace.atlassian.net/browse/RSTEST-396 and all works.

PR #454, opened 2025-08-25
RSDEV-794: Raise more readable error when DMPTool returns empty dmp list

rs-nicof:
## Description ##
This PR fixed the erro message when DMPTool server returns an empty list of plans

### DIFF
--- a/src/main/java/com/researchspace/webapp/integrations/dmptool/DMPToolDMPProviderImpl.java
+++ b/src/main/java/com/researchspace/webapp/integrations/dmptool/DMPToolDMPProviderImpl.java
@@ -27,8 +27,10 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpEntity;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpMethod;
+import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
+import org.springframework.web.client.HttpClientErrorException;
 import org.springframework.web.client.RestClientException;
 import org.springframework.web.client.RestTemplate;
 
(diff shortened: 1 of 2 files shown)


mKowalski256 (commented):
I've started test build at https://rsdev-794-dmp-list-empty-8.researchspace.com, and configured dmptool account there. I've also deleted existing DMPs from test account  `DMPTool Ramon's New Test Account (For ERROR))` -  I believe the reason for this account is that we can reproduce the issue? (we have another `operations@researchspace.com` account for testing valid DMPs).

The new message displayed fine, but 'unable to load DMPs' wording still looks very generic to me:
image

I think it could either suggest the reason ("Do you have any completed, non-test DMPs?"), or maybe the helpdocs link could scroll to the 'Importing DMPs into RSpace' subsection, otherwise I need to read/scroll through the most of the article to find out warning section telling me about possible problem.

mKowalski256:
> We can link to a particular section of the help doc page, but it has to be a header, so the page would have to be re-written to make the issue its own section with a sub-title rather than just an alert

Yes, the section 'Importing DMPs into RSpace' is actually [linkable](https://researchspace.helpdocs.io/article/o0wlhlgxnr#importing_dm_ps_into_r_space), the link could redirect there.

rlamacraft:
> > We can link to a particular section of the help doc page, but it has to be a header, so the page would have to be re-written to make the issue its own section with a sub-title rather than just an alert
> 
> Yes, the section 'Importing DMPs into RSpace' is actually [linkable](https://researchspace.helpdocs.io/article/o0wlhlgxnr#importing_dm_ps_into_r_space), the error link could redirect there.

done

mKowalski256 (approved):
Retested at https://rsdev-794-dmp-list-empty-10.researchspace.com, looks good - the error popup link now redirect to the article section which explains the problem.

PR #457, opened 2025-08-27
Rsdev 689 stoichiometry backend

rs-fraser:
## Description ##
This PR contains the backend changes for Stoichiometry including CRUD operations, DB migration to create the new tables, code to copy Stoichiometry as part of a document copy, and changes to get additional info from the new `chemistry` service endpoint. 

Requires the latest version of the `chemistry` service to run.

## Testing notes
The branch `rsdev-689-stoichiometry-ui` contains these back-end changes plus the UI which could be used for testing purposes. We are holding back on the UI PR until the UI branch has been further tested by chemists. 

Otherwise, swagger specs have been added for the Stoichiometry endpoint. To use them, create a document and add a chemical element. Check the DB or network tab to get the `RsChemElement.id` and use that as the `chemElementId` when creating a Stoichiometry. 

related `rspace-core-model` PR: https://github.com/rspace-os/rspace-core-model/pull/33


mKowalski256 (commented):
I've commented on a couple of things, but maybe the most significant bit I see missing on backend implementation is the revisioning of stoichiometry data. I think stoichiometry data needs to be revisioned as it is editable, and when browsing revision history of the ELN document user will expect to see the correct historical version of the stoichiometry, from the time of document revision.

To add revisioning, model classes need to be marked with `@Audited` annotations, the `_AUD` tables need to be created on db, and maybe some other things. 

To demonstrate the problem on current https://rsdev-689-stoichiometry-ui-31.researchspace.com build: if I create a stoichiometry table for the chemical, then make a few separate edits updating the stoichiometry, then check revision history of the document, then all revisions will show the latest state of the stoichiometry table. If I subsequently delete the stoichiometry table from latest version of the document, the historical versions that used to have the stoichiometry will show an error popup:
image

mKowalski256 (approved):
Looks good & works for me on the test build.

PR #462, opened 2025-09-02
RSDEV-796: block PI/LA accidentally moving content out of Shared folder

mKowalski256:
## Description ##
Switch to rspace-core-model version from [PR#35](https://github.com/rspace-os/rspace-core-model/pull/35), which means operations listed in RSDEV-796 scenario should all be blocked on model level.

## Testing notes
RSDEV-796.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3040,7 +3040,7 @@
     <jstl.version>1.1.2</jstl.version>
     <lucene.version>5.5.5</lucene.version>
     <netty.version>4.1.63.Final</netty.version>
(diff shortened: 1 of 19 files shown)


rs-fraser (approved):
Looks good. Tested on the [AWS build](https://rsdev-796-block-move-out-of-shared-39.researchspace.com/) following [the test plan](https://researchspace.atlassian.net/browse/RSDEV-796) and now get an error message when attempting to move the docs out of the shared folder.

PR #476, opened 2025-09-19
RSDEV-755: correctly index doc-type gallery attachment when uploading new version

mKowalski256:
The PR moves call of `indexFile(media);` few lines down, so when method is called, the `media.fileProperty` is updated to the new property. Previously, the indexing was run with the old `fileProperty`, so subsequent search couldn't find the connected record.

The real changes are just in two classes: `MediaManagerImpl.java` and `AttachmentSearchTest.java`, other files in the PR are just a spotless formatting missing from previous merges.

### DIFF
--- a/src/main/java/com/researchspace/service/impl/MediaManagerImpl.java
+++ b/src/main/java/com/researchspace/service/impl/MediaManagerImpl.java
@@ -490,9 +490,8 @@ public class MediaManagerImpl implements MediaManager {
           fileStore.createAndSaveFileProperty(fileType, user, updatedFileName, inputStream);
     }
 
+    // reset doc-type attachment thumbnail to force re-generation on next request
     if (media.isEcatDocument()) {
-      indexFile(media);
-      // reset the thumbnail to force re-generation on next request
       ((EcatDocumentFile) media).setThumbNail(null);
     }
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/researchspace/service/impl/MediaManagerImpl.java
mKowalski256, line 1: This is the file with a fix.

src/test/java/com/researchspace/service/AttachmentSearchTest.java
```diff
@@ -3,11 +3,12 @@
 import static com.axiope.search.SearchConstants.ATTACHMENT_SEARCH_OPTION;
 import static com.axiope.search.SearchConstants.FULL_TEXT_SEARCH_OPTION;
 import static com.researchspace.testutils.SearchTestUtils.createAdvSearchCfg;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
rs-fraser, line 6: These are junit 5 imports whereas the rest of the junit imports are v4.

rs-fraser (approved):
Looks good. Left 1 minor comment about junit imports

PR #490, opened 2025-10-03
PRT-991: Fix error when list of DMPUser fetched is empty

rs-nicof:
## Description ##
This PR fixes the error raised by the server when a DMPUser object is not fetched.
Now the error is not raised and the old-DMP source will be viewed as UNKNOWN

## Testing notes
- AWS instance: dhttps://dmponline-test.researchspace.com/ (as https://prt-991-dmp-fetch-error-2.researchspace.com/)


mKowalski256 (approved):
Looks good & fixes problem on p8085.

PR #493, opened 2025-10-06
Rsdev 440 user deletion forms

rs-fraser:
## Description ##
This PR fixes issues when deleting users, where deleting their forms would fail if the form was referenced as previous version or as a temporary version of another form.

The fix is to set these FK columns to null before attempting to delete.

See https://researchspace.atlassian.net/browse/RSDEV-440 for full details

### DIFF
--- a/src/main/java/com/researchspace/dao/hibernate/UserDeletionDaoHibernate.java
+++ b/src/main/java/com/researchspace/dao/hibernate/UserDeletionDaoHibernate.java
@@ -316,7 +316,7 @@ public class UserDeletionDaoHibernate implements UserDeletionDao {
   }
 
   private void deleteForms(Long userId, Session session) {
-   // Set the temp field column to NULL to avoid FK violation during deletion
+    // Set the temp field column to NULL to avoid FK violation during deletion
     execute(
         userId,
         session,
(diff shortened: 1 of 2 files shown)


mKowalski256 (commented):
The PR code looks good and it seems to solve all delete user scenarios where forms were not shared/used by other users. But I can still reproduce the problem with `previousVersion_id` constraint when going through some shared form scenario, I think that's triggered in code path that tries transferring the forms to the sysadmin.

#### Scenario to reproduce
Maybe there are easier steps, but that's the scenario I confirmed both on [AWS build](https://rsdev-440-user-deletion-forms-6.researchspace.com/) and localhost:
1. create two unrelated users, userA and userB
2. as userA, create a new formA, add a field to it, publish for global read, save and close the form (the form will be listed as 'version 0')
3. still logged as userA, edit formA - add second field to it, update the form (the form will be listed as 'version 1')
4. as userB, find formA on 'manage forms' screen, add it to the menu & create a document from it
5. as userA, edit formA - add third field to it, update the form (the form will be listed as 'version 2')
6. as sysadmin, try deleting userA
-- deletion will fail due to 
```
Cannot delete or update a parent row: a foreign key constraint fails (`rspace`.`RSForm`, CONSTRAINT `FK90A082A5813663E7` FOREIGN KEY (`previousVersion_id`) REFERENCES `RSForm` (`id`))
```

PR #498, opened 2025-10-08
fix: add missing React and React DOM typings

tyteen4a03:
## Description ##
Adds missing `@types/react` and `@types/react-dom` dev dependencies.


rlamacraft:
Looks fine to me. Type checker is happy that this doesn't introduce any new errors or warnings; I'm assuming my editor must just be checking automatically without the explicit dependency. Feel free to merge

PR #499, opened 2025-10-08
fix(inventory): misleading description for explicit access list

tyteen4a03:
## Description ##
Fixes misleading description for "Explicit Access List".

Fixes RSDEV-822


rlamacraft:
We don't use conventional commits, so no need for `fix(inventory)` in PR header. Instead please put the ticket number, in this case `RSDEV-822`, at the beginning so that's its easy for us to find the ticket when looking through the git history.

Otherwise the change looks correct to me. Normally jenkins would now run against the change and check for type errors or test failures but given what's being changed here I think we confidently merge this in.

PR #502, opened 2025-10-09
RSDEV-428 Change document editors's Gallery picker to use new Gallery

rlamacraft:
## Description ##
The file picker for choosing Gallery files when editing a rich text field in the ELN's document editor now uses the new Gallery, rather than the old, removing a blocker on removing the old gallery.

Note that whilst this change creates a new TinyMCE plugin to support this new picker, it utilises the existing jQuery code to generate the required HTML to insert into the document field based on the file type.

### DIFF
--- a/src/main/webapp/scripts/pages/workspace/mediaGalleryManager.js
+++ b/src/main/webapp/scripts/pages/workspace/mediaGalleryManager.js
@@ -1226,7 +1226,7 @@ function insertChemistryFileFromGallery(mediaType, data) {
   RS.blockPage("Inserting Chemical...");
   var promises = [];
   $(selected).each(function (index, val) {
-    var fileName = val.children.item(1).children.item(0).getAttribute("data-recordname");
+    var fileName = (data != null) ? data.name : val.children.item(1).children.item(0).getAttribute("data-recordname");
     promises.push(insertChemElement(val.id, fieldId, fileName));
   });
   Promise.all(promises)


mKowalski256:
> > * I seem unable to insert an item from chemistry subgallery - after selecting and clicking 'add' the dialog is closed, but nothing is inserted.
> > * ... same for attempt to insert a link to PDF from 'exports' subgallery
> 
> @mKowalski256 These are actually due to an existing bug with inserting links to Gallery items. If you paste a Global ID of most Gallery files into the TinyMCE editor we convert that to a link, but not so for chemistry files and PDF exports. Is this intentional, or something I should fix first?

Oh, I was just comparing to behaviour of old 'insert from Gallery', didn't realize it's a wider issue. I can't see how that would be intentional, more probably pasting a global ID was a scenario which we didn't test much / don't have full test coverage for.

rlamacraft:
> > > * I seem unable to insert an item from chemistry subgallery - after selecting and clicking 'add' the dialog is closed, but nothing is inserted.
> > > * ... same for attempt to insert a link to PDF from 'exports' subgallery
> > 
> > 
> > @mKowalski256 These are actually due to an existing bug with inserting links to Gallery items. If you paste a Global ID of most Gallery files into the TinyMCE editor we convert that to a link, but not so for chemistry files and PDF exports. Is this intentional, or something I should fix first?
> 
> Oh, I was just comparing to behaviour of old 'insert from Gallery', didn't realize it's a wider issue. I can't see how that would be intentional, more probably pasting a global ID was a scenario which we didn't test much / don't have full test coverage for.

Ok, so this change also fixes that bug too then

PR #518, opened 2025-10-15
consistently use 'Document' as type value for details of a Gallery Document

mKowalski256:
That's to fix inconsitency where `/workspace/getDetailedInformation` endpoint would say that specific Gallery Document has `type="Document"`, but `/gallery/getUploadedFiles` endpoint would say that specific Gallery Document has `type="Documents"`.


mKowalski256:
@rlamacraft that PR comes from our discussion this morning, I refactored model code a bit but then found out that when retrieving details Gallery Document the returned 'type' value is sometimes 'Document' and other times 'Documents'. I unified that to return 'Document' in both cases.

I didn't find any React code depending on the 'Documents' value, just some js code, but maybe take a look.

mKowalski256:
Also, with this PR the Gallery info panel will now show document type as 'Document' rather than 'Documents'. 

image

PR #519, opened 2025-10-15
RSDEV-756 show deployment-defined help links

tyteen4a03:
## Description ##
Shows a list of help links provided by the institution, if defined in deployment config.

Note: There's a bug where opening the menu in the Workspaces section would cause the menu to open at the bottom of the page. As other styling in the App Bar are also not present in the page, the bug is likely not caused by this PR.


src/main/webapp/ui/src/components/Help/HelpDocs.tsx
```diff
@@ -3,6 +3,13 @@ import AnalyticsContext from "../../stores/contexts/Analytics";
 import { observer } from "mobx-react-lite";
 import axios from "@/common/axios";
 import Intercom from "@intercom/messenger-js-sdk";
+import useUiNavigationData from "@/components/AppBar/useUiNavigationData";
+import IconButtonWithTooltip from "@/components/IconButtonWithTooltip";
+import HelpIcon from "@mui/icons-material/Help";
+import Menu from "@mui/material/Menu";
```
rlamacraft: Menu needs to be imported from ../DialogBoundary as currently the menu is broken on non-SPA pages like /workspace 

Image

src/main/webapp/ui/src/components/Help/HelpDocs.tsx
```diff
@@ -35,25 +42,8 @@ declare global {
...
-   * of the popup. This is so that the HelpDocs component can be used in various
-   * different UIs, from sidebars to FABs.
-   *
-   * Note: if the HelpDocs component is not rendered inside of an Analytics
-   * context then disabled will always be true.
-   */
-  Action: (props: {
-    onClick: () => void;
-    disabled: boolean;
-  }) => React.ReactNode;
```
rlamacraft: Oops, I should have cleaned this up when we stopped using a FAB. Glad the complexity is going now

src/main/webapp/ui/src/components/Help/HelpDocs.tsx
```diff
@@ -67,10 +57,14 @@ function loadScript(url: string): void {
   }
 }
 
+const MENU_ID = 'help-docs-menu';
+const MENU_OPEN_BUTTON_ID = 'help-docs-menu-open-button';
```
rlamacraft: HelpDocs is rendered by AppBar, which is used not just in the header of each page but also in the headers of many dialogs. Whilst I don't think HelpDocs is ever rendered by one of the dialogs (as IIRC they all pass a specified `helpPage`), it might be safer to have these ids generated by `useId`

PR #522, opened 2025-10-16
Rsdev 782 jupyter notebooks

nhanlon2:
## Description ##
***REQUIRED***

## Design decisions
***OPTIONAL***

## Testing notes
***OPTIONAL***


src/main/java/com/researchspace/api/v1/model/ApiDocument.java
```diff
@@ -20,6 +20,7 @@
     value = {
       "id",
       "globalId",
+      "version",
```
mKowalski256: This new property could be documented in swagger, in [DocumentInfo](https://github.com/ResearchSpace-ELN/rspace-web/blob/9a18c731180c2813d43c1875babd7c84e2109e8e/src/main/webapp/resources/rspace_api_specs_1_115_0.yaml#L3524) model definition.

PR #548, opened 2025-10-30
PRT-1013 remove card-style share dialog layout

tchoi-rs:
## Description ##
Removes the alternative card-style share dialog layout so the interface is the same, no matter how many items are being shared at one time.


mKowalski256 (commented):
As discussed on the JIRA ticket - the change does work as described, but it seems that we're loosing some functionality when removing the card-style layout of the dialog. Particularly for selection of multiple records, where some are already shared, the behaviour of unified table-style layout is very confusing.  

Let's not merge for now, until specs are re-discussed.

Issue #521, opened 2025-10-16
[BUG]: RSpace API: folderId stripped from URL in HATEOAS pagination


burgerga:
### Observed behavior

I'm requesting `GET https://leiden.researchspace.com/api/v1/folders/tree/242398?pageSize=50`

but then in the returned json I have:

```
...
  "_links": [
    {
      "link": "https://leiden.researchspace.com/api/v1/folders/tree?pageSize=50&pageNumber=1",
      "rel": "next"
    },
    {
      "link": "https://leiden.researchspace.com/api/v1/folders/tree?pageSize=50&pageNumber=0",
      "rel": "self"
    }
  ],
...
```
where the folderId (242398) has been removed from the path, so following these links it's incorrectly returning results from the root folder.

### Steps to Reproduce

1. Go to https://leiden.researchspace.com/public/apiDocs
2. Authenticate
3. Try to get folder tree with folderId (`GET ​/folders​/tree​/{folderId}`)
4. observe HATEOAS links don't have folderId anymore, breaking HATEOAS functionality

### Expected Behavior

Leave folderId in

### Which RSpace version did you encounter the bug with?

1.114.0

### Mobile / Desktop / API

API

### Operating system

MacOS

### Browser

Chrome

### Browser version

141.0.7390.67

PR #550, opened 2025-10-31
ISSUE-521: links in /folder/tree API response

mKowalski256:
Couple of changes here:

1. As raised in #521, the response from `/api/v1/folders/tree/{id}` endpoint was missing `/{id}` suffix in `self` link. This should be fixed now, while the call to `/api/v1/folders/tree` will stay the way it was, i.e. without suffix. 

2. I've noticed that `parentId` property in response to both `/folders/tree/{id}` and `/folders/tree` doesn't seem to have the expected value. The comment describing the property in java class, and I think sensible assumption, is that when listing content of the folder the `parentId` will store the id of that folder. But the code was actually putting the id of the parent of the listed folder. I've changed the behaviour and nothing seems to be broken, I don't think that's actually used anywhere in the code. Additionally, following code-review comment, I've renamed `parentId` property in the response object to `folderId`.

I've also removed some unused params (`user`/`isMediaFolder`) across the code flow.

### DIFF
--- a/src/main/java/com/researchspace/api/v1/controller/FolderApiController.java
+++ b/src/main/java/com/researchspace/api/v1/controller/FolderApiController.java
@@ -203,7 +203,7 @@ public class FolderApiController extends BaseApiController implements FolderApi 
       DocumentApiPaginationCriteria pgCrit,
       BindingResult errors,
       Supplier<Folder> folderSupplier,
-      boolean omitParentIdInLinks)
+      boolean omitFolderIdInLinks)
       throws BindException {
     // validate
     throwBindExceptionIfErrors(errors);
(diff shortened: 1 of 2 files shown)


rs-fraser (commented):
Code changes look good.

I tested on the AWS build and the folder id is added to the self links.

Only suggestion I have is renaming `parentId` to `folderId` in the main response from `/folder/tree/{id}`. 

`parentId` is ambiguous as each of the items in the folder have a `parentId`, which correctly points to the folder they are contained in, and which the request was made for.

The response is a representation of the folder contents, so `parentId` in the top-level of the response could be argued either way. Renaming to either `folderId` or even `id` would clearer IMO.

mKowalski256:
> Only suggestion I have is renaming `parentId` to `folderId` in the main response from `/folder/tree/{id}`.
> 
> `parentId` is ambiguous as each of the items in the folder have a `parentId`, which correctly points to the folder they are contained in, and which the request was made for.

Thank you for review, yes I actually had a thought about `parentId` name being confusing, but decided to keep it so that API is backward-compatible. But as you've noticed the same thing, I'll change it, and just document in [API Changelog](https://documentation.researchspace.com/article/dj4l39d70d-api-changelog).

rs-fraser (approved):

PR #551, opened 2025-11-03
Rsdev 890 fix tests

rs-fraser:
https://researchspace.atlassian.net/browse/RSDEV-890

## Description ##
There's a much more detailed write-up in the ticket (please read), but essentially, I was unable to run the `SpringTransactionalTests` with `mvn clean test` locally due to 3 tests failing. The tests that failed each called `WorkspaceController.createNotebook`, however the `WorkspaceController` bean is modified in `WorkspaceControllerTest`. This modification causes future tests which use `WorkspaceController` to potentially fail due to the bean being in an unexpected state (with some mocked dependencies rather than concrete ones). 

The fix here was to move the bulk of the work done by `createNotebook` into `WorkspaceService` for 2 reasons: 1) to avoid using the mutated WorkspaceController bean and 2) as creates better separation of concerns and allows the notebook creation code to be re-usable without having to call a controller method.

1 test was updated to use this new method. 2 of the tests that had been failing were updated to use the test utility `createNNotebooks`.

This issue isn't seen on Jenkins, which due to some jvm/jdk/maven/os difference must run the tests in a different order. I decided to run the tests alphabetically (`mvn clean test -Dsurefire.runOrder=alphabetical`) which cauese a differnt failure `InventoryImportManagerTest.testSampleCsvImportWithTemplateErrors:405 expected: <1> but was: <0>`

The issue with `InventoryImportManagerTest.testSampleCsvImportWithTemplateErrors` is a previous test class `InventoryItemCsvImportTest` was mocking some dependencies of `CsvContainerImporter`, so when that bean was later used in the failing test, it was in an inconsistent state. The fix here is to create spring beans specifically for the class that will be mutated, to stop polluting future tests in the shared spring context with beans in an inconsistent state.

## Testing notes
Please run `mvn clean test` locally, and report and failures on [the ticket](https://researchspace.atlassian.net/browse/RSDEV-890) and I'll investigate. As a bare minimum, all devs should be able to run the tests locally. 

I'll create a tech debt ticket to address the wider issue as there's still many instances of spring beans being modified affecting other tests, which can be easily reproduced by running `mvn clean test -Dsurefire.runOrder=random`.

### DIFF
--- a/src/test/java/com/researchspace/service/WorkspaceServiceTest.java
+++ b/src/test/java/com/researchspace/service/WorkspaceServiceTest.java
@@ -72,9 +72,6 @@ public class WorkspaceServiceTest {
     target = folder(TARGET_FOLDER_ID);
     user = TestFactory.createAnyUser("user");
     doc = docWithParent(RECORD_ID, source, user.getUsername());
-    org.mockito.Mockito.lenient()
-        .when(folderManager.getRootFolderForUser(any(User.class)))
-        .thenReturn(root);
   }
 
   private void mockHasMovePermission(boolean hasPermission) {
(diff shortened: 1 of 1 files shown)


mKowalski256 (approved):
The changes look good and work for me, although I didn't see local test failures caused by the ordering. 

The test run like this: 
```
mvn clean test -Dtest=WorkspaceControllerTest,RecordManagerTest,RecordApiManagerTest,SharingHandlerTest
```
does actually results in one failure on my local, which is in ` RecordManagerTest.moveSearchResultsAsPI`. That test does fail when run on it's own though, so that must be a different issue.

PR #557, opened 2025-11-06
Prt 1017 fix nested transaction rollback

rs-fraser:
https://researchspace.atlassian.net/browse/PRT-1017

## Description ##
Performing the workspace move action in a transaction introduced in this ticket https://researchspace.atlassian.net/browse/RSDEV-836 meant that exceptions thrown in a (now nested) transaction for sharing the moved record would cause the outer transaction to roll-back. 

That's not desirable in this case as a) it's a validation issue so is recoverable and b) this code path uses ServiceOperationResult to report errors, rather than relying on exceptions being thrown and c) multiple records can be moved at once, and each move can succeed or fail independently without rolling all of the moves back.

The fix is to run the now nested transaction in a new transaction, to not force the outer transaction to rollback in case of an issue.

There are also a couple of formatting changes.


mKowalski256 (approved):

PR #561, opened 2025-11-06
RSDEV-892 Remove unintentional alphanumeric restriction on signup password field

tchoi-rs:
## Description ##
Removes alphanumeric restriction on password field on sign-up / user creation pages, and clarifies this requirement on the signup page as well.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1715,21 +1715,17 @@
     <dependency>
       <groupId>com.github.rspace-os</groupId>
       <artifactId>rspace-core-util</artifactId>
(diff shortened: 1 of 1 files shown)


mKowalski256 (commented):
The change allows me to save all symbols for password on /signup page. But the modified way of enforcing the max length works in quite confusing way - if I try typing password longer than 255 chars, the input just semi-silently stops accepting new characters after reaching 255. I think the previous way, where user got an error message, was better.

Previous warning:
image

PR #564, opened 2025-11-11
RSDEV-854: apps page

nhanlon2:
## Description ##
***REQUIRED***

## Design decisions
***OPTIONAL***

## Testing notes
***OPTIONAL***

### DIFF
--- a/src/main/webapp/ui/src/eln/apps/integrations/Jupyter.tsx
+++ b/src/main/webapp/ui/src/eln/apps/integrations/Jupyter.tsx
@@ -33,13 +33,10 @@ function Jupyter(): React.ReactNode {
                         overflowX: "auto",
                       }}
                   >
-            {`
-%pip install rspace-client==2.6.2
-            
-            `}
+            {`%pip install rspace-client==2.6.2`}
               </pre>
                 </Typography>
-                <b>Run the cell then restart the kernel AND REFRESH THE BROWSER TAB RUNNING JUPYTER.</b>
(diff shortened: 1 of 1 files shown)


src/main/webapp/ui/src/eln/apps/integrations/Jupyter.tsx
```diff
@@ -0,0 +1,95 @@
...
+                        overflowX: "auto",
+                      }}
+                  >
+            {`
+%pip install rspace-client==2.6.2
+            
+            `}
+              </pre>
+                </Typography>
+                <b>Run the cell then restart the kernel AND REFRESH THE BROWSER TAB RUNNING JUPYTER.</b>
```
rs-fraser: If the refreshing the browser part is important, should the rest of the be normal then that part bolded? The caps make it seem a bit shouty to me. Maybe one for @vaidap

src/main/webapp/ui/src/eln/apps/CardListing.tsx
```diff
@@ -346,6 +347,7 @@ function CardListing({
         />
       )}
       {integrationStates.API_DIRECT.mode === mode && <ApiDirect />}
+      { integrationStates.API_DIRECT.mode === mode && <Jupyter />}
```
rs-fraser: Shouldn't there be an `integrationStates.JUPYTER` entry? Or, since I don't think either API_DIRECT or JUPYTER can be "enabled", should `integrationStates.API_DIRECT` be renamed to something the conveys it's a placeholder, like EXTERNAL or NA, and maybe add a comment explaining it?

PR #574, opened 2025-11-24
Start generating webpack bundle stats file JSON

tchoi-rs:
## Description ##
Start generating stats JSONs for work needed in #573 .


mKowalski256 (approved):
I'm not sure about what the change is doing, but the tests are passing and AWS build looks fine :)

PR #578, opened 2025-11-26
RSDEV-791: fix creation of filestore folder starting at subdirectory

mKowalski256:
## Description ##
This PR fixes problem with navigating/saving new NFS filestore when browsing NFS file system, on the new Gallery. The problem was that the `path` react variable used to store the remote path was uri-encoded on every navigation to subfolder, causing parts of the path to be re-coded multiple times. 

The PR changes the code so when constructing the `path` during the navigation, only the newly added suffix of the path is uri-encoded. Then, when path is selected for saving, the `selectedFolderPath` react variable stores decoded path, to avoid double-encoding on filestore save (POST) request.

### DIFF
--- a/src/main/webapp/ui/src/eln/gallery/components/AddFilestoreDialog.tsx
+++ b/src/main/webapp/ui/src/eln/gallery/components/AddFilestoreDialog.tsx
@@ -315,7 +315,7 @@ function FolderSelectionStep(props: {
           {selectedFilesystem
             .map(({ id, name }) => (
               <TreeListing
-                path={encodeURIComponent("/")}
+                path="%2F"
                 fsId={id}
                 fsName={name}
                 key={null}


src/main/webapp/ui/src/eln/gallery/components/AddFilestoreDialog.tsx
```diff
@@ -309,13 +309,13 @@ function FolderSelectionStep(props: {
...
             if (!selected) return;
-            setSelectedFolderPath(selectedFolder);
+            setSelectedFolderPath(decodeURIComponent(selectedFolder));
           }}
         >
           {selectedFilesystem
             .map(({ id, name }) => (
               <TreeListing
-                path="/"
+                path={encodeURIComponent("/")}
```
tchoi-rs: You can probably also inline this to `%2F`

PR #579, opened 2025-11-27
Refactor ConfirmationDialog to create new React roots instead of sharing one root

tchoi-rs:
## Description ##
This avoids the many pitfalls of the current impelmentation and avoids an infinite loop when the webpack entrypoints are optimised.

### DIFF
--- a/src/main/webapp/ui/src/my-rspace/profile/ConnectedAppsTable.tsx
+++ b/src/main/webapp/ui/src/my-rspace/profile/ConnectedAppsTable.tsx
@@ -118,7 +118,7 @@ export default function ConnectedAppsTable() {
                     <ConnectedAppsTableRow
                       // @ts-expect-error stableSort needs better types
                       app={app}
-                      onConfirmDisconnectApp={handleConfirmDisconnectApp}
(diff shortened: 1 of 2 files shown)


src/main/webapp/ui/src/my-rspace/profile/ConnectedAppsTable.tsx
```diff
@@ -0,0 +1,144 @@
...
+  };
+
+
+  useEffect(() => {
+    void fetchApps();
+  }, []);
+
+  const handleConfirmDisconnectApp = async (id: string) => {
+    try {
+      await axios.delete(`/userform/ajax/oAuthConnectedApps/${id}`);
```
mKowalski256: on test AWS build I see that `id` passed  here is `undefined`, and so the endpoint being called is `/userform/ajax/oAuthConnectedApps/undefined`

mKowalski256 (commented):
I've checked the modified dialogs and found issue with one of them - the one used for disconnecting oauth app on My Profile page. Otherwise, the PR looks good to me.

PR #585, opened 2025-12-03
Remove unnecessary property-based testing for Zenodo component test

tchoi-rs:
## Description ##
Change the Zenodo component test to traditional jest tests to reduce flakiness.


tchoi-rs:
@mKowalski256 this is fixed in the webpack entrypoints PR.

PR #588, opened 2025-12-04
Rsdev 910 export galaxy data

nhanlon2:
## Description ##
***REQUIRED***

## Design decisions
***OPTIONAL***

## Testing notes
***OPTIONAL***


src/test/resources/TestResources/archives/v0-18/0.18Export/doc_Experiment1-298/doc_Experiment1-298.xml
mKowalski256: I believe the dir `src/test/resources/TestResources/archives/v0-18/0.18Export` represents the export from early RSpace version, and is here to ensure whatever changes we do to import, it's still backward compatible i.e. will read the old archives fine. Therefore I think we shouldn't be changing the files in this TestResources folder. 

For testing latest xml model maybe create new dir like `src/test/resources/TestResources/archives/v2.17Export`, and put the new content there? Alternatively the `src/test/resources/TestResources/archives` collects various `.zip` archives that  were added over time, so the whole zip could be added rather than expanded folder.

src/main/java/com/researchspace/service/archive/ExternalWorkFlowsImporter.java
```diff
@@ -0,0 +1,125 @@
...
+                allExternalWorkFlowInvocations) {
+              if (invocationMetaData.getDataIds().contains(externalWorkFlowMetaData.getId())) {
+                ArchiveExternalWorkFlow workFlow =
+                    allWorkflows.findById(invocationMetaData.getWorkFlowId());
+                ExternalWorkFlow existingWorkFlow =
+                    externalWorkFlowDataManager.findWorkFlowByExtIdAndName(
+                        workFlow.getExtId(), workFlow.getName());
+                if (existingWorkFlow != null) {
+                  for (ExternalWorkFlowInvocation existingInvocation :
+                      existingWorkFlow.getExternalWorkflowInvocations()) {
```
mKowalski256: when testing import on localhost, for some time I was getting a `ConcurrentModificationException` coming from this line, I think because set of invocations in `existingWorkFlow` may be modified within constructor of `ExternalWorkFlowInvocation` in line 82.  

I believe that is depending on order in which the archive docs are imported, which does change, and so I didn't replicate it consistently. But I believe the issue would be fixed by iterating over set containing just the original invocations, i.e. by replacing that `for` line with:
```
  Set<ExternalWorkFlowInvocation> initialInvocations = new HashSet<>(existingWorkFlow.getExternalWorkflowInvocations());
  for (ExternalWorkFlowInvocation existingInvocation : initialInvocations) {
```

PR #598, opened 2025-12-15
refresh back-end docs for adding a new integration

rs-fraser:
## Description ##
A while ago we updated some front-end docs to make it easier for contributors to add new integrations and tinymce plugins. I've done the same for the back-end by refreshing the existing docs with some fairly generic steps about what needs to be updated to add a new app. Whoever adds the next integration can follow the steps and amend if needed.

### DIFF
--- a/DevDocs/DeveloperNotes/CreatingNewIntegration.md
+++ b/DevDocs/DeveloperNotes/CreatingNewIntegration.md
@@ -6,12 +6,46 @@ This document describes the steps required to add a new integration (App) to RSp
 
 The backend changes required for adding a new integration involve:
 
-1. Adding deployment properties (e.g. URLs) for the integration, if required
-2. Adding a system property to enable/disable the integration at the deployment level
-3. Adding a user preference to enable/disable the integration for each user
+1. Adding a system property to enable/disable the integration at the deployment level
+2. Adding a user-level toggle by creating an App row (and optional per-user config)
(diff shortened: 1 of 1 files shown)


DevDocs/DeveloperNotes/CreatingNewIntegration.md
```diff
@@ -0,0 +1,82 @@
...
+Some integrations require additional configuration (for example, user API keys or tokens).
+
+• App only needs to be enabled/disabled
+
+Some integrations are simply toggled on or off per user. These integrations use the `Preference` enum, which should be updated in the `rspace-core-model` project. Add the new `Preference` to the `booleanIntegrationPrefs` `EnumSet` in `IntegrationsHandlerImpl`.
+
+• App requires individual configuration
+
+Most integrations use `UserAppConfig` to store information specific to the user for that integration. If that is the case, add your integration to the list in `IntegrationsHandlerImpl.isAppConfigIntegration()`.
+
```
mKowalski256: If the integration is to support more configuration options beyond just enabled/disabled, these should be added through inserting rows into `PropertyDescriptor` + `AppConfigElementDescriptor` table. See e.g. 'fieldmark' app that requires user to provide `FIELDMARK_APIKEY`, defined in `sqlUpdates/changeLog-rsdev-369.xml`.

rs-fraser: made this clearer and added an example

PR #607, opened 2026-01-06
Set default "limiting reagent"

rs-fraser:
https://researchspace.atlassian.net/browse/PRT-1030

## Description ##
When a new stoichiometry table is created, the front-end shows the first reactant is selected as the "limiting reagent". However, in the initial state in the database no reactants are selected. This PR sets the first REACTANT (if any) molecule as the limiting reagent, to match what the user sees on the front-end. 

Also spotless has picked up a couple of formatting changes.

## Testing notes
As the front-end was by default selecting the first molecule before this fix, verify the first reagent is persisted as such either by exporting the document to pdf of html, or by looking at `StoichiometryMolecule.limiting_reagent` in the database.


mKowalski256 (approved):

PR #629, opened 2026-01-21
RSDEV-989: Add `raid` field in `ApiGroupInfo` object

rs-nicof:


rs-nicof:
Updated swagger definition of `ApiGroupInfo` as well

PR #645, opened 2026-02-02
public API /share endpoint accepts comma-separated shareItemIds

mKowalski256:
* `/api/v1/share` endpoint accepting list of record ids for which the shares should be returned
* returned share information object (ApiShareInfo) now contains `sharedTargetId`
* if `shareItemIds` list contains id of a shared (sub)folder, the response contains pseudo-share element describing target of the share (i.e. as if the folder was shared)
* 401 unauthorized error returned if user doesn't have access to any of the requested 'id'

### DIFF
--- a/src/main/java/com/researchspace/api/v1/controller/ApiGenericSearchConfig.java
+++ b/src/main/java/com/researchspace/api/v1/controller/ApiGenericSearchConfig.java
@@ -16,7 +16,7 @@ import org.springframework.util.MultiValueMap;
 @NoArgsConstructor
 public class ApiGenericSearchConfig extends ApiSearchConfig {
 
-  public static final int MAX_QUERY_LENGTH = 2000;
+  private static final int MAX_QUERY_LENGTH = 2000;
 
   @Size(max = MAX_QUERY_LENGTH, message = "Max query length is " + MAX_QUERY_LENGTH + ".")
   private String query;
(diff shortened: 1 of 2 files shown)


tchoi-rs:
> > Just checking, are we adding psuedo-shares into the existing share results array? I think the ID is a required field in there, and since group shares don't have IDs (as they are virtual), they would not have IDs and this would technically break BC.
> 
> yes, the folder pseudo-shares are part of result `shares` array. Should I put them into separate array?

I would err on being conservative with BC and put them in a separate array. (In v2 we can unify the concept of documents and folders)

src/main/webapp/resources/rspace_api_specs_2_20_0.yaml
```diff
@@ -4443,13 +4445,51 @@ definitions:
...
+      sharedTargetId: 1234
+      sharedTargetType: GROUP
+      permission: READ
+  ShareFolderInfo:
+    description: Represents the information about a shared folder
+    properties:
+      id:
+        type: integer
+        format: int64
+        description: Will always be null, as there is no underlying share object for folders
```
tchoi-rs: This is missing a `nullable: true`

mKowalski256:
> Looks good - are the OpenAPI specs auto-generated? It doesn't seem to have the new `folderShares` field.

No, the OpenAPI docs are updated manually, I've just updated the yaml to describe folder shares, can be seen on latest build at https://rsdev-747-extend-list-share-endpoint-26.researchspace.com/public/apiDocs.

PR #646, opened 2026-02-03
PRT-1039: Move GroupInfo creation into the manager to avoid Hibernate Session exception

rs-nicof:


mKowalski256 (approved):
Code-wise looks good & should solve the issue.

PR #648, opened 2026-02-04
PRT-1040 use BroadcastChannel to pass authToken

tchoi-rs:
## Description ##
Use BroadcastChannel to pass authToken when authenticating via GitHub.

Fixes PRT-1040


mKowalski256 (approved):
Codewise looks good to me, we can confirm it's fixing the issue after merging&redeploying to pangolins.

PR #653, opened 2026-02-11
Don't generate webpack stats unless in CI environment

tchoi-rs:
## Description ##
Don't generate webpack stats unless we're in CI mode, to reduce overhead when building locally.


mKowalski256 (approved):
Looks good - stats.json is no longer generated on my localhost, but I see it generated as a part of [jenkins build ](https://jenkins.researchspace.com/job/rspace-web/job/feat-dont-generate-webpack-stats/2/execution/node/4/ws/src/main/webapp/ui/dist/).

PR #727, opened 2026-04-10
Harden GitHub Actions and disable webpack bundle analysis step for now

tchoi-rs:
## Description ##
Hardens GitHub Actions and disable webpack bundle analyzer step which is currently OOMing.


mKowalski256 (approved):
Looks good to me, although `greetings.yml` is a workflow that @tilorspace worked on, did you agree on removing it?

PR #731, opened 2026-04-20
Allow creating a stoichiometry table without a reaction drawing

tchoi-rs:
## Description ##
This PR allows users to create a Stoichioimetry table without a reaction drawing.


src/main/webapp/scripts/pages/workspace/editor/tinymce5_configuration.js
```diff
@@ -559,6 +559,7 @@ function initTinyMCE(selector) {
 		}
 		// always load tinymceStoichiometry as it handles the case of chemistry not being enabled
 		localTinymcesetup.external_plugins["stoichiometry"] = "/ui/dist/tinymceStoichiometry.js";
+		addToMenuIfNotPresent(localTinymcesetup, " | stoichiometryMenuItem");
```
mKowalski256, line 562: [RSDEV-874](https://researchspace.atlassian.net/browse/RSDEV-874) says about a new icon in tinymce toolbar, but the code here only adds a menu item (i.e. visible in dropdown 'insert' menu, not on main toolbar).

src/main/webapp/ui/src/tinyMCE/stoichiometry/index.tsx
```diff
@@ -1,100 +1,290 @@
...
+      if (!hasElementChildren && textContent.length === 0) {
+        node.textContent = EMPTY_STOICHIOMETRY_TABLE_PLACEHOLDER;
+      }
+    };
+
+    const getStoichiometryTableOnlyText = (stoichiometry: {
+      id: number;
+      revision: number;
+    }): string => {
+      return `Stoichiometry Table ID ${stoichiometry.id} (click to edit)`;
```
mKowalski256: this `Stoichiometry Table ID ${stoichiometry.id} (click to edit)` label looks pretty confusing to me: 1st we don't usually show db ids to the user, 2nd the 'click to edit' doesn't really work, user needs to select the element then click on stoichiometry button - which is the normal way stoichiometry works, so I'd say doesn't require explanation.

How about label saying: `Stoichiometry Table (no preview)` instead - I'd think that's more useful, as compared to regular stoichiometry it's currently a bit confusing that you don't see the chemical reaction image.

mKowalski256 (commented):
The functionality looks good & works fine for me on [test AWS build](https://feat-stoichiometry-table-without-chem-10.researchspace.com/), but I've added a couple of suggestions to the code. 

---- 

Also I've notice a couple of UI issues that I'm unable to pin to any specific code line:

####  1. view mode field content div not expanding for new stoichiometry
1st one is about document view mode when text field contains a stoichiometry table without reaction drawing. The div showing the text field content is not extended vertically to show the whole content of the stoichiometry table, instead it applies a scrollbar. This doesn't seem right, and is not happening for regular stoichiometry tables that do have reaction drawing (in that case the field div is extended to show whole stoichiometry table). 

See the screenshot and video switching between two docs having two types of stoichiometry:

Image

https://github.com/user-attachments/assets/17712394-61d7-4bd3-a090-0ec99dea66e0

#### 2. (minor) context button hovering over the dialog when creating new stoich

When clicking to create new stoichiometry table, it seems that underlying tinymce element is selected which triggers stoichiometry context button that floats over dialog. The context button also appears after some other dialog actions, e.g. 'save changes'. That doesn't happen for stoichiometries with connected chemical reactions.

Image

https://github.com/user-attachments/assets/f4533c9e-d1ce-4887-aa80-68476a6bbe11

src/main/webapp/scripts/pages/workspace/editor/tinymce5_configuration.js
tchoi-rs: This is fixed now.

src/main/webapp/ui/src/tinyMCE/stoichiometry/index.tsx
tchoi-rs: This is fixed now.

tchoi-rs:
@mKowalski256 The height issue is fixed (only applies to brand new tables). The context menu icon one is a bit harder to fix and I'm inclined to leave it as is for now as the editor will be replaced soon.

mKowalski256 (approved):
Looks good to me on new AWS build!

PR #733, opened 2026-04-21
RSDEV-1025: fix by adding call to UserContentUpdater in addUserToGroup

nhanlon2:
## Description ##
***REQUIRED***

## Design decisions
***OPTIONAL***

## Testing notes
***OPTIONAL***

### DIFF
--- a/src/main/java/com/researchspace/service/impl/GroupManagerImpl.java
+++ b/src/main/java/com/researchspace/service/impl/GroupManagerImpl.java
@@ -110,7 +110,7 @@ public class GroupManagerImpl implements GroupManager {
   private @Autowired PermissionFactory permFactory;
   private @Autowired GroupDao groupDao;
   private @Autowired CollaborationGroupTrackerDao collabGrpTrackerDao;
-
+  private @Autowired UserContentUpdater userContentUpdater;
   private @Autowired CommunicationDao communicationDao;
   private @Autowired CommunityDao communityDao;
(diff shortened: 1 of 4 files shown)


src/main/java/com/researchspace/service/impl/GroupManagerImpl.java
```diff
@@ -180,6 +180,9 @@ public Group addUserToGroup(String usernameToAdd, Long groupId, RoleInGroup role
 
     Group group = groupDao.get(groupId);
     User userToAdd = userDao.getUserByUsername(usernameToAdd);
+    // Legacy users that have never logged in and dont belong to any group will not have the correct
```
rs-nicof, line 183: Can you pls add a test for this scenario on `GroupManagerTest`?

rs-nicof (changes requested):

rs-nicof:
PR code looks good! 
I've requested to add a test for the scenario we are fixing

src/main/java/com/researchspace/service/impl/GroupManagerImpl.java
nhanlon2: I have added a crude test that asserts mockbean is called - we have so many interdependent tests that using real beans was breaking other tests.

nhanlon2: Have done

PR #738, opened 2026-04-25
Update AboutRSpaceDialog.tsx to stop using hardcoded year and use .getFullYear instead

tuxRamon:
## Description ##
Noticed that the about page has the year 2025 hardcoded into it (see screenshot). I have changed it to use getFullYear so that the year will always stay the current one

## Design decisions
N/A

## Testing notes
N/A

Screenshot:
image

### DIFF
--- a/src/main/webapp/ui/src/components/AppBar/AboutRSpaceDialog.tsx
+++ b/src/main/webapp/ui/src/components/AppBar/AboutRSpaceDialog.tsx
@@ -119,7 +119,7 @@ export function AboutRSpaceContent(): React.ReactElement {
         RSpace is licensed under AGPL.
       </Typography>
       <Typography variant="caption" align="center" color="textSecondary">
-        © 2025 ResearchSpace
+        © {new Date().getFullYear()} ResearchSpace
       </Typography>
 
       <Box mt={2} mb={3}>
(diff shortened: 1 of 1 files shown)


mKowalski256 (commented):
Actually, the copyright date means the date when the software (or its version) was released - not necessarily the current year. We set this date in `pom.xml` file, then bump it at the beginning of each year, so the date matches the release. That's how the copyright date seen in footer is generated (e.g. on RSpace Login page). RSpace version number displayed in the footer (also on 'about' page) is also taken from `pom.xml` file.

Image

While reactified 'about' page reads RSpace version number from `/public/version` endpoint (which boils down to version set in `pom.xml`), apparently the copyright date doesn't currently have such endpoint, and so was hardcoded as a quick solution. We should either add an endpoint like `/public/copyrightYear` and read it from 'about' page, or maybe extend `/public/version` to have more information. Or, as a quick fix, we could just bump hardcoded date to `2026` - and then remember to bump it at the beginning of 2027 (same time when we bump copyright date in `pom.xml`).

tchoi-rs:
This doesn't actually change the implementation to `.getFullYear()` as the PR suggests.

tuxRamon:
@mKowalski256 upto you how you decide to implement it now that I am aware of the reasoning behind it being the year of the software release and not the current year. I would avoid adding more info to /public/version. Loads of internal scripts rely on that endpoint being in it's current format.

PR #740, opened 2026-04-27
RSDEV-1023: s3 nfs client

mKowalski256:
## Description

Implements the S3 Gallery filestore connector (RSDEV-1023), part of the S3 read/write epic (RSDEV-994).

Changes span `rspace-web` and `rspace-core-model` (see connected [core-model PR](https://github.com/rspace-os/rspace-core-model/pull/63)):
- New `AwsS3Client` implementing `NfsClient` — folder browsing, file/folder metadata lookup, and file download for S3-compatible storage (AWS S3, Cloudflare R2, etc.)
- `S3UtilitiesImpl` refactored from archive-export-only to a general-purpose S3 operations class, wrapped by a new `S3UtilitiesFactory` Spring component that creates separate instances for NFS vs export use cases
- `NfsFactory` wired to create `AwsS3Client` for the new S3 filestore type, with `NfsRSpaceProvidedAuthentication` for credential-free auth
- Sysadmin UI: S3 added as a filestore option in the admin console
- Gallery React UI: remote file rendering with last-modified date support and `modificationDate` typed consistently as `Date | undefined`

## Design decisions

- **Virtual folder detection**: S3 has no real folders. `getObjectDetails` tries `headObject` first (real file); on 404, probes with `listObjectsV2` (prefix has content) and `headObject` on `path + "/"` (explicit folder placeholder) — covering all common S3 folder conventions
- **`S3Exception` over `NoSuchKeyException`**: catches the parent class and checks `statusCode() == 404`, so the connector works with providers like Cloudflare R2 that return generic 404s instead of the typed SDK exception
- **Pagination**: `listFolderContents` loops with `continuationToken` to fetch all pages, since AWS S3 caps each response at 1000 objects
- **`TempFileDeletingInputStream`**: wraps the download temp file and deletes it on `close()`, preventing temp file leaks without needing a separate release hook
- **`tryConnectAndReadTarget`** throws `UnsupportedOperationException` — S3 connectivity is validated at `S3Utilities` initialisation time via `HeadBucket`

## Testing notes

Unit tests in `AwsS3ClientTest` (13 tests), `S3UtilitiesTest` (pagination), and `NfsFactoryTest` (S3 wiring).

Real-connection integration tests in `S3UtilitiesRealConnectionTest` require:
- `-Dnightly` system property for AWS S3 tests
- `-Dnightly-s3cloudflare` for Cloudflare R2 tests
- Credentials set in `deployment.properties` (not committed)

## Manual testing/review

* specification in JIRA [RSDEV-1023](https://researchspace.atlassian.net/browse/RSDEV-1023)
* some dev documentation added to [ops2 doc](https://ops2.researchspace.com/globalId/SD13539)
* test AWS build running at [rsdev-1023-s3-nfs-client-51.researchspace.com](https://rsdev-1023-s3-nfs-client-51.researchspace.com), has File Systems added for both Cloudflare and AWS S3, but deployment-properties credentials are for Cloudflare so only that will work (unless restarted with modified deployment props).


src/main/java/com/researchspace/service/aws/impl/S3ExportUtilitiesImpl.java
```diff
@@ -0,0 +1,115 @@
...
+      s3Utilities =
+          s3UtilitiesFactory.createS3UtilitiesForAwsArchiveExport(
+              s3ExportRegion, s3ExportBucketName);
+      s3Presigner = S3Presigner.builder().region(Region.of(s3ExportRegion)).build();
+    } catch (Exception e) {
+      log.error(
+          "Error building S3 client with region {} and bucket {}",
+          s3ExportRegion,
+          s3ExportBucketName,
+          e);
```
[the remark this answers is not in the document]
mKowalski256: after  reviewing,  won't  do.

src/main/resources/sqlUpdates/changeLog-0.26.xml
```diff
@@ -169,19 +169,19 @@
...
-   Other updates should go above. That's because customChange code should be the 
+   new changesets and leave this reminder in place!
+
+     Please only put below liquibase changeSets containing customChange command.
+   Other updates should go above. That's because customChange code should be the
    last changeSet in a given changeLog file.  -->
-   
-    <changeSet id="nfsFileSystem_import" author="matthias" context="run">
+
+<!--    <changeSet id="nfsFileSystem_import" author="matthias" context="run">
```
tchoi-rs, line 178: Is there a reason why we're changing historical records here? Migrations shouldn't be touched (but squashing them is OK)

tchoi-rs (changes requested):
Frontend portion looks fine to me, there is one question about liquibase though

src/main/resources/sqlUpdates/changeLog-0.26.xml
mKowalski256: Yeah, the reason why I commented out that historical changeset is that it was running  `NfsFileSystemImporter.java` class, which was using some obsolete static methods that I removed from rspace-core-model as a part of [connected PR](https://github.com/rspace-os/rspace-core-model/pull/63). 

There shouldn't be any problems caused by removing/commenting out this changeset, as its goal was to migrate filesystem configuration into the database, which was aimed at early customers who had filesystems configuration in deployment.properties. All these customers are already migrated, and newer/fresh deployments will not have these deployment properties. There are actually more historical liquibase changesets we could remove, and I see now I even raised [RSDEV-416](https://researchspace.atlassian.net/browse/RSDEV-416) for it.

tchoi-rs (approved):

rs-fraser (approved):
Code changes look good.

Tested on the AWS build and works as expected.

PR #742, opened 2026-04-28
RSDEV-1077: Fix internal link search by including blockUI template

rs-fraser:
## Description ##
The internal link dialog's search has been broken since PRT-1016 (cf1bdfe45) upgraded Mustache.js to v4.2.0. The `RS.blockPage()` function renders a `#blockUIContentTemplate` Mustache template, but that template (from `blockUI.html`) was never included in the TinyMCE iframe's `internalLink.jsp`. Mustache 4.2.0 throws on undefined templates, silently crashing the search handler before the AJAX call executes.

Also fixes the same latent issue in the MS Teams tab config page. The pint to tab feature on Teams is deprecated https://documentation.researchspace.com/l/en/article/i95u9itfgu, but I've added the import so that if it's resurrected at any point then we aren't tripped up by this issue. 

## Design decisions
Fixed by including the missing template rather than making `RS.blockPage()` defensive — the template *should* be present wherever `global.js` is loaded.

## Testing notes
AWS build: https://rsdev-1077-internal-links-3.researchspace.com/
Open a document
Click the internal link button on the toolbar, or Insert → Internal Link from the menu
Enter the global ID of a document 
Click Search


mKowalski256 (approved):
Looks good & I've verified that the test scenario is fixed on branch build, compared to regular build. The 'browse' tab of the 'insert link' plugin also didn't work in regular build, and is fixed in branch build.

PR #744, opened 2026-04-29
Rsdev 833 user deletion fails with shared template

mrowe-tan:
## Description ##
- Addresses https://researchspace.atlassian.net/browse/RSDEV-833
- Shared templates are now transferred from the deleted user to the sysadmin account performing the user deletion
- Shared templates are also unshared as part of this process

## Testing notes
See the ticket for testing notes.


src/main/java/com/researchspace/admin/service/impl/SysAdminManagerImpl.java
```diff
@@ -161,6 +162,8 @@ private UserUsageInfo setUserUsageInfoProperties(
         .signupSource(user.getSignupSource().toString())
         .lastLogin(user.getLastLogin())
         .hasFormsUsedByOtherUsers(formDao.hasUserPublishedFormsUsedInOtherRecords(user))
+        .hasTemplatesUsedByOtherUsers(
+            recordGroupSharingDao.getTemplatesSharedByUser(user).size() > 0)
```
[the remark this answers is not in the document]
mrowe-tan: Initially I was going to leave this, since the performance impact seems to be negligible.  However, what I failed to notice myself was that this method was run on a loop when the Users page is loaded.  Whilst the performance impact is still unlikely to be particularly large, and probably not noticeable, I made the suggested change anyway.

src/main/webapp/ui/src/eln/sysadmin/users/index.tsx
```diff
@@ -874,6 +874,30 @@ const DeleteAction = ({
                       that are not used by others will be deleted.
                     </Typography>
                   )}
+                  {(user.hasTemplatesUsedByOtherUsers && !user.hasFormsUsedByOtherUsers) && (
+                      <Typography variant="body2" sx={{ mb: 1 }}>
+                        The user you are trying to delete is{" "}
+                        <strong>
+                          the owner of Templates that are used by other users.
```
tchoi-rs: This can be simplified to say `Forms / Templates` and we can get rid of the differing copy

tchoi-rs (changes requested):

src/main/java/com/researchspace/service/impl/TemplateTransferService.java
```diff
@@ -0,0 +1,115 @@
...
+      for (BaseRecord template : sharedRecords) {
+        List<RecordGroupSharing> sharings =
+            recordGroupSharingDao.getRecordGroupSharingsForRecord(template.getId());
+        for (RecordGroupSharing sharing : sharings) {
+          recordSharingHandler.unshare(sharing.getId(), originalOwner);
+        }
+      }
+
+      List<Long> templateIds =
+          sharedRecords.stream().map(BaseRecord::getId).collect(Collectors.toList());
```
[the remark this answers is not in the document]
mrowe-tan: This should be fixed by way of the changes to getTemplatesSharedByUser, as below.

src/main/java/com/researchspace/webapp/controller/UserProfileController.java
```diff
@@ -196,15 +200,19 @@ public ModelAndView showUserForm(
...
     User user;
 
     if (userId == null) {
       user = sessionUser; // own profile by default
     } else {
-      user = userManager.getUser(userId + "");
+      try {
+        user = userManager.getUser(userId + "");
+      } catch (ObjectRetrievalFailureException e) {
+        throw new UserNotFoundException(USER_NOT_FOUND);
```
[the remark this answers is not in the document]
mrowe-tan: This comment was not implemented due to the fact that the error is displayed on screen for the user, and adding the stack trace to the exception muddies things.  However, the message was updated to include the user ID, which should make things as clear as they need to be.

src/main/webapp/ui/src/eln/sysadmin/users/index.tsx
```diff
@@ -792,6 +792,15 @@ const DeleteAction = ({
       ),
     );
 
+  function getDeleteMessage(user: User): string {
+    let deleteMessage =
```
tchoi-rs: It's fine to just leave it as `user.hasFormsUsedByOtherUsers || user.hasTemplatesUsedByOtherUsers)` and say there are forms and / or templates

tchoi-rs (changes requested):

src/main/java/com/researchspace/dao/hibernate/RecordDaoHibernate.java
```diff
@@ -647,4 +647,106 @@ public Long getRecordCountForUser(RecordTypeFilter recordFilter, User user) {
...
+    // with two parameters...
+    Query<?> query =
+        session
+            .createQuery(
+                "UPDATE RecordToFolder rtf SET folder=:destination WHERE id IN (SELECT id FROM"
+                    + " RecordToFolder WHERE folder.owner=:originalOwner AND record.id IN :ids)")
+            .setParameter("destination", destinationFolder)
+            .setParameter("ids", recordIds)
+            .setParameter("originalOwner", currentOwner);
+    query.executeUpdate();
```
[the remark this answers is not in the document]
mrowe-tan: This query is fine - it was purely the syntax of the where clause when translated by Hibernate for use in the database that was causing issues, hence why it's formatted as it is.

src/main/java/com/researchspace/service/impl/TemplateTransferService.java
```diff
@@ -0,0 +1,110 @@
...
+
+      for (BaseRecord template : sharedRecords) {
+        auditTrailService.notify(
+            new GenericEvent(newOwner, template, AuditAction.TRANSFER, description));
+      }
+    }
+  }
+
+  private Folder determineDeletedTemplatesFolder(User originalOwner, User newOwner) {
+    Folder templateFolder = folderManager.getTemplateFolderForUser(newOwner);
```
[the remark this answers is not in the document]
mrowe-tan: This is a potentially valid concern.  However, users cannot delete their own Templates folders as these are system-created folders.  I'll not implement this for now, since if the Templates folder is missing then there are likely to be larger issues.

src/main/java/com/researchspace/dao/hibernate/RecordGroupSharingDaoHibernateImpl.java
```diff
@@ -186,6 +186,33 @@ private Criteria getSharedRecordsWithUserQuery(User user) {
...
+    Session session = getSession();
+    Object result =
+        session
+            .createQuery(
+                "select count (rgs.id) from RecordGroupSharing rgs join BaseRecord br ON"
+                    + " rgs.shared.id = br.id WHERE br.type like '%template%' AND br.owner=:owner")
+            .setParameter("owner", user)
+            .uniqueResult();
+    Long count = (Long) result;
+    return count > 0;
```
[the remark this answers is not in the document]
mrowe-tan: The above suggested change does not actually work.  It seems that Hibernate is not able to navigate to the BaseRecord object from the "shared" field in the same way that the Java model is able to.  I could be doing this wrong of course, but I'll leave this as is for now.

src/main/webapp/ui/src/eln/sysadmin/users/index.tsx
```diff
@@ -792,6 +792,15 @@ const DeleteAction = ({
...
+    (user.hasFormsUsedByOtherUsers && !user.hasTemplatesUsedByOtherUsers) ? "Transfer Forms And Delete"
+        : (user.hasTemplatesUsedByOtherUsers && !user.hasFormsUsedByOtherUsers) ? "Transfer Templates And Delete"
+            : (user.hasFormsUsedByOtherUsers && user.hasTemplatesUsedByOtherUsers) ? "Transfer Forms And Templates And Delete"
+                : "Delete";
+    return deleteMessage;
```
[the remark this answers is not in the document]
mrowe-tan: This is no longer relevant given subsequent changes to the dialog.

src/main/webapp/ui/src/eln/sysadmin/users/index.tsx
mrowe-tan: The dialog text has been changed as suggested to now not have three different versions.  Following a discussion with Vaida the template/form deletion text has also been moved into an Alert box to bring it to the user's attention.

src/main/webapp/ui/src/eln/sysadmin/users/index.tsx
mrowe-tan: As above.

src/main/java/com/researchspace/admin/service/impl/SysAdminManagerImpl.java
```diff
@@ -161,6 +162,7 @@ private UserUsageInfo setUserUsageInfoProperties(
         .signupSource(user.getSignupSource().toString())
         .lastLogin(user.getLastLogin())
         .hasFormsUsedByOtherUsers(formDao.hasUserPublishedFormsUsedInOtherRecords(user))
+        .hasTemplatesUsedByOtherUsers(recordGroupSharingDao.hasUserSharedTemplates(user))
```
mKowalski256: as for the new line
```
.hasTemplatesUsedByOtherUsers(recordGroupSharingDao.hasUserSharedTemplates(user))
```
This mostly works fine, but as `.hasUserSharedTemplates()` checks templates currently shared by the user, it will handle some cases incorrectly. Notably: 
1. the other user's document may be linked to the template that was shared in the past, but is no longer shared - in that case user deletion will not find the template
2. the user may have shared templates, but nobody used them - in that case the unused template will be unnecessarily transferred to sysadmin.

----
compare it to the line above yours, handling the forms:
```
.hasFormsUsedByOtherUsers(formDao.hasUserPublishedFormsUsedInOtherRecords(user))
```
which, if you look into `formDao.hasUserPublishedFormsUsedInOtherRecords()` implementation, actually checks if there are existing documents, belonging to other users, based on the form. The template check should follow the similar logic, and it's just the templates that are actually referenced by existing documents that should be transferred to sysadmin, irrelevant whether they are currently shared or not.

src/main/java/com/researchspace/dao/hibernate/RecordDaoHibernate.java
```diff
@@ -647,4 +647,106 @@ public Long getRecordCountForUser(RecordTypeFilter recordFilter, User user) {
...
+            .setParameter("ids", templateIds);
+    query.executeUpdate();
+
+    // Modify the template ACL so that the new owner has the same level of
+    // access as the original owner.
+    query =
+        getSession()
+            .createNativeQuery(
+                "UPDATE BaseRecord br SET acl = (SELECT REPLACE(br.acl, :originalOwner, :newOwner)"
+                    + " ) WHERE id IN :ids")
```
mKowalski256: AI agent is telling me this 'REPLACE" may be problematic, although I didn't actually confirm that. Based on my understanding of SQL, the agent  may be  right?

```
ACL REPLACE is brittle — REPLACE(acl, 'john', 'admin') could corrupt ACLs if usernames overlap (e.g. john vs johnson)
```

mKowalski256 (changes requested):
I'm sharing a couple of early comments coming from my testing and AI review - the one about migrating currently shared templates I confirmed, the other about SQL REPLACE problem I didn't but think it may be right.

src/main/java/com/researchspace/admin/service/impl/SysAdminManagerImpl.java
mrowe-tan: I've refactored these methods to no longer be based on the RecordGroupSharing table, but on the actual usage of templates from BaseRecords.  This should deal with the case in which users have shared templates but which are not used, and should also still take into account the condition in which documents using templates have been deleted.

src/main/java/com/researchspace/dao/hibernate/RecordDaoHibernate.java
mrowe-tan: Yes, this does seem to make sense.  It may not have actually been too much of an issue since templates are unshared before reaching this point in the code, but just to be safe I've updated the query to search and replace exact usernames, by way of checking whether they are either directly at the beginning of the ACL string, or in the middle of the string and bound by the & and = characters.

src/main/java/com/researchspace/dao/hibernate/RecordDaoHibernate.java
```diff
@@ -647,4 +647,147 @@ public Long getRecordCountForUser(RecordTypeFilter recordFilter, User user) {
...
+  }
+
+  @Override
+  public boolean hasUserSharedTemplatesUsedByOtherUsers(User user) {
+    Session session = getSession();
+    Object result =
+        session
+            .createQuery(
+                "select count (br.id) from BaseRecord br join StructuredDocument sd on br.id ="
+                    + " sd.template.id and br.owner=:owner")
```
mKowalski256: AI is telling me this is the culprit of templates being deleted if used just by the template owner - there should be additional `AND sd.owner != :owner` clause to filter out docs that shouldn't trigger template transfer.

src/main/webapp/ui/src/eln/sysadmin/users/index.tsx
```diff
@@ -792,6 +793,15 @@ const DeleteAction = ({
...
+  }
```
mKowalski256: the AI tells me `getDeleteMessage` function is no longer called from anywhere (i.e. it's dead code)

src/main/java/com/researchspace/service/impl/TemplateTransferService.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+@Service
+public class TemplateTransferService implements TransferService {
+
+  public static final String DELETED_USER_TEMPLATES_FOLDER = "Deleted Users";
+  public static final String DELETED_USER_NAME_SUFFIX = "(Deleted)";
```
mKowalski256: I'm thinking if adding whitespace between username and suffix wouldn't be a good idea. It'd not only look nicer on screens displaying the username, but actually whitespace is normally not allowed in usernames, while other chars (like parenthesis) potentially are. So overriding createdBy etc. fields with username + " (Deleted)" would uniquely distinguish deleted accounts from ones where user just selected a fancy username.

src/main/java/com/researchspace/dao/hibernate/RecordDaoHibernate.java
```diff
@@ -647,4 +647,147 @@ public Long getRecordCountForUser(RecordTypeFilter recordFilter, User user) {
...
+                "UPDATE BaseRecord_AUD b SET owner_id=:newOwner_id,"
+                    + " createdBy=:updatedOriginalOwnerName WHERE id IN :ids")
+            .setParameter("newOwner_id", newOwner.getId())
+            .setParameter(
+                "updatedOriginalOwnerName",
+                null != updatedOriginalOwnerName
+                    ? updatedOriginalOwnerName
+                    : originalOwner.getUsername())
+            .setParameter("ids", templateIds);
+    query.executeUpdate();
```
mKowalski256, line 764: I think we shouldn't be touching audit tables at all in this functionality, unless I'm missing something?

src/main/java/com/researchspace/dao/hibernate/RecordGroupSharingDaoHibernateImpl.java
```diff
@@ -431,6 +431,18 @@ private String translateOrderBy(String orderBy) {
...
+    Session session = getSession();
+    Query<RecordGroupSharing> query =
+        session
+            .createQuery(
+                " from RecordGroupSharing rgs where rgs.shared.owner.id=:userId ",
+                RecordGroupSharing.class)
+            .setParameter("userId", user.getId());
+    List<RecordGroupSharing> rc = query.list();
+    return rc;
+  }
```
mKowalski256: this is unused code I think.

src/main/java/com/researchspace/dao/hibernate/RecordDaoHibernate.java
```diff
@@ -647,4 +647,147 @@ public Long getRecordCountForUser(RecordTypeFilter recordFilter, User user) {
...
+    query =
+        getSession()
+            .createNativeQuery(
+                "UPDATE BaseRecord br SET acl = (SELECT REPLACE(br.acl, :originalOwnerInMiddle,"
+                    + " :newOwnerInMiddle) ) WHERE id IN :ids")
+            .setParameter("originalOwnerInMiddle", "&" + originalOwner.getUsername() + "")
+            .setParameter("newOwnerInMiddle", "&" + newOwner.getUsername() + "")
+            .setParameter("ids", templateIds);
+    query.executeUpdate();
+  }
```
mKowalski256, line 790: I think we have acl permission recalculation code somewhere, but I'll trust these replacements ;)

mKowalski256 (changes requested):
I did round of review on [latest AWS build](https://rsdev-833-user-deletion-fails-with-shared-template-27.researchspace.com/), it seems to be almost there but I did noted some issues (explained in code comments).

I was running the manual test scenario that I've added to [RSDEV-833](https://researchspace.atlassian.net/browse/RSDEV-833) ticket, the code is currently failing at step 6., where it transfers template of the user whose shared template wasn't used by anyone other than the user themselves.

src/main/java/com/researchspace/dao/hibernate/RecordDaoHibernate.java
mrowe-tan: The change to owner_id is required to this constraint on BaseRecord_AUD:
`CONSTRAINT `FK_BaseRecord_AUD_owner_id` FOREIGN KEY (`owner_id`) REFERENCES `User` (`id`) ON DELETE NO ACTION ON UPDATE NO ACTION`
The correct option may be to change this constraint, but for the purposes of this ticket we'll stick with updating the owner_id to the new template owner.  This is safe since audit entries are removed when users are deleted anyway.  We will, however, not change the createdBy value any more.

tchoi-rs (approved):
Frontend part looks fine to me!

src/main/java/com/researchspace/dao/hibernate/RecordDaoHibernate.java
```diff
@@ -647,4 +647,144 @@ public Long getRecordCountForUser(RecordTypeFilter recordFilter, User user) {
...
+                    + " br.owner = :owner AND sd.template.id IS NOT NULL) AND br.owner != :owner")
+            .setParameter("owner", user)
+            .uniqueResult();
+    Long count = (Long) result;
+    return count > 0;
+  }
+
+  @SuppressWarnings("unchecked")
+  @Override
+  public List<BaseRecord> getTemplatesSharedByUserAndUsedByOtherUsers(User user) {
```
mKowalski256: better method name would be `getTemplatesOwnedByUserAndUsedByOtherUsers` (that's what the method is doing, it doesn't check shared status)

src/main/java/com/researchspace/dao/RecordDao.java
```diff
@@ -163,4 +164,25 @@ List<String> getTextDataFromOntologyFilesOwnedByUserIfSharedWithAGroup(
...
+   */
+  boolean hasUserSharedTemplatesUsedByOtherUsers(User u);
+
+  /**
+   * Gets all templates shared from a specific user.
+   *
+   * @param u user
+   * @return
+   */
+  List<BaseRecord> getTemplatesSharedByUserAndUsedByOtherUsers(User u);
```
mKowalski256: similarly here, method name could be updated, and javadocs should definitely by updated.

src/main/java/com/researchspace/service/RecordManager.java
```diff
@@ -656,4 +656,43 @@ RecordInformation decorateRecordInfo(
...
+   */
+  boolean hasUserSharedTemplatesUsedByOtherUsers(User u);
+
+  /**
+   * Gets all templates shared from a specific user.
+   *
+   * @param u user
+   * @return
+   */
+  List<BaseRecord> getTemplatesSharedByUserAndUsedByOtherUsers(User u);
```
mKowalski256: similarly here, with method name/javadocs update

mKowalski256 (changes requested):
When re-testing on [latest AWS build](https://rsdev-833-user-deletion-fails-with-shared-template-33.researchspace.com/), I was able to run the manual test scenario from RSDEV-833 fine, and I marked just a couple of places where method name (and java doc) should be updated.

But then I found a problem with deletion that maybe is a case missing from the original specification. When user shared a template used by others that has a gallery attachment, sysadmin attempt to delete this user fails with constraint error like this:
```
ERROR 11 May 2026 07:19:44,581 - SqlExceptionHelper.logExceptions | Cannot delete or update a parent row: a foreign key constraint fails (`rspace`.`FieldAttachment`, CONSTRAINT `FK2478505F3319B855` FOREIGN KEY (`mediafile_id`) REFERENCES `EcatMediaFile` (`id`) ON DELETE NO ACTION ON UPDATE NO ACTION)
``` 

I debugged a bit, if I understand correctly the code transfering the template to sysadmin does run as expected. The constraint error may come from the fact that deletion tries to subsequently clear original owner's Gallery, and fails when attempting to remove files attached to the template. It is probably that we don't need just transfer a template document to sysadmin, but also transfer all Gallery items used in the template, so these are not deleted on user deletion (the constraint that blocks deletion is correct). 

I updated the test scenario RSDEV-833 to include file attachment. Let me know if you need more details to reproduce the issue.

mrowe-tan:
@greptileai

src/main/webapp/ui/src/eln/sysadmin/users/index.tsx
```diff
@@ -858,22 +859,24 @@ const DeleteAction = ({
...
+                          <strong>
+                            the owner of Forms and/or Templates that are used by other users.
+                          </strong>
+                          To ensure continued access to these Forms/Templates, the system
+                          <strong> will transfer ownership</strong> of those files to
+                          <strong> this System Administrator</strong> account. Forms and
+                          Templates that are not used by others will be deleted.
+                        </Typography>
+                      </Alert>
+                  )}
```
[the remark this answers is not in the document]
mrowe-tan: The Alert cannot be moved out of the DialogContentText since that would affect the rest of the dialog.  Also I see no errors in the console.  Since this is a sysadmin dialog also I'll ignore this for now.

src/main/webapp/ui/src/eln/sysadmin/users/index.tsx
```diff
@@ -858,22 +859,24 @@ const DeleteAction = ({
               <DialogTitle>Deletion Confirmation</DialogTitle>
               <DialogContent>
                 <DialogContentText variant="body2" sx={{ mb: 2 }}>
+                  {(user.hasFormsUsedByOtherUsers || user.hasTemplatesUsedByOtherUsers) && (
+                      <Alert severity="info" sx={{ mb: 1 }}>
+                        <Typography variant="body2">
+                          The user you are trying to delete is{" "}
+                          <strong>
+                            the owner of Forms and/or Templates that are used by other users.
+                          </strong>
```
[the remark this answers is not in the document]
mrowe-tan: Fair observation, this will be implemented.

src/main/webapp/ui/src/eln/sysadmin/users/index.tsx
```diff
@@ -910,11 +913,7 @@ const DeleteAction = ({
                   type="submit"
                   loading={false}
                   disabled={false}
-                  label={
-                    user.hasFormsUsedByOtherUsers
-                      ? "Transfer Forms And Delete"
-                      : "Delete"
-                  }
+                  label={(user.hasFormsUsedByOtherUsers || user.hasTemplatesUsedByOtherUsers) ? "Transfer Forms/Templates And Delete" : "Delete"}
```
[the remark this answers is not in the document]
mrowe-tan: This was indeed my original implementation, but after discussion with Vaida and Tim it was decided to keep the text succinct, as it currently is.  I will thus not be implementing this change.

src/main/java/com/researchspace/dao/hibernate/RecordDaoHibernate.java
```diff
@@ -647,4 +647,204 @@ public Long getRecordCountForUser(RecordTypeFilter recordFilter, User user) {
...
+
+  /**
+   * Transfers ownership of templates from one user to another user.
+   *
+   * @param originalOwner - Original owner of the templates
+   * @param newOwner - New owner of the templates
+   * @param templateIds - IDs of the templates to be transferred
+   * @param updatedOriginalOwnerName - An updated name, if any, for the original owner. (e.g. In the
+   *     case the user has been deleted and updated with a suffix to denote that.)
+   */
```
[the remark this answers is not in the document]
mrowe-tan: Fair point.

src/main/webapp/ui/src/eln/sysadmin/users/index.tsx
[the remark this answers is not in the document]
mrowe-tan: See as for Copilot review, above.

src/main/java/com/researchspace/dao/hibernate/RecordDaoHibernate.java
```diff
@@ -647,4 +647,204 @@ public Long getRecordCountForUser(RecordTypeFilter recordFilter, User user) {
...
+  public void transferTemplates(
+      User originalOwner, User newOwner, List<Long> templateIds, String updatedOriginalOwnerName) {
```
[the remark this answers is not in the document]
mrowe-tan: As for Copilot, above.

src/main/java/com/researchspace/service/impl/TemplateTransferService.java
```diff
@@ -0,0 +1,173 @@
...
+
+  private List<String> buildRelativePath(Long recordId, Long galleryRootId) {
+    Folder parent = folderDao.getParentFolder(recordId);
+    if (parent == null || parent.getId().equals(galleryRootId)) {
+      return new ArrayList<>();
+    }
+    List<String> path = buildRelativePath(parent.getId(), galleryRootId);
+    path.add(parent.getName());
+    return path;
+  }
```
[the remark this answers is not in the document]
mrowe-tan: Resolved by Claude command line.

src/main/java/com/researchspace/service/impl/TemplateTransferService.java
```diff
@@ -0,0 +1,173 @@
...
+
+      String deletedUserName = originalOwner.getUsername() + DELETED_USER_NAME_SUFFIX;
+      recordManager.transferTemplates(originalOwner, newOwner, templateIds, deletedUserName);
+
+      transferGalleryItemsForTemplates(originalOwner, newOwner, templateIds);
+
+      String description =
+          String.format(
+              "Ownership of template transferred from %s to %s",
+              originalOwner.getUsername(), newOwner.getUsername());
```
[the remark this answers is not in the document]
mrowe-tan: Resolved by Claude command line (with a tiny manual fix by me).

src/main/java/com/researchspace/dao/hibernate/UserDeletionDaoHibernate.java
```diff
@@ -631,6 +631,16 @@ private void deleteRecords(Long userId, Session session) {
     executeDeleteByRecordOwner(userId, session, FIELD, STRUCTURED_DOCUMENT_ID);
     executeDeleteByRecordOwner(userId, session, "Field_AUD", STRUCTURED_DOCUMENT_ID);
 
+    // Any templates created by this user and used by other users will have
+    // been transferred to a sysadmin account by this point, but we need to
+    // null any references that the user makes to their own templates here
+    execute(
+        userId,
+        session,
+        "update StructuredDocument sdA join BaseRecord brA set template_id = null where sdA.id in"
```
mKowalski256: Notice that the update here uses 'join' between sdA and brA but without 'on' clause in first line. That gives cartesian product which is later limited by subqueries, and eventually I think the result is fine, but possibly some edge case document may end up with template_id being nullified? Adding `on brA.id = sdA.id` after join would sort it, so that's easiest fix.

----

AI also suggested to me that another, simpler query would do the job, although in subtly different way:
```
update StructuredDocument sd join BaseRecord br on br.id = sd.id
  set sd.template_id = null
  where br.owner_id = :id and sd.template_id is not null
```
- this is just clearing `template_id` connection on all documents belonging to the user, which are about to be deleted anyway as part of user deletion. That's not exactly equivalent of your original query which clears `template_id` only for documents linking non-transferred templates (as per comment in the code), but could be better, i.e. I find it much more readable. If you decide to go with simpler query, please also update the comment in lines 634-636.

mKowalski256 (changes requested):
The code seems mostly fine - I've commented on one SQL which should probably be updated. I was able to go through manual deletion scenario from RSDEV-833 fine.

One thing that is definitely missing in the PR is the IT-level unit test. We have `UserDeletionManagerTestIT` class that is covering various user-deletion scenarios on real database (i.e. where the FK constraints would block the deletion), and that should be expanded to cover scenario of user having a template used by other.

src/main/java/com/researchspace/dao/hibernate/UserDeletionDaoHibernate.java
mrowe-tan: I've added the simplified SQL and updated the comments as requested.

src/main/java/com/researchspace/dao/hibernate/RecordDaoHibernate.java
```diff
@@ -647,4 +647,215 @@ public Long getRecordCountForUser(RecordTypeFilter recordFilter, User user) {
...
+            .setParameter("ids", templateIds);
+    query.executeUpdate();
+
+    query =
+        getSession()
+            .createNativeQuery(
+                "UPDATE BaseRecord br SET acl = (SELECT REPLACE(br.acl, :originalOwnerInMiddle,"
+                    + " :newOwnerInMiddle) ) WHERE id IN :ids")
+            .setParameter("originalOwnerInMiddle", "&" + originalOwner.getUsername() + "")
+            .setParameter("newOwnerInMiddle", "&" + newOwner.getUsername() + "")
```
mKowalski256: Greptile pointed to it in latest comment, and my AI seems to believe there is indeed the issue in combined two-stage ACL update. Basically it says that username string should be anchored on right by `+ "="`, similarly as in first ACL update (above). Can you double-check?

mrowe-tan: Yes, that would be safer.  I've made the update now.

mKowalski256 (approved):

PR #750, opened 2026-04-30
RSDEV-1084: Fix stoichiometry import revision — post-import Envers fixup

nhanlon2:
After XML archive import, stoichiometry references in field HTML had revision=null because Envers audit records are not available during the import transaction. Frontend now requires revision to be a number, causing broken stoichiometry display after import.

Add StoichiometryImportRevisionFixupManager that runs after import transaction commits, queries Envers for actual revisions, and rewrites field HTML with correct revision numbers.

- Create interface + impl in com.researchspace.service.archive
- Wire into ExportImportImpl.doImport() after archiveImporter returns
- Add 8 unit tests covering all paths (null fixup, multi-stoich, Envers failure, existing revision skip)
- Update integration test to assert non-null revision from DB

## Description ##
***REQUIRED***

## Design decisions
***OPTIONAL***

## Testing notes
***OPTIONAL***

### DIFF
--- a/src/main/java/com/researchspace/service/archive/StoichiometryImportRevisionFixupManagerImpl.java
+++ b/src/main/java/com/researchspace/service/archive/StoichiometryImportRevisionFixupManagerImpl.java
@@ -70,9 +70,6 @@ public class StoichiometryImportRevisionFixupManagerImpl
     String updatedData = field.getFieldData();
 
     for (StoichiometryDTO dto : stoichiometries) {
-      if (dto.getRevision() != null) {
-        continue;
-      }
       Long stoichId = dto.getId();
       AuditedEntity<Stoichiometry> audited =
           auditManager.getNewestRevisionForEntity(Stoichiometry.class, stoichId);
--- a/src/main/java/com/researchspace/service/archive/StoichiometryImportRevisionFixupManagerImpl.java
+++ b/src/main/java/com/researchspace/service/archive/StoichiometryImportRevisionFixupManagerImpl.java
@@ -33,7 +33,7 @@ public class StoichiometryImportRevisionFixupManagerImpl
 
   @Override
   public void fixupStoichiometryRevisions(ImportArchiveReport report, User user) {
-    if (!report.isSuccessful()) {
+    if (report.getImportedRecords().isEmpty()) {
       return;
     }
     for (BaseRecord record : report.getImportedRecords()) {
--- a/src/main/java/com/researchspace/service/archive/StoichiometryImportRevisionFixupManagerImpl.java
+++ b/src/main/java/com/researchspace/service/archive/StoichiometryImportRevisionFixupManagerImpl.java
@@ -55,10 +55,10 @@ public class StoichiometryImportRevisionFixupManagerImpl
         fixupField(field, user);
       } catch (Exception e) {
         log.warn(
-            "Failed to fixup stoichiometry revisions for field id={} in record id={}: {}",
+            "Failed to fixup stoichiometry revisions for field id={} in record id={}",
             field.getId(),
             recordId,
-            e.getMessage());
+            e);
       }
     }
   }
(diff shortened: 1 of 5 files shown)


nhanlon2:
Testing manually for correctness and ONLY copilot assigned to review until verified

src/main/java/com/researchspace/service/archive/StoichiometryImportRevisionFixupManagerImpl.java
```diff
@@ -0,0 +1,97 @@
...
+import org.springframework.stereotype.Service;
+
+@Service("stoichiometryImportRevisionFixupManager")
+public class StoichiometryImportRevisionFixupManagerImpl
+    implements StoichiometryImportRevisionFixupManager {
+
+  private static final Logger log =
+      LoggerFactory.getLogger(StoichiometryImportRevisionFixupManagerImpl.class);
+
+  private static final String STOICHIOMETRY_ATTR = "data-stoichiometry-table";
```
rs-fraser, line 27: As of last week there's a new front-end `data-stoichiometry-table-only` (holding "empty" stoich tables that aren't attached to a chemical) attribute which will also need handled here (and anywhere else `data-stoichiometry-table` appears). Probably best as a separate ticket though.

src/test/java/com/researchspace/service/archive/StoichiometryImportRevisionFixupManagerTest.java
```diff
@@ -0,0 +1,202 @@
...
+  @Mock private AuditManager auditManager;
+  @Mock private FieldManager fieldManager;
+  @Mock private User user;
+  @Mock private ImportArchiveReport report;
+
+  @InjectMocks private StoichiometryImportRevisionFixupManagerImpl testee;
+
+  @BeforeEach
+  void setUp() {
+    Mockito.lenient().when(report.isSuccessful()).thenReturn(true);
```
rs-fraser: I don't think this is used. When I remove the `lenient()` all 8 tests fail with `UnnecessaryStubbingException`

src/test/java/com/researchspace/service/archive/StoichiometryImportRevisionFixupManagerTest.java
```diff
@@ -0,0 +1,202 @@
...
+
+  private StructuredDocument mockStructuredDocument(long id) {
+    StructuredDocument doc = Mockito.mock(StructuredDocument.class);
+    when(doc.getId()).thenReturn(id);
+    return doc;
+  }
+
+  @Test
+  void whenReportNotSuccessfulStillDoesFixup() {
+    Mockito.lenient().when(report.isSuccessful()).thenReturn(false);
```
rs-fraser: this line's not needed.

rs-fraser (approved):
Looks good. As discussed we now need to handle the new `data-stoichiometry-table-only` attribute but will handle that separately. Also added a couple comments about `lenient`

src/main/java/com/researchspace/service/archive/StoichiometryImportRevisionFixupManagerImpl.java
nhanlon2: Yes - RSdev-1091 is handling this and further tickets may also be needed after that is done.

src/test/java/com/researchspace/service/archive/StoichiometryImportRevisionFixupManagerTest.java
nhanlon2: removed

src/test/java/com/researchspace/service/archive/StoichiometryImportRevisionFixupManagerTest.java
nhanlon2: removed

PR #751, opened 2026-04-30
Rsdev 1087 galaxy html pdf word export

nhanlon2:
## Description ##
***REQUIRED***

## Design decisions
***OPTIONAL***

## Testing notes
***OPTIONAL***

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,6 @@ samples/
 .factorypath
 .java-version
 src/main/webapp/ui/junit.xml
+
+# other tooling
+/.claude/
--- a/.gitignore
+++ b/.gitignore
@@ -63,6 +63,3 @@ samples/
 .factorypath
 .java-version
 src/main/webapp/ui/junit.xml
-
-# other tooling
-/.claude/
--- a/src/main/java/com/researchspace/export/pdf/HTMLStringGeneratorForExport.java
+++ b/src/main/java/com/researchspace/export/pdf/HTMLStringGeneratorForExport.java
@@ -6,6 +6,7 @@ import static org.apache.commons.text.StringEscapeUtils.escapeHtml4;
 import com.researchspace.archive.ArchivalNfsFile;
 import com.researchspace.archive.model.ArchiveModelFactory;
 import com.researchspace.core.util.FieldParserConstants;
+import com.researchspace.export.externalworkflows.ExternalWorkflowHtmlGenerator;
 import com.researchspace.linkedelements.FieldContents;
 import com.researchspace.linkedelements.FieldParser;
 import com.researchspace.model.EcatCommentItem;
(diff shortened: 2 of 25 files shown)


nhanlon2:
For review by copilot

.gitignore
```diff
@@ -63,3 +63,4 @@ samples/
 .factorypath
 .java-version
 src/main/webapp/ui/junit.xml
+.github
```
mKowalski256: we have some github-project configs in this location, I think it probably should stay as not ignored

src/main/java/com/researchspace/export/pdf/HTMLStringGeneratorForExport.java
```diff
@@ -235,6 +251,15 @@ private Set<String> appendListsOfMaterials(
     return igsnInventoryLinkedItems;
   }
 
+  private void appendExternalWorkflowData(StringBuffer sbf, Field field,Set<ExternalWorkFlowData> externalWorkFlowData ) {
+    if (field.getId() == null) {
+      return;
+    }
+    Set<ExternalWorkFlowData> externalWorkFlowDataMatchingField =
+       externalWorkFlowData.stream().filter(data -> data.getRspacecontainerid() == field.getId()).collect(java.util.stream.Collectors.toSet());
+    sbf.append(externalWorkflowHtmlGenerator.getHtmlForExternalWorkflowData(externalWorkFlowData));
```
mKowalski256: Claude says that is the line causing results to be added to every field, i.e. the code filters by field but then passes unfiltered variable  (`externalWorkFlowData` rather than `externalWorkFlowDataMatchingField`).

src/main/resources/velocityTemplates/archive/documentView.vm
```diff
@@ -24,6 +24,7 @@
     <div style="width:960px" >
     $field.fieldDataPrintable
     </div>
+     $externalWorkflowHtmlGenerator.getHtmlForArchivedExternalWorkflowData($field.externalWorkFlowData, $field.externalWorkFlowInvocations)
```
mKowalski256: Claude says this line stops HTML archive from having Galaxy tables, the method doesn't exist on `ExternalWorkflowHtmlGenerator` class.

mKowalski256 (changes requested):
I've checked on [AWS build](https://rsdev-1087-galaxy-html-pdf-word-export-8.researchspace.com/) and I've run AI-supported review, which pointed to a couple of issues (I marked specific codelines):

1. when exporting multi-field ELN doc (e.g. created out of Experiment form) that has  galaxy workflow data, the table is printed for every field rather than just one that should have data attached

2. html export doesn't actually produce the table, just prints some code

Image

.gitignore
nhanlon2: removed

src/main/java/com/researchspace/export/pdf/HTMLStringGeneratorForExport.java
nhanlon2: fixed

src/main/resources/velocityTemplates/archive/documentView.vm
nhanlon2: fixed

PR #755, opened 2026-05-04
Re-compress all images

tchoi-rs:
## Description ##
Re-compresses all RSpace images to reduce download sizes.


mKowalski256 (changes requested):
I'd feel more comfortable if we limit the changset a bit:

1. don't change `src/main/resources/sqlUpdates`, I'm not sure if that won't break things
2. don't change `src/test/resources/TestResources/archives` folder, this is historical test data
3. `src/main/webapp/scripts/bower_components` contains 3rd party libs, do we really want to recompress images there? maybe we do.
4. I'd skip recompression of .svg files in general, the filesize gain is minimal for these files while we risk things breaking in subtle way

There are also a couple of tests broken due to size changes in TestResources:
```
Error:  Failures: 
Error:    PublicControllerTest.testBanner:118->assertDefaultImage:150 expected:<41430> but was:<14166>
Error:    PublicControllerTest.testBannerWithWrongFileTypeReturnsDefault:95 expected:<41430> but was:<14166>
```

tchoi-rs:
1. For the SQL Update change, it's for a historical version (1.42) and it sets the image thumbnails for rows that don't have them. It should not affect any modern versions of RSpace.
2. If they are historical and not used in modern RSpace, maybe they should be removed outright?
3. Yes, as they are displayed in the browser (e.g. TinyMCE icons). It's easier to compress everything than selectively figure out which ones are being used and which ones aren't.
4. The SVG updates uses [svgcleaner](https://github.com/razrfalcon/svgcleanerhttps://github.com/razrfalcon/svgcleaner) to remove redundant information without losing functionality.
5. I don't understand why we need the banner property - it regurgitates a config variable and if it's broken then it's the sysadmin's fault. The related controller should be simplified.

mKowalski256:
> 1. For the SQL Update change, it's for a historical version (1.42) and it sets the image thumbnails for rows that don't have them. It should not affect any modern versions of RSpace.

ok.

> 2. If they are historical and not used in modern RSpace, maybe they should be removed outright?

The files in `src/test/resources/TestResources/archives` are used by integration tests covering backward compatibility of export-import mechanism i.e. that xml archives exported in historical versions of RSpace are (hopefully) importable in all later RSpace versions. That's one place where we actually want historical test data to stay historical. Possibly recompressing images wouldn't break anything there, but in principle I'd avoid any changes in that dir.

> 3. Yes, as they are displayed in the browser (e.g. TinyMCE icons). It's easier to compress everything than selectively figure out which ones are being used and which ones aren't.
> 4. The SVG updates uses [svgcleaner](https://github.com/razrfalcon/svgcleanerhttps://github.com/razrfalcon/svgcleaner) to remove redundant information without losing functionality.

ok.

> 5. I don't understand why we need the banner property - it regurgitates a config variable and if it's broken then it's the sysadmin's fault. The related controller should be simplified.

Customers often want logo of their institution being displayed, that's why we provide it dynamically through `/banner` url. We do provide default RSpace logo, but if sysadmin breaks it it's accepted to be their fault.

mKowalski256:
Also let's update one more test assertion in full unit test suite ([jenkins run](https://jenkins.researchspace.com/job/rspace-web/job/chore-recompress-images/)):

```
[ERROR] Failures: 
[ERROR]   ContainerApiManagerTest.saveNewContainerWithImage:846 expected: <102469> but was: <87958>
```

PR #760, opened 2026-05-06
fix-PRT-1070: ketcher mounts dialog on each button press

nhanlon2:
## Description ##
***REQUIRED***

## Design decisions
***OPTIONAL***

## Testing notes
***OPTIONAL***

### DIFF
--- a/src/main/webapp/ui/src/tinyMCE/ketcher/KetcherTinyMce.tsx
+++ b/src/main/webapp/ui/src/tinyMCE/ketcher/KetcherTinyMce.tsx
@@ -48,7 +48,6 @@ const EMPTY_MOLECULE_MESSAGE =
 const ketcherDialogRoots = new WeakMap<HTMLElement, Root>();
 
 let ketcherDialogListenerRegistered = false;
-let ketcherDialogRenderCount = 0;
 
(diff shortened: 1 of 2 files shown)


src/main/webapp/ui/src/tinyMCE/ketcher/KetcherTinyMce.tsx
```diff
@@ -130,13 +131,14 @@ function renderKetcherDialog(): void {
...
 
+  ketcherDialogRenderCount++;
   const root = ketcherDialogRoots.get(wrapperDiv) ?? createRoot(wrapperDiv);
   ketcherDialogRoots.set(wrapperDiv, root);
   root.render(
     <ThemeProvider theme={theme}>
       <Analytics>
         <Alerts>
-          <KetcherTinyMce />
+          <KetcherTinyMce key={ketcherDialogRenderCount} />
```
tchoi-rs: This doesn't look semantically correct; instead we should pass in a callback so that when Cancel is called, `root.unmount()` is called.

tchoi-rs (changes requested):

tchoi-rs (approved):

PR #765, opened 2026-05-12
Set npm minimum release age

tchoi-rs:
## Description ##
Sets the minimum release age of npm package installs to 7 days to mitigate most supply chain attacks.


mKowalski256 (approved):
I understand we can still pin the specific newer version if needed?

tchoi-rs:
@mKowalski256 No, but we can add exceptions.

PR #776, opened 2026-05-20
RSDEV-1109: 'move to s3' dialog for gallery items

mKowalski256:
Implemented by Claude Code AI agent on behalf of mKowalski256.

Implements core S3 write operations and the "Move to S3" gallery action (RSDEV-1109).

Changes
- S3NfsClient.java — upload, server-side copy (5 GB guard, collision check), delete, per-object S3 metadata via WriteAttribution
- WritableNfsClient.java — new sub-interface replacing supportWritePermission(); adds supportsServerSideTransfer() capability flag
- FilestoreWriteManagerImpl.java — uploadToFilestore, transferBetweenFilestores (S3↔S3 only)
- GalleryFilestoresApiController.java — POST /move, /copy, /transfer; old /irods/{copy,move} marked `@Deprecated` and delegated
- MoveToS3.tsx — dialog supporting upload mode (Gallery→S3) and transfer mode (S3↔S3); retain-copy checkbox
- useS3Filestores.ts — hook fetching S3 filestores, exposing move/copy/transfer operations
- ActionsMenu.tsx — "Move to S3" menu item; auto-selects transfer mode for S3 filestore items
- rspace_api_specs_2_23_0.yaml — Swagger docs for all new endpoints; old endpoints marked deprecated


src/main/java/com/researchspace/service/aws/impl/S3UtilitiesImpl.java
```diff
@@ -293,12 +305,46 @@ public DeleteObjectResponse deleteFromS3(String folderPath, String fileName) {
...
+      CopyObjectRequest.Builder builder =
+          CopyObjectRequest.builder()
+              .sourceBucket(sourceBucket)
+              .sourceKey(sourceKey)
+              .destinationBucket(bucketName)
+              .destinationKey(destKey);
+      if (metadata != null && !metadata.isEmpty()) {
+        builder.metadata(metadata).metadataDirective(MetadataDirective.REPLACE);
+      }
+      s3Client.copyObject(builder.build());
```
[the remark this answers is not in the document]
mKowalski256: the code does compile and works - closing.

src/test/java/com/researchspace/service/aws/impl/S3UtilitiesTest.java
```diff
@@ -93,4 +127,61 @@ public void listFolderContents_paginatesWhenTruncated() throws Exception {
...
+
+    verify(mockS3Client)
+        .copyObject(
+            argThat(
+                (CopyObjectRequest r) ->
+                    r != null
+                        && "source-bucket".equals(r.sourceBucket())
+                        && "src/file.txt".equals(r.sourceKey())
+                        && "dest-bucket".equals(r.destinationBucket())
+                        && "dst/file.txt".equals(r.destinationKey())));
```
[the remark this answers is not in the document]
mKowalski256: the test code does compile and works - closing.

src/main/webapp/ui/src/eln/gallery/components/ActionsMenu.tsx
```diff
@@ -937,6 +973,58 @@ function ActionsMenu({
...
+          <AccentMenuItem
+            title="Move to S3"
+            subheader={moveToS3Allowed
+              .get()
+              .map(() => "")
+              .orElseGet(([e]) => e.message)}
+            backgroundColor={IRODS_COLOR.background}
+            avatarBackgroundColor="#ffffff"
+            foregroundColor={IRODS_COLOR.contrastText}
+            avatar={<CardMedia image={S3Logo} />}
```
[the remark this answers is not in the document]
mKowalski256: This was deliberate - S3 row should use the same greenish tint as iRODS, with white only behind the avatar box.

src/main/webapp/ui/src/eln/gallery/components/MoveToS3.spec.tsx
```diff
@@ -0,0 +1,562 @@
...
+        await Then["the transfer button should be enabled"]();
+      },
+    );
+
+    feature(
+      "Clicking Transfer calls the transfer endpoint and shows a success alert",
+      async ({ Given, When, Then }) => {
+        await Given["the Move to S3 dialog in transfer mode is mounted"]();
+        await Then["the dialog title 'Transfer to S3' should be visible"]();
+        await When[
```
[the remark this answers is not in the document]
mKowalski256: fixed.

src/main/webapp/ui/src/eln/gallery/components/useS3Filestores.ts
```diff
@@ -0,0 +1,292 @@
...
+                    }),
+                ),
+              ).map((details) =>
+                mkAlert({
+                  variant: "warning",
+                  message: "Moving some files to S3 failed.",
+                  details,
+                  isInfinite: true,
+                }),
+              ),
```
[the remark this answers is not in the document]
mKowalski256: should be fixed now

src/main/java/com/researchspace/service/impl/FilestoreWriteManagerImpl.java
```diff
@@ -0,0 +1,257 @@
...
+    String sourceFilestoreRootPath = resolveAbsoluteFilestorePath(sourceFilestore);
+    String destFilestoreRootPath = resolveAbsoluteFilestorePath(destFilestore);
+    String absoluteSourcePath =
+        StringUtils.isBlank(sourceFilestoreRootPath)
+            ? request.getSourcePath()
+            : sourceFilestoreRootPath + "/" + StringUtils.stripStart(request.getSourcePath(), "/");
+    String absoluteDestPath =
+        StringUtils.isBlank(destFilestoreRootPath)
+            ? request.getDestPath()
+            : destFilestoreRootPath + "/" + request.getDestPath();
```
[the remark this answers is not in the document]
mKowalski256: I've committed the fix.

src/main/java/com/researchspace/netfiles/WritableNfsClient.java
```diff
@@ -0,0 +1,134 @@
...
+  default ApiExternalStorageOperationResult uploadFilesToNfs(
+      String destinationPath, Map<Long, File> mapRecordIdToFile) {
+    ApiExternalStorageOperationResult result = new ApiExternalStorageOperationResult();
+    for (Map.Entry<Long, File> entry : mapRecordIdToFile.entrySet()) {
+      Long recordId = entry.getKey();
+      File file = entry.getValue();
+      try {
+        uploadFile(file, destinationPath);
+        result.add(new ApiExternalStorageOperationInfo(recordId, null, file.getName(), true));
+      } catch (Exception e) {
```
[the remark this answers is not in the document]
mKowalski256: this is deliberate here; the external file systems generally don't provide valid unique persistent `externalStorageId` -  only iRODS does that. So for s3 there is no id to save. We could potentially save target file path in `ExternalStorageLocation`, but I don't see any strong reason to track that - filestore paths are fragile and often broken, we only save them when user decides to link filestore record in ELN document, as we have no other choice. Just tracking the uploaded paths seems unnecessary.

src/main/java/com/researchspace/service/impl/FilestoreWriteManagerImpl.java
```diff
@@ -0,0 +1,257 @@
...
+        succeededMediaFiles.add(mediaFile);
+
+        if (info.getExternalStorageId() != null) {
+          ExternalStorageLocation loc = new ExternalStorageLocation();
+          loc.setFileStore(filestore);
+          loc.setOperationUser(user);
+          loc.setExternalStorageId(info.getExternalStorageId());
+          loc.setConnectedMediaFile(mediaFile);
+          externalStorageManager.saveExternalStorageLocation(loc);
+        }
```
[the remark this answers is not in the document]
mKowalski256: per previous comment, this behaviour is fine.

src/main/webapp/ui/src/eln/gallery/components/ActionsMenu.tsx
```diff
@@ -900,43 +943,99 @@ function ActionsMenu({
...
-              .map(({ id }) => idToString(id)),
-          )
-            .map((selectedIds) => (
-              <MoveToIrods
-                key={selectedIds.join(",")}
-                selectedIds={selectedIds}
-                dialogOpen={irodsOpen}
-                setDialogOpen={(newState) => {
-                  setIrodsOpen(newState);
+          {showNetfileActions && (
```
mKowalski256, line 946: adding `showNetFileActions` check here is to fix inconsistency in new Gallery, which was showing Filestores subgallery button only if `netfilestores.enabled` is `true` in deployment properties, but was showing 'move to iRODS' (and now also 'move to S3') actions regardless of that property.

Hiding the 'move to ..' actions on servers that don't have filestores configured makes more sense and is how it worked in old Gallery (i.e. 'move to iRODS' was not visible unless `netfilestores.enabled=true`).

src/main/webapp/ui/src/eln/gallery/components/MoveToS3.tsx
```diff
@@ -0,0 +1,329 @@
...
+import ValidatingSubmitButton from "../../../components/ValidatingSubmitButton";
+import Result from "../../../util/result";
+import AnalyticsContext from "../../../stores/contexts/Analytics";
+import { ACCENT_COLOR } from "../../../assets/branding/s3";
+import useS3Filestores, {
+  type S3Filestore,
+  type S3TransferSource,
+} from "./useS3Filestores";
+
+const CustomDialog = styled(Dialog)(() => ({
```
tchoi-rs: Please do not use `styled` in new code

mKowalski256:
Should be good to review now! Test instance running at https://rsdev-1109-move-to-s3-21.researchspace.com - has a couple s3 file systems added (through Sysadmin UI), and also iRODS connection (for testing refactored 'move to iRODS') optons.

@tchoi-rs could you please review the front-end changes, the summary:
* The Gallery actions menu gains a "Move to S3" item, sitting alongside the existing "Move to iRODS" option and following the same enable/disable rules (available for regular Gallery items; blocked for system folders).
*  Clicking it opens a dialog that fetches the user's configured S3 filestores from the API and presents a destination dropdown. A retain-copy checkbox controls whether the operation is a move or a copy. If no S3 filestores are configured, an empty-state message is shown instead. 
*  When the selected file is already inside an S3 filestore (i.e. a remote file being browsed), the same "Move to S3" action opens the dialog in transfer mode instead: the source filestore is excluded from the destination dropdown, and submitting calls the `/transfer` endpoint rather than `/move` or `/copy`.
* A new useS3Filestores hook handles fetching the filestore list and wraps the three API calls (move, copy, transfer), including per-file success/failure parsing from the response and displaying appropriate alerts

@rs-nicof could you please review the back-end changes:
* The S3 filestore connector was previously read-only. This branch adds write support across three new REST endpoints under `/api/v1/gallery/filestores`:
  - POST `/{id}/move` and `/copy` — upload RSpace Gallery items to an S3 filestore. Each S3 object gets user-defined metadata attached (uploader username, RSpace record ID and name)
  - POST `/{id}/transfer` — server-side S3↔S3 copy between two filestores using AWS CopyObject (no data streams through RSpace). A `deleteSource` flag controls move vs. copy semantics. 
* The old /api/v1/gallery/irods/{copy,move} endpoints are kept but marked @Deprecated and delegate to the new handler - that's so we have unified API that can handle different filestores now (iRODS + S3) and in the future
* On the Java side, the supportWritePermission() boolean on NfsClient was replaced with WritableNfsClient sub-interface. S3NfsClient (previously read-only) and IRODSClient both implement it. A supportsServerSideTransfer() capability flag on the interface replaces instanceof S3NfsClient checks in the service layer

tchoi-rs (approved):
All looks fine (it's not using the latest frontend standards but I think it's fine to leave as is)

src/main/webapp/ui/src/eln/gallery/components/MoveToS3.tsx
mKowalski256: thank you @tchoi-rs, is https://github.com/rspace-os/rspace-web/pull/776/commits/c7795b2b00d3885756bab76e6d018e0d5f2749df the right fix?  (I updated sibling 'move to iRODS' dialog too)

src/main/java/com/researchspace/netfiles/irods/IRODSClient.java
```diff
@@ -192,8 +192,21 @@ public NfsFolderDetails queryForNfsFolder(NfsTarget nfsTarget) throws NfsExcepti
...
+  public String uploadFile(File source, String destDirectoryPath) {
+    throw new UnsupportedOperationException(
+        "IRODSClient.uploadFile primitive not yet wired up; use uploadFilesToNfs for now");
+  }
+
+  @Override
+  public void deleteFile(String absolutePath) {
+    throw new UnsupportedOperationException(
+        "IRODSClient.deleteFile primitive not yet wired up; use deleteFilesFromNfs for now");
+  }
```
rs-nicof, line 204: Should we possibly create a (tech debt) jira not to forget to fill the GAP by wiring up those 2 methods and use the interface in the proper way?

src/main/java/com/researchspace/netfiles/s3/S3NfsClient.java
```diff
@@ -157,6 +161,119 @@ private static String getParentPath(String path) {
...
+        && sourceDetails.getSizeInBytes() > S3_SINGLE_OP_COPY_LIMIT_BYTES) {
+      throw new IOException(
+          "Source object size "
+              + sourceDetails.getSizeInBytes()
+              + " bytes exceeds S3 single-op CopyObject limit of "
+              + S3_SINGLE_OP_COPY_LIMIT_BYTES
+              + " bytes (5 GB); multipart copy is not yet supported");
+    }
+    S3NfsClient destS3 = (S3NfsClient) destClient;
+    S3FolderContentItem destDetails = destS3.s3Utilities.getObjectDetails(destKey);
```
rs-nicof: This looks a bit strange to call an inner parameter's method.
why not creating a method in `S3NfsClient.getObjectDetails(destKey)` that calls the internal utilities? 

Have you possibly thought about making a Spring service and wire it up? 🤔

src/main/java/com/researchspace/service/impl/FilestoreWriteManagerImpl.java
```diff
@@ -0,0 +1,257 @@
...
+          new ObjectError(
+              "sourceFilestore", "Could not find file store with id: " + sourceFilestoreId));
+    }
+    if (destFilestore == null) {
+      errors.addError(
+          new ObjectError(
+              "destFilestore",
+              "Could not find file store with id: " + request.getDestFilestoreId()));
+    }
+    throwBindExceptionIfErrors(errors);
```
rs-nicof, line 139: Should we also check the case `sourceFilestore == destFilestore`?

src/main/java/com/researchspace/service/impl/FilestoreWriteManagerImpl.java
```diff
@@ -0,0 +1,257 @@
...
+            ? request.getSourcePath()
+            : sourceFilestoreRootPath + "/" + StringUtils.stripStart(request.getSourcePath(), "/");
+    String absoluteDestPath =
+        StringUtils.isBlank(destFilestoreRootPath)
+            ? request.getDestPath()
+            : destFilestoreRootPath + "/" + StringUtils.stripStart(request.getDestPath(), "/");
+    ApiExternalStorageOperationResult result = new ApiExternalStorageOperationResult();
+    try {
+      sourceClient.copyObject(
+          absoluteSourcePath, destClient, absoluteDestPath, attribution.metadataForRecord(null));
```
rs-nicof, line 171: Should we also check the case `sourcePath == destPath`?

src/main/java/com/researchspace/service/impl/FilestoreWriteManagerImpl.java
```diff
@@ -0,0 +1,257 @@
...
+    WriteAttribution attribution = new WriteAttribution(user.getUsername(), recordNames);
+
+    ApiExternalStorageOperationResult operationResult = new ApiExternalStorageOperationResult();
+    Set<EcatMediaFile> succeededMediaFiles = new LinkedHashSet<>();
+    try {
+      operationResult =
+          nfsManager.uploadFilesToNfs(
+              mediaFileMapById.values(), absolutePath, writableClient, attribution);
+      for (ApiExternalStorageOperationInfo info : operationResult.getFileInfoDetails()) {
+        if (!Boolean.TRUE.equals(info.getSucceeded())) {
```
rs-nicof, line 82: do you think `(Boolean.FALSE.equals(info.getSucceeded()))` would be more readable?

src/main/java/com/researchspace/service/impl/FilestoreWriteManagerImpl.java
```diff
@@ -0,0 +1,257 @@
...
+          loc.setConnectedMediaFile(mediaFile);
+          externalStorageManager.saveExternalStorageLocation(loc);
+        }
+      }
+    } catch (Exception e) {
+      log.error("Error uploading files to filestore: ", e);
+      errors.addError(new ObjectError("nfsClient", e.getMessage()));
+      throwBindExceptionIfErrors(errors);
+    }
+    return new UploadOutcome(operationResult, succeededMediaFiles);
```
rs-nicof, line 105: The list of succeded and failed files are already into `operationResult` with all the record ids.
Have you considered not to build `succeededMediaFiles` and return directly the `operationResult` (maybe adding a `getListOfSuccededRecords` as a new method)?

src/main/java/com/researchspace/api/v1/controller/GalleryIrodsApiController.java
```diff
@@ -173,49 +164,47 @@ private ApiExternalStorageOperationResult performCopyToIRODS(
...
   }
 
+  /**
+   * Resolves an absolute iRODS path from a filestore's relative path and the filesystem's
+   * configured {@code IRODS_HOME_DIR}.
+   *
+   * <p>Kept for backwards compatibility with existing tests. The unified handler uses the
+   * equivalent logic in {@link GalleryFilestoresApiController#resolveAbsoluteFilestorePath}.
+   */
   protected String getAbsoluteFilestorePathForIrods(NfsFileStore nfsFileStore) {
```
rs-nicof: If it is not used in production code and it is used in only in the test `testGettingAbsoluteTargetFilestorePath` then do you think we can remove the code and the test?

src/main/java/com/researchspace/api/v1/GalleryFilestoresApi.java
```diff
@@ -66,4 +70,56 @@ void loginToFilesystem(
...
+  /**
+   * Moves Gallery items into the filestore identified by {@code filestoreId}. After successful
+   * upload, the RSpace media file records are deleted and {@code ExternalStorageLocation} link rows
+   * are created. Unwritable backends produce 501 Not Implemented.
+   */
+  @PostMapping(
+      value = "/filestores/{filestoreId}/move",
+      produces = MediaType.APPLICATION_JSON_VALUE,
+      consumes = MediaType.APPLICATION_JSON_VALUE)
+  @ResponseStatus(HttpStatus.OK)
```
rs-nicof, line 85: Do you think `@ResponseStatus(HttpStatus.CREATED)` would be more applicable here?

src/main/java/com/researchspace/api/v1/GalleryFilestoresApi.java
```diff
@@ -66,4 +70,56 @@ void loginToFilesystem(
...
+  /**
+   * Copies Gallery items into the filestore identified by {@code filestoreId} (does not delete from
+   * RSpace). An {@code ExternalStorageLocation} link row is created per success. Unwritable
+   * backends produce 501.
+   */
+  @PostMapping(
+      value = "/filestores/{filestoreId}/copy",
+      produces = MediaType.APPLICATION_JSON_VALUE,
+      consumes = MediaType.APPLICATION_JSON_VALUE)
+  @ResponseStatus(HttpStatus.OK)
```
rs-nicof, line 102: Do you think `@ResponseStatus(HttpStatus.CREATED)` would be more applicable here?

src/main/java/com/researchspace/api/v1/GalleryFilestoresApi.java
```diff
@@ -66,4 +70,56 @@ void loginToFilesystem(
...
+   * source) to the filestore identified by {@code destFilestoreId} in the body. Currently supports
+   * S3↔S3 only; other source/destination backends produce 501. Rejects same source+dest filestore
+   * id (within-filestore moves are out of scope of this subtask). When {@code deleteSource} is
+   * true, the source object is deleted after a successful copy.
+   */
+  @PostMapping(
+      value = "/filestores/{filestoreId}/transfer",
+      produces = MediaType.APPLICATION_JSON_VALUE,
+      consumes = MediaType.APPLICATION_JSON_VALUE)
+  @ResponseStatus(HttpStatus.OK)
```
rs-nicof, line 121: Do you think `@ResponseStatus(HttpStatus.CREATED)` would be more applicable here?

src/main/java/com/researchspace/service/impl/FilestoreWriteManagerImpl.java
```diff
@@ -0,0 +1,257 @@
...
+      Set<Long> recordIds, User user, BindingResult errors) throws BindException {
+    Set<EcatMediaFile> filesRetrieved = new LinkedHashSet<>();
+    for (Long recordId : recordIds) {
+      try {
+        filesRetrieved.add(baseRecordManager.retrieveMediaFile(user, recordId));
+      } catch (ObjectRetrievalFailureException ex) {
+        errors.addError(new ObjectError("recordIds", ex.getMessage()));
+      } finally {
+        throwBindExceptionIfErrors(errors);
+      }
```
rs-nicof: This finally block needs to be put outside the the FOR LOOP otherwise it aborts the loop after the first error

rs-nicof (changes requested):
I've tested on the AWS instance from the UI enabling Irods filestore and moving files between two different S3 filestores. The functionality works as expected ✅ 

I have also added few comments on the PR to be looked at and possibly addressed ✍️

src/main/java/com/researchspace/netfiles/irods/IRODSClient.java
mKowalski256: Are we planning to add these operations to iRODS client though? I think it's fine that given operation is implemented by some filestore connectors, and not the others, we had it before with samba/sftp/irods supporting different action.

src/main/java/com/researchspace/netfiles/s3/S3NfsClient.java
mKowalski256: Ok yes I'll add getter to not chain the calls. 

As for one service that won't work for NfsClient, as `S3Utilities` are initialized for given filesystem, and there may be multiple S3 file systems configured. The transfer code needs to use the target S3Utilities, wired for target filesystem.

src/main/java/com/researchspace/service/impl/FilestoreWriteManagerImpl.java
mKowalski256: The check is already there - see line 120.

src/main/java/com/researchspace/service/impl/FilestoreWriteManagerImpl.java
mKowalski256: Ok, will do.

src/main/java/com/researchspace/service/impl/FilestoreWriteManagerImpl.java
mKowalski256: yeah, although the two are not really equivalents due to null handling - `info.succeeded` is `Boolean` object, not primitive. I'd rather keep the way it is.

src/main/java/com/researchspace/service/impl/FilestoreWriteManagerImpl.java
mKowalski256: `operationResult` only has ids of succeeded uploads as it's a lightweight DTO object. The full entities returned in `succeededMediaFiles` are immediately useful in 'move' scenario, when the code calls `deletionManager.deleteMediaFileSet()`. If these were not there, the moving code would need to immediately re-fetch the entities from db.

src/main/java/com/researchspace/api/v1/controller/GalleryIrodsApiController.java
mKowalski256: Ok, will do.

src/main/java/com/researchspace/api/v1/GalleryFilestoresApi.java
mKowalski256: `GalleryIrodsApi` used to return `HttpStatus.OK` for the same operations and I think that's more correct code, as the endpoint doesn't return created entity, just the summary result of the requested operation.

src/main/java/com/researchspace/service/impl/FilestoreWriteManagerImpl.java
mKowalski256: Good catch, yes I'll fix.

PR #785, opened 2026-05-22
RSDEV-1129-clustermaket-branding: fixed leftover clustermarket text

nhanlon2:
## Description

  Extends the Clustermarket → Calira rebrand from #763 (commit 6f5a0f0) into two
  spots that still leaked the old internal key to end-users: the system-settings
  admin page, which rendered the raw PropertyDescriptor name
  (`clustermarket.available`) as the row label, and the Clustermarket integration
  card's `website` link in the Apps page.

  Per the design decision in #763, the underlying identifier (`clustermarket.available`
  PropertyDescriptor name, `SystemPropertyName.CLUSTER_MARKET_AVAILABLE`, etc.) is
  intentionally **not** renamed — this is a presentation-only change, so no
  Liquibase migration is required and existing permission checks / REST API
  calls / Java references all continue to resolve as before.

  ## Changes

  - `bundles/system/system.properties`: new key
    `system.property.label.clustermarket.available=calira.available` for the
    display label.
  - `WEB-INF/pages/system/settings_ajax.jsp` +
    `WEB-INF/pages/system/community_settings_ajax.jsp`:
    - Mustache `systemSettingRowTemplate` now renders `{{setting.label}}` instead
      of `{{setting.name}}` inside the link, and emits an additional
      `data-label="{{setting.label}}"` attribute on the `.settingRow` element
      (the existing `data-name` attribute, used by the save handler to identify
      the property server-side, is unchanged).
    - New hidden `systemSettingsLabels` block mirrors the existing
      `systemSettingsDescriptions` block and holds the localized override for
      `clustermarket.available`.
  - `scripts/pages/system/settings_mod.js` +
    `scripts/pages/system/community_settings_mod.js`:
    - `_printSettings` looks up the override from `#systemSettingsLabels` and
      falls back to the raw property name when no override is defined, so every
      other row continues to render exactly as before.
    - `saveSetting` reads `$settingRow.data('label')` (falling back to
      `settingName`) and uses it in the success toast and failure apprise
      messages, so users no longer see `clustermarket.available` after saving.
      The POST payload and `settings[]` cache keys still use the internal name.
  - `ui/src/eln/apps/integrations/Clustermarket.tsx`: updated the integration
    card's `website` prop from `clustermarket.com` to `calira.co`.

  ## Design decisions
  
  - **Presentation-only.** No DB migration, no rename of
    `SystemPropertyName.CLUSTER_MARKET_AVAILABLE`, no change to the
    `PropertyDescriptor.name` value, no change to bundle keys' suffixes
    (`...clustermarket.available`). This keeps the change minimal and
    consistent with the boundary established in #763.
  - **Generic mechanism, single consumer.** The label-override lookup is
    implemented as a generic mechanism (mirrors the existing
    `systemSettingsDescriptions` pattern) but only `clustermarket.available`
    has an override entry today. Future rebrands can drop a new
    `<div id="<key>.label">` block in the JSP + a corresponding bundle key
    without touching the JS.
  - **`data-label` attribute** over a duplicate JS lookup: the lookup happens
    once at row render time and the result is cached on the DOM, so the save
    handler doesn't need to re-query `#systemSettingsLabels`.

  ## Testing notes

  Manual:
  1. Log in as a sysadmin and open `/system/config#systemsettings`.
  2. Under *Scientific Tools and Specialized Data*, confirm the row now reads
     `calira.available` (not `clustermarket.available`); hover the label and
     confirm the description popover still mentions Calira.
  3. Toggle the value (Allowed / Denied / Denied by default) and confirm the
     success toast reads `System Setting 'calira.available' updated to '<value>'`.
  4. Repeat under `/community/admin/.../settings` (community-admin view) to
     confirm the community settings page picks up the same label.
  5. Confirm every other integration row still renders its existing
    `systemSettingsDescriptions` pattern) but only `clustermarket.available`
    has an override entry today. Future rebrands can drop a new
    `<div id="<key>.label">` block in the JSP + a corresponding bundle key
    without touching the JS.
  - **`data-label` attribute** over a duplicate JS lookup: the lookup happens
    once at row render time and the result is cached on the DOM, so the save
    handler doesn't need to re-query `#systemSettingsLabels`.

  ## Testing notes

  Manual:
  1. Log in as a sysadmin and open `/system/config#systemsettings`.
  2. Under *Scientific Tools and Specialized Data*, confirm the row now reads
     `calira.available` (not `clustermarket.available`); hover the label and
     confirm the description popover still mentions Calira.
  3. Toggle the value (Allowed / Denied / Denied by default) and confirm the
     success toast reads `System Setting 'calira.available' updated to '<value>'`.
  4. Repeat under `/community/admin/.../settings` (community-admin view) to
     confirm the community settings page picks up the same label.
  5. Confirm every other integration row still renders its existing
     `<integration>.available` key unchanged (label fallback).

  > AI & Third Party code policy: changes in this PR were AI-assisted; please
  > review per the AI and Third Party code policy in Drata before merging.

  Note: there's also a GroupManager.java modification still uncommitted in the working tree that doesn't appear related to this rebrand — worth deciding whether to revert it or move
   it to a separate branch before pushing.

### DIFF
--- a/src/main/resources/bundles/system/system.properties
+++ b/src/main/resources/bundles/system/system.properties
@@ -348,7 +348,7 @@ system.property.description.pyrat.available=Makes PyRAT integration available. A
   users can link to animals in the PyRAT database.
 system.property.description.clustermarket.available=Makes Calira (formerly Clustermarket) integration available. After enabling the integration, \
   users can link to data regarding laboratory equipment.
-system.property.label.clustermarket.available=calira.available
+system.property.label.clustermarket.available=clustermarket.available
 system.property.description.omero.available=Makes Omero integration available. After enabling the integration, \
   users can link to Omero image data.
 system.property.description.dryad.available=Makes Dryad integration available. User can create new dryad submission\
--- a/src/main/resources/bundles/system/system.properties
+++ b/src/main/resources/bundles/system/system.properties
@@ -348,6 +348,7 @@ system.property.description.pyrat.available=Makes PyRAT integration available. A
   users can link to animals in the PyRAT database.
 system.property.description.clustermarket.available=Makes Calira integration available. After enabling the integration, \
   users can link to data regarding laboratory equipment.
+system.property.label.clustermarket.available=calira.available
 system.property.description.omero.available=Makes Omero integration available. After enabling the integration, \
   users can link to Omero image data.
 system.property.description.dryad.available=Makes Dryad integration available. User can create new dryad submission\
--- a/src/main/resources/bundles/system/system.properties
+++ b/src/main/resources/bundles/system/system.properties
@@ -346,7 +346,7 @@ system.property.description.repo.available=Makes {0} repository integration avai
   users can submit exports to a {0} repository.
 system.property.description.pyrat.available=Makes PyRAT integration available. After enabling the integration, \
   users can link to animals in the PyRAT database.
-system.property.description.clustermarket.available=Makes Calira integration available. After enabling the integration, \
+system.property.description.clustermarket.available=Makes Calira (formerly Clustermarket) integration available. After enabling the integration, \
   users can link to data regarding laboratory equipment.
 system.property.label.clustermarket.available=calira.available
 system.property.description.omero.available=Makes Omero integration available. After enabling the integration, \
--- a/src/main/resources/deployments/defaultDeployment.properties
+++ b/src/main/resources/deployments/defaultDeployment.properties
@@ -440,6 +440,8 @@ inventory.import.subSamplesLimit=2000
 ## AWS configuration properties
 aws.s3.hasS3Access=false
 
+
+#Clustermarket is now rebranded as Calira.
 #clustermarket.client.id=
 #clustermarket.secret=
 clustermarket.api.url=https://api.clustermarket.com/v1/
--- a/src/main/resources/deployments/dev/deployment.properties
+++ b/src/main/resources/deployments/dev/deployment.properties
@@ -71,7 +71,7 @@ egnyte.client.id=
 egnyte.internal.app.client.id=
 
 
-# clustermarket
+# clustermarket - now rebranded as Calira
 clustermarket.client.id=
 clustermarket.secret=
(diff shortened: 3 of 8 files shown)


nhanlon2:
@greptileai

src/main/resources/bundles/system/system.properties
```diff
@@ -348,6 +348,7 @@ system.property.description.pyrat.available=Makes PyRAT integration available. A
   users can link to animals in the PyRAT database.
 system.property.description.clustermarket.available=Makes Calira integration available. After enabling the integration, \
   users can link to data regarding laboratory equipment.
+system.property.label.clustermarket.available=calira.available
```
tchoi-rs: The settings key on the system page should agree with `deployment.properties`, therefore in sysadmin pages this should remain as `clustermarket.available`.

cc @mKowalski256

tchoi-rs (changes requested):

src/main/resources/bundles/system/system.properties
mKowalski256: hmm, I'd say the label is fine to be different than database-stored option name, as label is UI thing that should be familiar to user (sysadmin), independently of how thing is implemented. Maybe we should rename clustermarket->calira everywhere in the code, including database-stored option name, but that seems excessive. I'd accept the way it is now.

tchoi-rs: No, I mean that it should continue to say `clustermarket.available` because that is what the sysadmin is expecting it to be called. If this is changed to `calira.available` this would confuse sysadmins that expect `calira.available` to work in `deployment.properties` which doesn't work.

mKowalski256: `deployment.properties` is a separate mechanism from System->Configuration properties, the two doesn't affect each other. I.e. there was never `clustermarket.available` in deployment properties file - such property only lives in database, and is changeable by sysadmin changing it in the UI. Unless sysdamin looks into database they don't know/care how the underlying property is named. So that's why I'm arguing that we can change the label fine.

tchoi-rs (approved):

PR #786, opened 2026-05-22
RSDEV-1130-delete-user-stoichiometry-link: 1st cut

nhanlon2:
Title: RSDEV-1130: delete user with stoichiometry inventory link

  Body:

  ## Description
  Jira: [RSDEV-1130](https://researchspace.atlassian.net/browse/RSDEV-1130)

  Fixes sysadmin user deletion failing with a foreign-key error when the user
  has created a stoichiometry that has an inventory link, or owns an inventory
  item that another user's stoichiometry links to.

  Failure observed in the ticket:

  Cannot delete or update a parent row: a foreign key constraint fails
  (rspace.StoichiometryInventoryLink,
   CONSTRAINT fk_stoichiometry_molecule FOREIGN KEY (stoichiometry_molecule_id)
   REFERENCES StoichiometryMolecule (id))

  `UserDeletionDaoHibernate.deleteRecords` uses native bulk SQL to delete
  `StoichiometryMolecule`, which bypasses the ORM cascade declared on
  `StoichiometryMolecule.inventoryLink`. The DB-level FK has no
  `ON DELETE CASCADE`, so the molecule delete is rejected whenever a
  `StoichiometryInventoryLink` row exists. The same class of failure is
  reachable from `deleteInventoryItems` for the four inventory target columns of
  `StoichiometryInventoryLink` (`sample_id`, `subSample_id`, `container_id`,
  `instrumentEntity_id`).

  ## Design decisions
  - **Code-only fix, no Liquibase change.** Mirrors the file's existing pattern
    of explicit ordered raw-SQL deletes (`_AUD` first, then main) rather than
    introducing DB-level cascades. Keeps deletion order auditable in one Java file
    and consistent with the surrounding `StoichiometryMolecule_AUD` /
    `RecordAttachment_AUD` blocks.
  - `deleteRecords` cleans up link rows joined via the user's stoichiometry chain
    (covers the single-user ticket reproducer).
  - `deleteInventoryItems` cleans up link rows that target the user's inventory
    regardless of who owns the linking stoichiometry (covers the cross-user case
    reachable when a collaborator with WRITE permission links to the deleted
    user's inventory).
  - Two new ITs in `UserDeletionManagerTestIT`: one for the single-user
    reproducer and one for the cross-user inventory-target path. IT variants for
    Container and InstrumentEntity targets are intentionally omitted — the SQL
    is structurally identical to the Sample/SubSample variant; covered by review.

  ## Testing notes
  New integration tests run under `mvn verify` (or directly):

  mvn test -Dtest=UserDeletionManagerTestIT#testDeleteUserWithStoichiometryAndInventoryLink
  mvn test -Dtest=UserDeletionManagerTestIT#testDeleteUserOwningInventoryLinkedFromOtherUsersStoichiometry

  Manual repro (sanity check): create user → enable chemistry → create inventory
  sample → create doc with a stoichiometry → link an inventory subsample to a
  molecule → log out → log in as sysadmin → delete the user → confirm no FK
  error and the user is removed.

Second sanity check: do the same as above but cross-user. One user shares their inventory items with another giving write access. The other user then links their stoichiometry to those items. The inventory owning user is then deleted - it should be a success with no FK constraint errors.

  ---
  Per Drata, please follow the AI and Third Party code policy when opening this PR.

### DIFF
--- a/src/main/java/com/researchspace/dao/hibernate/UserDeletionDaoHibernate.java
+++ b/src/main/java/com/researchspace/dao/hibernate/UserDeletionDaoHibernate.java
@@ -171,6 +171,69 @@ public class UserDeletionDaoHibernate implements UserDeletionDao {
             + " :id");
     execute(userId, session, "delete from Basket where owner_id = :id");
 
+    // Delete StoichiometryInventoryLink rows targeting this user's inventory items.
+    // The link's FKs to Sample/SubSample/Container/InstrumentEntity have no
+    // ON DELETE CASCADE, so we must remove them before the inventory rows.
+    // Order: _AUD first, then main (matches the file convention).
+
+    // sample_id targets
+    execute(
+        userId,
+        session,
+        "delete sil from StoichiometryInventoryLink_AUD sil"
+            + " left join Sample s on sil.sample_id = s.id"
+            + " where s.owner_id = :id");
+    execute(
+        userId,
+        session,
+        "delete sil from StoichiometryInventoryLink sil"
+            + " left join Sample s on sil.sample_id = s.id"
+            + " where s.owner_id = :id");
+
+    // subSample_id targets (SubSample has no owner_id; resolve via parent Sample)
+    execute(
+        userId,
+        session,
+        "delete sil from StoichiometryInventoryLink_AUD sil"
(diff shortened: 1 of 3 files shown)


nhanlon2:
@greptileai

PR #788, opened 2026-05-24
Upgrade to MUI v9

tchoi-rs:
## Description ##
Upgrades RSpace to MUI v9, resolves some style overriding issues from legacy code, and some layout simplifications / bugfixes as a result of the upgrade.

(Change the destination branch after the Vite HMR branch is updated)


tchoi-rs:
@greptileai

tchoi-rs:
@greptileai

src/main/webapp/styles/simplicity/typo.css
```diff
@@ -93,16 +93,9 @@ body td.field-value-inner span[style*="line-height: 36pt;"], body .journalPagePa
...
-
-img {
-  border: 0; }
-
 /* LINKS */
-button, button:link, button:active, a, a:link, a:active {
-  color: #1465b7;
-  text-decoration: none !important; }
+a, a:link, a:active {
+  text-decoration: none; }
```
mKowalski256, line 98: removal of `color` here is I believe behind the change of blue shade in the Workspace.

mKowalski256 (commented):
I did some testing on [test AWS build](https://feat-mui-v9-4f812fb4-54.researchspace.com/import/archiveImport) and didn't find any broken functionalities, but the branch made various smaller changes to styling, which seem possibly accidental and not necessarily wanted.

----
On Workspace page, there are a few unexpected changes to the main toolbar:
1. `Create` button is much smaller now - I believe we want the old size
2. The buttons displayed on the toolbar used to be grouped visually, with small horizontal empty spaces right of 'calendar' and 'labgroup records' icon. Now the divider spaces seem to be gone.
3. The colour of link / buttons seems to be a different shade of blue (lighter), and also doesn't seem to be universally applied (see color 'email' of the user details pill). I've noted in the code comment that this is due to changed `typo.css` styling.

Screenshot showing issues 1-3:

Image

----

The top navbar styling issue:
4. 'Account Menu' popup doesn't apply spacing on the left of username/email, so there is no visible space between icon and detail:

Image

----

On Gallery page:
5. After opening the page, even though no item is selected, the 'Actions' menu is not greyed out, as it used to be. Clicking on it does nothing.

Image

----

On Inventory page:
6. In listing, the 'name' column is much wider and/or no longer applies `word-break: break-all` styling, which I think is not what we want as in narrow 3-pane-panel you get a horizontal scrollbar (for longer sample names you'll get them even on wide-screen)
7. I've noted that on sample/subsample view, 'Details' -> 'Total Quantity' (or 'Quantity') now is coloured as light-grey, similar to 'no value' colouring, while previously it was black.
8. I've noted that on subsample view, when you open section of 'Sample Fields', the styling of h5 headers is much different than elsewhere (and inconsistent with prior behaviour), e.g. these headers don't apply `text-transform: uppercase;` rule, and also use larger font-size.

Image

---- 

On Document Editor page:
9. 'Insert from Gallery' Dialog: shows subgallery name (e.g. 'Images') in blue color, rather than black
10. 'Insert from Gallery' Dialog:  The new styling around 'description' input is cut on the right in the dialog 

Image

mKowalski256:
That's most of raised issues now gone in[ latest build](https://feat-mui-v9-4f812fb4-61.researchspace.com/), but 2. 3. and 7. are still there:

2. there is no spacing separation between button groups on workspace toolbar
3. the workspace still uses a mix of blue shades: record table uses `#2196f3`, user details popup uses `#00adefff` and footer link uses `color: -webkit-link`. previously the unified color was set in `typo.css` file through `color: #1465b7; ` in `a, a:link, a:active` rule that is deleted in this PR, I think that could be just restored

image

7. inventory quantity value still shows as 'disabled' - as discussed before.

mKowalski256 (approved):

PR #803, opened 2026-05-29
RSDEV-1131: Related Items link fields for Inventory

nhanlon2:
## Description

Implements RSDEV-1131: (**which contains a full test plan**): structured **"Related Items" Link fields** for Inventory, so inventory items and sample templates can hold typed, versioned links to other records. 45 commits, ~150 files (roughly two thirds frontend, one third backend plus a Liquibase migration).

**Note - NO linking to version history of Gallery Items as yet - this requires developing a UI to display version history of Gallery Items first.**

Note - the changes to permissions notifications on unshare were due to a caching issue where user was not notified of unshare events.

### What users get

- **Link fields on inventory items**: a new "Link" extra-field type alongside Text and Number. Each link has a relationship type (full DataCite 4.7 vocabulary plus the PIDINST `IsCalibratedBy`/`Calibrates` pair, 40 values), a target Global ID, and an optional version pin.
- **Supported targets**: Inventory items (samples, subsamples, containers, instruments, sample templates) and ELN records (documents, notebooks, gallery files). Self-links are rejected.
- **Link editor** with three ways to set a target: Browse Inventory picker, Browse ELN tree picker (folders expand, notebooks expand and select, documents and gallery files are pickable leaves), or typing a Global ID. Typed ids are validated server-side against the exact Global ID on Apply.
- **Version pinning**: a link can pin a specific user-facing version of its target (samples, subsamples, containers, templates, ELN documents). Pins are edited inside the link editor via a version-history dialog; pinned info dialogs and Open honour the pinned version. Notebooks and gallery files are not pinnable.
- **Link card** rendering one inline row: relation chip, target chip with type icon, info button, version pill, Open, and Edit when editable.
- **Info dialogs** for both target kinds: inventory targets show the MoreInfo record body (historical snapshot when pinned, with a "may not be the latest version" notice); ELN targets show a document/gallery body with preview, Download, Upload new version, and back-references.
- **Target-state pills** resolved lazily per card from a new summary endpoint: **"Target deleted"** for soft-deleted targets and **"No access"** for ELN targets the viewer cannot read. Open is removed for deleted or unreadable ELN targets (their routes only produce error pages) but kept for deleted Inventory targets (the trash viewer works) and for unreadable Inventory targets (every logged-in user keeps the limited-read view).
- **Back-references both ways**: "Inventory items linking to this item" on inventory records (name, Global ID, relation, linked version) and "Related inventory items" on ELN documents, notebooks, and gallery files (in the React info panels and the legacy record-info panel). Both lists are permission-filtered server-side.
- **Templates**: sample templates can define Link fields with an optional whitelist of allowed relationship types; samples created from the template are constrained to it, and "Update all samples" propagates new link fields (including mandatory ones) without erroring on empty values.
- **Editor/Save lifecycle**: record-level Save is blocked with an explanatory message while any field editor is open or holds unapplied link changes, so in-progress edits are never silently discarded.

### API surface (`/api/inventory/v1`)

- `GET /linkTargets/{globalId}/summary` (new): current state of a link target {globalId, name, type, deleted, readable}, permission-redacted per viewer.
- `GET /{samples|subSamples|containers|sampleTemplates}/{id}/referencingItems` and a generic Global ID variant for ELN targets (new): permission-filtered back-references. Requests for targets the caller cannot read are rejected with the same "not found" error a nonexistent id produces.
- `ApiExtraField` gains the `link` payload {relationType, targetGlobalId, versionPin}; sample/template field DTOs gain the relation-type whitelist.

### Backend

- `InventoryLink` entity rows (in `rspace-core-model`) with Hibernate Envers auditing; `InventoryLinkDao`/`InventoryLinkManager`/`InventoryLinkValidator` follow the existing controller, manager, DAO layering.
- `LinkTargetResolver` (live existence and read-permission checks, exact prefix match) and `LinkTargetSnapshotResolver` (Envers-based version and snapshot resolution for pinned links and the summary endpoint).
- Liquibase migration `changeLog-rsdev1131-step1.xml`: `InventoryLink` and `InventoryEntityField` tables plus `ExtraField` columns, each with `_AUD` counterparts.
- User-facing validation messages are externalized to `bundles/inventory/inventory.properties`.

### Frontend

- Inventory side under `src/Inventory/components/Fields/Link/` (card, editors, pickers, info dialogs, version dialog, target-summary hook) and `ExtraFields/`; template fields under `src/Inventory/Sample|Template/Fields/`.
- ELN side: `useReferencingInventoryItems` hook and `ReferencingInventoryItemsPanel` in the Gallery info panel, plus the legacy `recordInfoPanel` table.

## Design decisions

- **Lazy target summaries** (RSDEV-1182): the summary is fetched per rendered card, not serialized eagerly on every LIST response, to keep list endpoints cheap. Decision recorded in the ticket.
- **Non-disclosure invariant** (`DevDocs/adr/0002-link-target-state-non-disclosure.md`): unreadable targets are redacted to a globalId-only summary that is field-for-field identical to a nonexistent target's, so the endpoint cannot be used to probe which records exist. This is why the pill says "No access" rather than "Item unshared", and why the two pills can never co-occur. A unit test pins the invariant. Ubiquitous language is in `DevDocs/CONTEXT.md`.
- **Exact Global ID validation**: both validation layers require the stored prefix to match the requested one, so e.g. `GL150` is rejected when only folder `FL150` shares the number; existence is never confirmed for unreadable records.
- **Pin semantics**: a pin belongs to a specific target, so retargeting resets the staged pin to Latest; pins only take effect on Update/Apply; unpinned links follow the latest version. Saving a record without touching its link never rewrites the link (no spurious audit revisions).
- **`rspace-core-model` is temporarily pinned to this feature's branch build** (`RSDEV-1131-Inventory-link-related-items-4a3efa575b-1`). It will be repointed to a released version before merge; the two open review threads track this deliberately.

## Testing notes

- Backend: pure unit tests for the validators, managers, resolvers, and DTOs (`mvn test -Dfast=true`); author-run `*IT` coverage for the Envers snapshot paths; MVC ITs for the referencing-items endpoints and template link fields.
- Frontend: Vitest suites for the link card, both editors, both pickers, the version dialog, info dialogs, hooks, and the workspace/gallery panels.
- A 17-section manual test plan (pickers, validation, pinning, pills, back-references, permissions, persistence/audit, editor/save lifecycle, regression sweep) lives in the RSDEV-1131 description.
- **Deferred**: Instruments are fully supported by the backend and plumbing but cannot be exercised end-to-end until an Instruments UI exists; there is no versioned viewer for sample templates (a pinned template link opens the live template while its info dialog shows the pinned snapshot).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -17,6 +17,7 @@ These instructions apply to the entire repository unless a deeper `AGENTS.md` ov
 - When searching the repo, prefer `rg` and `rg --files`.
 - Do not edit minified files unless the user explicitly asks for it.
 - Plan documents (design notes, multi-step implementation plans, scratch analyses) must be written to the `.claude/` folder, which is gitignored. Never place plan files at the repository root or anywhere else inside `src/`. The `.claude/` folder is the only sanctioned location for ephemeral working notes that should not be committed.
+- **Never run the Maven `install` phase or install dependencies into the local Maven repository.** Do not run `mvn install`, `./mvnw install`, `mvn install:install-file`, `mvn deploy:deploy-file`, or any command that would write artefacts into `~/.m2/repository`. This applies to this project, to sibling projects in the workspace (notably `rspace-core-model`), and to any dependency the user is iterating on. Reason: locally installed artefacts silently shadow the jitpack-built artefact at the same Maven coordinate, so the running JVM ends up with bytecode that does not match what the pom claims to use. That mismatch produces failures that look like a code bug but are really a classpath bug, and it wastes hours to track down. Stick to phases that do not touch `~/.m2`: `mvn compile`, `mvn package`, `mvn test`, `mvn verify`. Dependencies are consumed via jitpack at the version pinned in `pom.xml` — if the user wants to test an unpublished change to a sibling project, ask them to push it and trigger a remote build, then bump the pom to the new commit-hash version. Do not work around this by installing locally.
 
 ## Project Overview
 
@@ -73,6 +74,13 @@ The `rspacedbuser` credentials must match `src/main/resources/rs.properties`.
 # Subsequent runs (keep existing DB)
 ./mvnw jetty:run -Denvironment=keepdbintact -Dspring.profiles.active=run -DreactDevMode=true
 ```
+
+> **Dev cache-busting for legacy assets:** in dev mode (`reactDevMode=true`), legacy
+> `/scripts/` and `/styles/` assets (e.g. `recordInfoPanel.js`) are served without a `?v=`
+> cache-buster, so the browser keeps serving a previously cached copy after you edit them.
+> If a legacy JS/CSS change does not show up, hard-refresh (Cmd+Shift+R), or add
+> `-DlegacyAssetCacheBustingInDevMode=true` to the `jetty:run` command so those assets are
+> cache-busted automatically. The React/Vite bundle is unaffected (it is always served fresh).
 
 Access at `http://localhost:8080`. Test users: `user1a`–`user8h`, admin: `sysadmin1`.
 
@@ -189,6 +197,14 @@ pnpm run test -- src/components/MyComponent/__tests__/MyComponent.test.tsx
 - **Test setup:** Global setup in `src/__tests__/setup.ts` polyfills `localStorage`, `sessionStorage`, `TextEncoder`/`TextDecoder`.
 - **Console suppression:** Use `silenceConsole()` from test helpers to suppress expected errors.
 - **Test timeout:** 20 seconds (configured in `vitest.config.ts`).
+- **Vitest must execute with `src/main/webapp/ui` as its working directory.** The `vite.config.ts` that owns module aliases (`@/` -> `src/`, `Styles` -> `src/util/styles.ts`) lives there. Running `vitest` from any other directory produces module-resolution failures such as `Cannot find package '@/__tests__/customQueries'` or `Cannot find package 'Styles'`, which look like a source bug but are a cwd problem. Since the pnpm migration the package.json lives at the repo root, so the sanctioned entry is `pnpm test <path-relative-to-ui>` from the repo root (the script cd's into `ui` itself). `npx vitest run <path>` from inside `ui` also works. Beware: `pnpm exec vitest` from inside `ui` can fail with `ERR_PNPM_RECURSIVE_EXEC_NO_PACKAGE` because `ui` no longer has its own package.json.
+- **Lint dialect required by this repo's ESLint config — write it compliant from the first draft instead of fixing up afterwards:**
+  - Use `expect(node).toBeInTheDocument()` / `not.toBeInTheDocument()`, never `.toBeTruthy()` / `toBeNull()` for DOM existence.
+  - Use `expect(node).toHaveAttribute(name, value)`, never `expect(node.getAttribute(name)).toBe(value)`.
+  - Use `expect(button).toBeDisabled()`, never `expect((button as HTMLButtonElement).disabled).toBe(true)`.
+  - For `vi.mocked(obj.method)`, extract the method to a local first (`const m = obj.method; vi.mocked(m)`), otherwise `@typescript-eslint/unbound-method` fires.
+  - `testing-library/no-container` and `no-node-access` are enabled. If you genuinely need `container.querySelector` (rare — usually to inspect a non-semantic child of a MUI component), wrap the block in `/* eslint-disable testing-library/no-node-access, testing-library/no-container */` with a comment explaining why.
+- **MUI `Select` inside `FormField` has a broken accessible name.** Its `aria-labelledby` is set to its own id rather than the label's, so `getByRole('combobox', { name: /field type/i })` returns no matches and `getByLabelText('Field type')` resolves to the FormControl wrapper rather than the select itself, so neither query reliably reaches the control. The reliable query is a stable `data-testid` on the display element: pass `SelectDisplayProps={{ "data-testid": "MySelect" } as React.HTMLAttributes<HTMLDivElement>}` and then `screen.getByTestId('MySelect')`.
 
 ### Internationalization (i18n)
 
--- a/src/main/java/com/researchspace/api/v1/controller/InventoryReferencingItemsApiController.java
+++ b/src/main/java/com/researchspace/api/v1/controller/InventoryReferencingItemsApiController.java
@@ -1,0 +1,56 @@
+package com.researchspace.api.v1.controller;
+
+import com.researchspace.api.v1.InventoryReferencingItemsApi;
+import com.researchspace.api.v1.model.ApiInventoryReferencingItems;
+import com.researchspace.model.User;
+import com.researchspace.model.core.GlobalIdPrefix;
+import com.researchspace.model.core.GlobalIdentifier;
+import com.researchspace.service.inventory.InventoryLinkManager;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestAttribute;
+
+@ApiController
+public class InventoryReferencingItemsApiController implements InventoryReferencingItemsApi {
+
+  @Autowired private InventoryLinkManager inventoryLinkManager;
+
+  @Override
+  public ApiInventoryReferencingItems getReferencingItemsForSample(
+      @PathVariable Long id, @RequestAttribute(name = "user") User user) {
+    return forItem(GlobalIdPrefix.SA, id, user);
+  }
+
+  @Override
+  public ApiInventoryReferencingItems getReferencingItemsForSubSample(
+      @PathVariable Long id, @RequestAttribute(name = "user") User user) {
+    return forItem(GlobalIdPrefix.SS, id, user);
+  }
+
+  @Override
+  public ApiInventoryReferencingItems getReferencingItemsForContainer(
+      @PathVariable Long id, @RequestAttribute(name = "user") User user) {
+    return forItem(GlobalIdPrefix.IC, id, user);
+  }
+
+  @Override
+  public ApiInventoryReferencingItems getReferencingItemsForInstrument(
+      @PathVariable Long id, @RequestAttribute(name = "user") User user) {
+    return forItem(GlobalIdPrefix.IN, id, user);
+  }
+
+  @Override
+  public ApiInventoryReferencingItems getReferencingItemsForGlobalId(
+      @PathVariable String globalId, @RequestAttribute(name = "user") User user) {
+    ApiInventoryReferencingItems result = new ApiInventoryReferencingItems();
+    result.setReferencingItems(inventoryLinkManager.findReferencingItems(globalId, user));
+    return result;
+  }
+
+  private ApiInventoryReferencingItems forItem(GlobalIdPrefix prefix, Long id, User user) {
+    String globalId = new GlobalIdentifier(prefix, id).getIdString();
+    ApiInventoryReferencingItems result = new ApiInventoryReferencingItems();
+    result.setReferencingItems(inventoryLinkManager.findReferencingItems(globalId, user));
+    return result;
+  }
+}
--- a/src/main/java/com/researchspace/dao/InventoryLinkDao.java
+++ b/src/main/java/com/researchspace/dao/InventoryLinkDao.java
@@ -1,0 +1,28 @@
+package com.researchspace.dao;
+
+import com.researchspace.model.core.GlobalIdPrefix;
+import com.researchspace.model.inventory.field.ExtraLinkField;
+import com.researchspace.model.inventory.field.InventoryLink;
+import com.researchspace.model.inventory.field.InventoryLinkField;
+import java.util.List;
+
+/** DAO for the InventoryLink table backing Inventory Link extra-fields. */
+public interface InventoryLinkDao extends GenericDao<InventoryLink, Long> {
+
+  /**
+   * Returns the ExtraLinkField rows whose link points at the supplied target record, matched by
+   * (prefix, database id) so that version-pinned links (e.g. {@code SD123v5}) still surface on the
+   * base record ({@code SD123}). Filters out soft-deleted fields and soft-deleted links. Used to
+   * drive the "referencing items" back-ref lookup by joining each field back to its parent
+   * inventory record.
+   */
+  List<ExtraLinkField> findReferencingLinkFields(GlobalIdPrefix targetPrefix, Long targetDbId);
+
+  /**
+   * The template-defined structured link fields (on samples) whose non-deleted link targets the
+   * given record. Counterpart of {@link #findReferencingLinkFields} for {@code
+   * InventoryLinkField}s, so structured links appear in back-references like extra-field links.
+   */
+  List<InventoryLinkField> findReferencingStructuredLinkFields(
+      GlobalIdPrefix targetPrefix, Long targetDbId);
+}
--- a/src/main/java/com/researchspace/dao/hibernate/UserDeletionDaoHibernate.java
+++ b/src/main/java/com/researchspace/dao/hibernate/UserDeletionDaoHibernate.java
@@ -424,6 +424,29 @@ public class UserDeletionDaoHibernate implements UserDeletionDao {
 
     execute(userId, session, "delete from Container_AUD where owner_id = :id");
     execute(userId, session, "delete from Container where owner_id = :id");
+
+    // Hard-delete InventoryLink rows orphaned by the field deletions above. The FKs point
+    // field -> link, so once the user's ExtraField / InventoryEntityField rows (and their _AUD
+    // copies) are gone the rows are unreachable garbage that would otherwise leak forever
+    // (mirrors the StoichiometryInventoryLink cleanup). The sweep only matches rows with no
+    // remaining referrer in any live or audited field table, so links belonging to other
+    // users' surviving fields are untouched. _AUD first, then main.
+    executeUnparameterised(
+        session,
+        "delete il from InventoryLink_AUD il"
+            + " left join ExtraField ef on ef.link_id = il.id"
+            + " left join ExtraField_AUD efa on efa.link_id = il.id"
+            + " left join InventoryEntityField ief on ief.link_id = il.id"
+            + " left join InventoryEntityField_AUD iefa on iefa.link_id = il.id"
+            + " where ef.id is null and efa.id is null and ief.id is null and iefa.id is null");
+    executeUnparameterised(
+        session,
+        "delete il from InventoryLink il"
+            + " left join ExtraField ef on ef.link_id = il.id"
+            + " left join ExtraField_AUD efa on efa.link_id = il.id"
+            + " left join InventoryEntityField ief on ief.link_id = il.id"
+            + " left join InventoryEntityField_AUD iefa on iefa.link_id = il.id"
+            + " where ef.id is null and efa.id is null and ief.id is null and iefa.id is null");
   }
 
   private void deleteAppConfigs(Long userId, Session session) {
@@ -702,6 +725,15 @@ public class UserDeletionDaoHibernate implements UserDeletionDao {
     Query<?> query = session.createSQLQuery(idQuery);
     query.setParameter("id", userId);
     query.executeUpdate();
+    session.flush();
+  }
+
+  /**
+   * Like {@link #execute} but for statements with no bind parameters. The cross-user orphan sweeps
+   * have no {@code :id} placeholder, and binding one anyway makes Hibernate throw at runtime.
+   */
+  private void executeUnparameterised(Session session, String sql) {
+    session.createSQLQuery(sql).executeUpdate();
     session.flush();
   }
 
--- a/src/main/java/com/researchspace/service/inventory/InventoryLinkValidator.java
+++ b/src/main/java/com/researchspace/service/inventory/InventoryLinkValidator.java
@@ -1,0 +1,115 @@
+package com.researchspace.service.inventory;
+
+import com.researchspace.api.v1.model.ApiInventoryLink;
+import com.researchspace.model.core.GlobalIdPrefix;
+import com.researchspace.model.core.GlobalIdentifier;
+import java.util.EnumSet;
+import java.util.Set;
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.validation.Errors;
+
+/**
+ * Validates an incoming {@link ApiInventoryLink} payload independent of database state. Checks
+ * relationType against the DataCite vocabulary, syntactic GlobalID parse, that the target prefix is
+ * a supported link target (Inventory items including sample templates, or ELN documents, notebooks
+ * and gallery files), and forbids self-links. Existence and read-permission checks live in the
+ * manager layer, where the acting user is available.
+ */
+public class InventoryLinkValidator {
+
+  /**
+   * True when the prefix is a kind that links may target (shared with the manager's write-path
+   * validation).
+   */
+  public static boolean isAllowedTargetPrefix(GlobalIdPrefix prefix) {
+    return prefix != null && ALLOWED_TARGET_PREFIXES.contains(prefix);
+  }
+
+  /** True when target points at the same record as sourceGlobalId (version suffixes ignored). */
+  public static boolean isSelfLink(GlobalIdentifier target, String sourceGlobalId) {
+    if (StringUtils.isBlank(sourceGlobalId) || target == null) {
+      return false;
+    }
+    try {
+      GlobalIdentifier source = new GlobalIdentifier(sourceGlobalId);
+      return source.getPrefix() == target.getPrefix() && source.getDbId().equals(target.getDbId());
+    } catch (IllegalArgumentException ex) {
+      return false;
+    }
+  }
+
+  private static final Set<GlobalIdPrefix> ALLOWED_TARGET_PREFIXES =
+      EnumSet.of(
+          GlobalIdPrefix.SA,
+          GlobalIdPrefix.SS,
+          GlobalIdPrefix.IC,
+          GlobalIdPrefix.IN,
+          GlobalIdPrefix.IT,
+          GlobalIdPrefix.SD,
+          GlobalIdPrefix.NB,
+          GlobalIdPrefix.GL);
+
+  /**
+   * Validates the supplied link. Errors are rejected as {@code relationType} / {@code
+   * targetGlobalId} field codes relative to whatever nested path the passed {@link Errors} object
+   * is currently positioned at. This validator does not push a {@code link} path itself: callers
+   * that want the errors to surface under {@code link.*} must {@code pushNestedPath("link")} on the
+   * Errors object before calling (as {@code ApiExtraFieldsHelper} does). {@code sourceGlobalId}
+   * (the GlobalID of the parent item the link is being attached to) is used to detect self-links.
+   */
+  public void validate(ApiInventoryLink link, String sourceGlobalId, Errors errors) {
+    if (link == null) {
+      return;
+    }
+    validateRelationType(link, errors);
+    validateTarget(link, sourceGlobalId, errors);
+  }
+
+  private void validateRelationType(ApiInventoryLink link, Errors errors) {
+    if (!DataCiteRelationType.isValid(link.getRelationType())) {
+      errors.rejectValue(
+          "relationType",
+          "errors.inventory.field.link.relationTypeInvalid",
+          new Object[] {link.getRelationType()},
+          "relationType not in DataCite vocabulary");
+    }
+  }
+
+  private void validateTarget(ApiInventoryLink link, String sourceGlobalId, Errors errors) {
+    String target = link.getTargetGlobalId();
+    if (StringUtils.isBlank(target)) {
+      errors.rejectValue(
+          "targetGlobalId",
+          "errors.inventory.field.link.targetNotFound",
+          new Object[] {target},
+          "targetGlobalId is required");
+      return;
+    }
+    GlobalIdentifier parsed;
+    try {
+      parsed = new GlobalIdentifier(target);
+    } catch (IllegalArgumentException ex) {
+      errors.rejectValue(
+          "targetGlobalId",
+          "errors.inventory.field.link.targetNotFound",
+          new Object[] {target},
+          "targetGlobalId is not a valid GlobalID");
+      return;
+    }
+    if (!ALLOWED_TARGET_PREFIXES.contains(parsed.getPrefix())) {
+      errors.rejectValue(
+          "targetGlobalId",
+          "errors.inventory.field.link.targetKindUnsupported",
+          new Object[] {parsed.getPrefix().name()},
+          "target kind not supported");
+      return;
+    }
+    if (isSelfLink(parsed, sourceGlobalId)) {
+      errors.rejectValue(
+          "targetGlobalId",
+          "errors.inventory.field.link.selfLinkForbidden",
+          new Object[] {target},
+          "link target cannot equal its source item");
+    }
+  }
+}
--- a/src/main/java/com/researchspace/service/inventory/impl/InventoryLinkManagerImpl.java
+++ b/src/main/java/com/researchspace/service/inventory/impl/InventoryLinkManagerImpl.java
@@ -1,0 +1,182 @@
+package com.researchspace.service.inventory.impl;
+
+import com.researchspace.api.v1.auth.ApiRuntimeException;
+import com.researchspace.api.v1.model.ApiInventoryLink;
+import com.researchspace.api.v1.model.ApiInventoryLinkTargetSummary;
+import com.researchspace.api.v1.model.ApiInventoryReferencingItem;
+import com.researchspace.dao.InventoryLinkDao;
+import com.researchspace.model.User;
+import com.researchspace.model.core.GlobalIdentifier;
+import com.researchspace.model.inventory.InventoryRecord;
+import com.researchspace.model.inventory.field.ExtraLinkField;
+import com.researchspace.model.inventory.field.InventoryLink;
+import com.researchspace.model.inventory.field.InventoryLinkField;
+import com.researchspace.service.inventory.DataCiteRelationType;
+import com.researchspace.service.inventory.InventoryLinkManager;
+import com.researchspace.service.inventory.InventoryLinkValidator;
+import com.researchspace.service.inventory.InventoryPermissionUtils;
+import com.researchspace.service.inventory.LinkTargetResolver;
+import com.researchspace.service.inventory.LinkTargetSnapshotResolver;
+import java.util.ArrayList;
+import java.util.List;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+@Service
+public class InventoryLinkManagerImpl implements InventoryLinkManager {
+
+  @Autowired private InventoryLinkDao linkDao;
+  @Autowired private InventoryPermissionUtils permissionUtils;
+  @Autowired private LinkTargetResolver linkTargetResolver;
+  @Autowired private LinkTargetSnapshotResolver snapshotResolver;
+
+  @Override
+  public InventoryLink createLink(ApiInventoryLink apiLink, User actor) {
+    validateForWrite(apiLink);
+    assertTargetExistsAndReadable(apiLink, actor);
+    InventoryLink entity = new InventoryLink();
+    applyApiToEntity(apiLink, entity);
+    return linkDao.save(entity);
+  }
+
+  @Override
+  public InventoryLink updateLink(InventoryLink existing, ApiInventoryLink apiLink, User actor) {
+    validateForWrite(apiLink);
+    assertTargetExistsAndReadable(apiLink, actor);
+    applyApiToEntity(apiLink, existing);
+    return linkDao.save(existing);
+  }
+
+  @Override
+  public void deleteLink(InventoryLink existing, User actor) {
+    existing.setDeleted(true);
+    linkDao.save(existing);
+  }
+
+  @Override
+  public List<ApiInventoryReferencingItem> findReferencingItems(String targetGlobalId, User actor) {
+    GlobalIdentifier target;
+    try {
+      target = new GlobalIdentifier(targetGlobalId);
+    } catch (IllegalArgumentException ex) {
+      // a malformed Global ID does not resolve to any target; surface it as a clean API error
+      // rather than letting the raw IllegalArgumentException escape to the HTTP layer
+      throw new ApiRuntimeException("errors.inventory.field.link.targetNotFound", targetGlobalId);
+    }
+    // the caller must be able to READ the target itself: filtering the sources alone would
+    // still confirm an unreadable record's existence and reveal its inbound links to anyone
+    // who guesses its id. Same error as the malformed/missing case, so nothing is disclosed.
+    if (!linkTargetResolver.targetExistsAndIsReadable(target, actor)) {
+      throw new ApiRuntimeException("errors.inventory.field.link.targetNotFound", targetGlobalId);
+    }
+    List<ApiInventoryReferencingItem> rows = new ArrayList<>();
+    for (ExtraLinkField field :
+        linkDao.findReferencingLinkFields(target.getPrefix(), target.getDbId())) {
+      addRowIfReadable(rows, field.getInventoryRecord(), field.getLink(), actor);
+    }
+    // template-defined structured link fields are first-class links too
+    for (InventoryLinkField field :
+        linkDao.findReferencingStructuredLinkFields(target.getPrefix(), target.getDbId())) {
+      addRowIfReadable(rows, field.getInventoryRecord(), field.getLink(), actor);
+    }
+    return rows;
+  }
+
+  private void addRowIfReadable(
+      List<ApiInventoryReferencingItem> rows,
+      InventoryRecord parent,
+      InventoryLink link,
+      User actor) {
+    if (parent == null || link == null || parent.isDeleted()) {
+      return;
+    }
+    if (!permissionUtils.canUserReadInventoryRecord(parent, actor)) {
+      return;
+    }
+    ApiInventoryReferencingItem row = new ApiInventoryReferencingItem();
+    row.setSourceGlobalId(parent.getOid().toString());
+    row.setSourceName(parent.getName());
+    row.setSourceType(parent.getType().toString());
+    row.setRelationType(link.getRelationType());
+    row.setVersionPin(link.getVersionPin());
+    if (link.getModifiedAt() != null) {
+      row.setModifiedAtMillis(link.getModifiedAt().getTime());
+    }
+    rows.add(row);
+  }
+
+  @Override
+  public ApiInventoryLinkTargetSummary getTargetSummary(String targetGlobalId, User actor) {
+    GlobalIdentifier gid = parseAllowedTargetOrThrow(targetGlobalId);
+    // current state of the base record: pin/revision deliberately null so the
+    // "Target deleted" pill reflects the record as it is now, not as pinned
+    return snapshotResolver.resolveSummary(gid.getPrefix(), gid.getDbId(), null, null, actor);
+  }
+
+  /**
+   * Rejects links whose target does not resolve to a real record the actor can READ. Applies to all
+   * targets, Inventory and ELN alike. The version suffix (if any) is ignored: only the base record
+   * needs to exist and be readable.
+   */
+  /**
+   * Structural validation of an incoming link payload, mirroring {@link InventoryLinkValidator}.
+   * The controller-layer validator can be bypassed (an extra-field update payload that omits {@code
+   * type} skips the LINK validation branch, and structured sample fields never run it), so the
+   * manager enforces the same rules at the single write path: a parseable target id, an allowed
+   * target kind, and a DataCite relation type. Failures map to 422 with the same bundle keys the
+   * validator uses.
+   */
+  private void validateForWrite(ApiInventoryLink apiLink) {
+    parseAllowedTargetOrThrow(apiLink.getTargetGlobalId());
+    if (!DataCiteRelationType.isValid(apiLink.getRelationType())) {
+      throw new ApiRuntimeException(
+          "errors.inventory.field.link.relationTypeInvalid", apiLink.getRelationType());
+    }
+  }
+
+  /**
+   * Parses a target Global ID and asserts its prefix is an allowed link target kind, mapping
+   * failures to the same 422 i18n errors on both the write path and the summary read path.
+   */
+  private GlobalIdentifier parseAllowedTargetOrThrow(String targetGlobalId) {
+    GlobalIdentifier gid;
+    try {
+      gid = new GlobalIdentifier(targetGlobalId);
+    } catch (IllegalArgumentException | NullPointerException ex) {
+      throw new ApiRuntimeException("errors.inventory.field.link.targetNotFound", targetGlobalId);
+    }
+    if (!InventoryLinkValidator.isAllowedTargetPrefix(gid.getPrefix())) {
+      throw new ApiRuntimeException(
+          "errors.inventory.field.link.targetKindUnsupported", gid.getPrefix().name());
+    }
+    return gid;
+  }
+
+  private void assertTargetExistsAndReadable(ApiInventoryLink apiLink, User actor) {
+    GlobalIdentifier gid = new GlobalIdentifier(apiLink.getTargetGlobalId());
+    if (!linkTargetResolver.targetExistsAndIsReadable(gid, actor)) {
+      throw new ApiRuntimeException(
+          "errors.inventory.field.link.targetNotFound", apiLink.getTargetGlobalId());
+    }
+  }
+
+  private void applyApiToEntity(ApiInventoryLink api, InventoryLink entity) {
+    entity.setRelationType(api.getRelationType());
+    GlobalIdentifier gid = new GlobalIdentifier(api.getTargetGlobalId());
+    // persist the unsuffixed base id: the version lives in versionPin, so a
+    // "vN" suffix on the incoming id must not be doubly encoded in the row
+    entity.setTargetGlobalId(new GlobalIdentifier(gid.getPrefix(), gid.getDbId()).getIdString());
+    entity.setTargetPrefix(gid.getPrefix());
+    entity.setTargetDbId(gid.getDbId());
+    Long versionPin = gid.hasVersionId() ? gid.getVersionId() : api.getVersionPin();
+    entity.setVersionPin(versionPin);
+    // Capture the exact audit revision the pinned version maps to. Null version means "latest",
+    // which the snapshot resolver resolves dynamically at read time, so no revision is stored.
+    Long revisionId =
+        versionPin == null
+            ? null
+            : snapshotResolver.resolveRevisionForVersion(
+                gid.getPrefix(), gid.getDbId(), versionPin);
+    entity.setTargetRevisionId(revisionId);
+  }
+}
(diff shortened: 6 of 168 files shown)


nhanlon2:
@greptileai

nhanlon2:
@greptileai

nhanlon2:
@greptileai

src/main/java/com/researchspace/service/inventory/impl/InventoryLinkManagerImpl.java
```diff
@@ -0,0 +1,81 @@
...
+    entity.setRelationType(api.getRelationType());
+    entity.setTargetGlobalId(api.getTargetGlobalId());
+    GlobalIdentifier gid = new GlobalIdentifier(api.getTargetGlobalId());
+    entity.setTargetPrefix(gid.getPrefix());
+    entity.setTargetDbId(gid.getDbId());
+    if (gid.hasVersionId()) {
+      entity.setVersionPin(gid.getVersionId());
+    } else {
+      entity.setVersionPin(api.getVersionPin());
+    }
```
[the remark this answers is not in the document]
nhanlon2: Fixed in 22f82ef6c: applyApiToEntity now persists the unsuffixed base Global ID and derives the pin into versionPin (the DAO lookups already query by parsed prefix+dbId, so back-references match either way).

src/main/webapp/ui/src/Inventory/components/Fields/Link/InventoryInfoDialog.tsx
```diff
@@ -0,0 +1,104 @@
...
+  onClose: () => void;
+}
+
+const GLOBAL_ID_PATTERN = /^([A-Z]{2})(\d+)(?:v\d+)?$/;
+const PREFIX_TO_PATH: Record<string, string> = {
+  SA: "samples",
+  SS: "subSamples",
+  IC: "containers",
+  IT: "sampleTemplates",
+};
```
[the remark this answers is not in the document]
nhanlon2: Fixed in an earlier commit: InventoryInfoDialog now resolves endpoints via the shared INVENTORY_PREFIX_TO_API_PATH map, which includes IN -> instruments.

src/main/java/com/researchspace/service/inventory/InventoryLinkValidator.java
```diff
@@ -0,0 +1,94 @@
...
+public class InventoryLinkValidator {
+
+  private static final Set<GlobalIdPrefix> ALLOWED_INVENTORY_PREFIXES =
+      EnumSet.of(GlobalIdPrefix.SA, GlobalIdPrefix.SS, GlobalIdPrefix.IC, GlobalIdPrefix.IN);
+
+  /**
+   * Validates the supplied link. Errors are recorded under {@code link.*} sub-paths of the passed
+   * Errors object. {@code sourceGlobalId} (the GlobalID of the parent item the link is being
+   * attached to) is used to detect self-links.
+   */
```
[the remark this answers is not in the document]
nhanlon2: Fixed in an earlier commit: the Javadoc now states that errors are rejected relative to the current nested path and that callers wanting link.* must pushNestedPath("link") themselves.

AGENTS.md
```diff
@@ -154,6 +154,14 @@ npx vitest run src/components/MyComponent/__tests__/MyComponent.test.tsx
 - **Test setup:** Global setup in `src/__tests__/setup.ts` polyfills `localStorage`, `sessionStorage`, `TextEncoder`/`TextDecoder`.
 - **Console suppression:** Use `silenceConsole()` from test helpers to suppress expected errors.
 - **Test timeout:** 20 seconds (configured in `vitest.config.ts`).
+- - **Always run Vitest from `src/main/webapp/ui`.** The `vite.config.ts` that owns module aliases (most notably `Styles` -> `src/util/styles.ts`) lives there. Running `vitest` from a higher directory produces module-resolution failures such as `Cannot find package 'Styles'`, which look like a source bug but are a cwd problem. Either `cd src/main/webapp/ui` first or pass it explicitly to `npx vitest run --root`.
```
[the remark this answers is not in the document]
nhanlon2: Fixed in an earlier commit: the doubled list marker is no longer present in AGENTS.md.

src/main/webapp/ui/src/Inventory/components/Fields/Link/__tests__/LinkField.test.tsx
```diff
@@ -0,0 +1,266 @@
...
+
+describe("LinkField", () => {
+  afterEach(cleanup);
+
+  it("renders relation type, target global id, and name", () => {
+    renderField();
+    expect(screen.getByText(/calibration cert/i)).toBeTruthy();
+    expect(screen.getByText(/iscalibratedby/i)).toBeTruthy();
+    expect(screen.getByText(/sa42/i)).toBeTruthy();
+  });
```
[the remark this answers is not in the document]
nhanlon2: Fixed in 22f82ef6c: the remaining closest()/not.toBeNull() assertions are converted to toBeInTheDocument().

src/main/webapp/ui/src/Inventory/components/Fields/Link/__tests__/LinkField.test.tsx
```diff
@@ -0,0 +1,266 @@
...
+  it("renders pinned version label when versionPin is set", () => {
+    renderField({ link: { ...baseLink, versionPin: 4 } });
+    expect(screen.getByText(/v4|pinned to v4/i)).toBeTruthy();
+  });
+
+  it("shows 'Target deleted' badge and hides 'Open in Inventory' when the target is deleted", () => {
+    renderField({ targetDeleted: true });
+    expect(screen.getByText(/target deleted/i)).toBeTruthy();
+    expect(screen.queryByRole("button", { name: /open in inventory/i })).toBeNull();
+  });
```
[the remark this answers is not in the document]
nhanlon2: Fixed in 22f82ef6c: converted to toBeInTheDocument() per the repo lint dialect.

src/main/webapp/ui/src/Inventory/components/MoreInfoSidebar/__tests__/ReferencingRecords.test.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+    renderPanel();
+    const user = userEvent.setup();
+    await user.click(screen.getByRole("button", { name: /show references/i }));
+
+    await waitFor(() =>
+      expect(screen.getByText(/referenced in eln documents/i)).toBeTruthy(),
+    );
+    expect(screen.getByText(/linked from inventory items/i)).toBeTruthy();
+  });
```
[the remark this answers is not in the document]
nhanlon2: Addressed in an earlier commit: this test file was replaced and the current MoreInfoSidebar tests use jest-dom toBeInTheDocument() assertions.

src/main/webapp/ui/src/Inventory/components/MoreInfoSidebar/__tests__/ReferencingRecords.test.tsx
```diff
@@ -0,0 +1,116 @@
...
+      });
+    });
+
+    renderPanel();
+    const user = userEvent.setup();
+    await user.click(screen.getByRole("button", { name: /show references/i }));
+
+    await waitFor(() => expect(screen.getByText(/SA77/)).toBeTruthy());
+    expect(screen.getByText(/IsCalibratedBy/)).toBeTruthy();
+    expect(screen.getByText(/Source sample/)).toBeTruthy();
```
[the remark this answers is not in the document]
nhanlon2: Addressed in an earlier commit: this test file was replaced and the current MoreInfoSidebar tests use jest-dom toBeInTheDocument() assertions.

src/main/webapp/ui/src/components/VersionLockPicker/__tests__/VersionLockPicker.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+    cleanup();
+  });
+
+  it("renders a row for each revision and a 'latest' row", async () => {
+    renderComponent({ currentSelection: LATEST_SELECTION, onChange: vi.fn() });
+    expect(await screen.findByText(/version 1/i)).toBeTruthy();
+    expect(await screen.findByText(/version 2/i)).toBeTruthy();
+    expect(await screen.findByText(/version 3/i)).toBeTruthy();
+    expect(screen.getByText(/latest/i)).toBeTruthy();
+  });
```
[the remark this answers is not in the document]
nhanlon2: Fixed in an earlier commit: VersionLockPicker.test.tsx no longer uses toBeTruthy()/toBeNull() for DOM existence.

src/main/webapp/ui/src/stores/models/InventoryBaseRecord.tsx
```diff
@@ -1388,22 +1424,38 @@ export default class InventoryBaseRecord
...
         editing: false,
         initial: false,
-        content: field.type !== updatedField.type ? "" : field.content,
+        content: typeChanged ? "" : field.content,
       };
+      if (updatedField.type === "Link") {
+        attrs.link = updatedField.link ?? null;
+      } else if (typeChanged) {
+        attrs.link = null;
+      }
```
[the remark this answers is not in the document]
nhanlon2: Fixed in an earlier commit: updateExtraField now preserves the existing link when updatedField.link is undefined (rename-only edits no longer wipe the payload).

src/main/webapp/ui/src/stores/models/InventoryBaseRecord.tsx
```diff
@@ -1388,22 +1424,38 @@ export default class InventoryBaseRecord
...
       const justNameAndType = pick("name", "type");
-      const areEqual = sameKeysAndValues(
-        justNameAndType(field),
-        justNameAndType(updatedField),
-      );
+      const areEqual =
+        sameKeysAndValues(
+          justNameAndType(field),
+          justNameAndType(updatedField),
+        ) && updatedField.type !== "Link";
```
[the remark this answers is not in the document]
nhanlon2: Fixed in an earlier commit: equality now includes a linkUnchanged check on relationType/targetGlobalId/versionPin, so unchanged Link fields no longer dirty the record.

src/main/webapp/ui/src/Inventory/components/Fields/Link/VersionLockDialog.tsx
```diff
@@ -0,0 +1,201 @@
...
+export default function VersionLockDialog(
+  props: VersionLockDialogProps,
+): React.ReactElement | null {
+  const parsed = parseGlobalId(props.globalId);
+  const initialSelection: VersionLockSelection =
+    props.currentVersionPin == null
+      ? LATEST_SELECTION
+      : props.currentVersionPin;
+  const [selection, setSelection] =
+    useState<VersionLockSelection>(initialSelection);
```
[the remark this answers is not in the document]
nhanlon2: Fixed in an earlier commit: VersionLockDialog re-syncs selection from currentVersionPin in a useEffect on each open, exactly as suggested.

src/main/java/com/researchspace/api/v1/controller/InventoryReferencingItemsApiController.java
```diff
@@ -0,0 +1,56 @@
...
+    return forItem(GlobalIdPrefix.IC, id, user);
+  }
+
+  @Override
+  public ApiInventoryReferencingItems getReferencingItemsForInstrument(
+      @PathVariable Long id, @RequestAttribute(name = "user") User user) {
+    return forItem(GlobalIdPrefix.IN, id, user);
+  }
+
+  @Override
```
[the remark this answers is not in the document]
nhanlon2: Already addressed in the current code: findReferencingItems (the single path behind all five endpoints) rejects targets the caller cannot READ with the same non-disclosing targetNotFound error before any source query runs, and InventoryLinkManagerImplUnitTest.findReferencingItemsRejectsTargetTheCallerCannotReadWithoutQuerying covers the gate.

src/main/webapp/ui/src/Inventory/components/Fields/ExtraFields/ExtraFields.tsx
```diff
@@ -25,12 +27,84 @@ function ExtraFields({
...
+                        }
+                        icon={<CloseIcon fontSize="small" />}
+                      />
+                    </div>
+                  )}
+                  <LinkField
+                    name={ef.name}
+                    link={
+                      (ef as typeof ef & {
+                        link: {
```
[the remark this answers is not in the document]
nhanlon2: Fixed in bf8391707: the Link branch now renders LinkField only when link?.targetGlobalId is present and shows a NoValue placeholder otherwise, keeping the settings/delete affordances.

src/main/webapp/ui/src/Inventory/components/Fields/Link/iconForGlobalId.ts
```diff
@@ -0,0 +1,85 @@
...
+  globalId: string,
+  versionPin: number | null,
+): string {
+  const match = GLOBAL_ID_WITH_ID.exec(globalId);
+  if (match && match[1] === "GL") {
+    return `/gallery/item/${match[2]}`;
+  }
+  const versionSuffix = versionPin != null ? `v${versionPin}` : "";
+  return `/globalId/${globalId}${versionSuffix}`;
+}
```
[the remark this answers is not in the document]
nhanlon2: Fixed in bf8391707: openUrlForTarget rebuilds the URL from the parsed base id, so a legacy suffixed id like SS42v7 with pin 7 yields /globalId/SS42v7 rather than a double suffix.

src/main/java/com/researchspace/service/inventory/impl/LinkTargetResolverImpl.java
```diff
@@ -0,0 +1,78 @@
...
+  private static final Set<GlobalIdPrefix> INVENTORY_PREFIXES =
+      EnumSet.of(
+          GlobalIdPrefix.SA,
+          GlobalIdPrefix.SS,
+          GlobalIdPrefix.IC,
+          GlobalIdPrefix.IN,
+          GlobalIdPrefix.IT);
+
+  private static final Set<GlobalIdPrefix> ELN_BASE_RECORD_PREFIXES =
+      EnumSet.of(GlobalIdPrefix.SD, GlobalIdPrefix.NB, GlobalIdPrefix.FL, GlobalIdPrefix.GL);
```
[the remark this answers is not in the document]
nhanlon2: Fixed in bf8391707: FL removed from ELN_BASE_RECORD_PREFIXES, with a regression test asserting FL resolves false without querying BaseRecordManager.

src/main/java/com/researchspace/service/inventory/impl/LinkTargetResolverImpl.java
```diff
@@ -0,0 +1,78 @@
...
+import javax.ws.rs.NotFoundException;
+import org.apache.shiro.authz.AuthorizationException;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.orm.ObjectRetrievalFailureException;
+import org.springframework.stereotype.Component;
+
+/**
+ * Resolves a link target across modules: Inventory items via {@link InventoryPermissionUtils}, ELN
+ * items (documents, notebooks, folders, gallery files) via {@link BaseRecordManager}. Any failure
+ * to resolve a record (not found, not readable, unsupported prefix) is reported as "not readable"
```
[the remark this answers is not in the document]
nhanlon2: Fixed in bf8391707: the Javadoc now lists documents, notebooks and gallery files, and notes the prefixes mirror InventoryLinkValidator's allowed target kinds.

src/main/java/com/researchspace/dao/InventoryLinkDao.java
```diff
@@ -0,0 +1,30 @@
...
+import com.researchspace.model.inventory.field.ExtraLinkField;
+import com.researchspace.model.inventory.field.InventoryLink;
+import com.researchspace.model.inventory.field.InventoryLinkField;
+import java.util.List;
+
+/** DAO for the InventoryLink table backing Inventory Link extra-fields. */
+public interface InventoryLinkDao extends GenericDao<InventoryLink, Long> {
+
+  /** Returns all non-deleted links whose target equals the given GlobalID (ignoring version). */
+
```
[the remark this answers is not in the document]
nhanlon2: Fixed in bf8391707: the dangling Javadoc fragment is removed.

nhanlon2:
@greptileai

nhanlon2:
@greptileai

src/main/java/com/researchspace/dao/hibernate/UserDeletionDaoHibernate.java
```diff
@@ -424,6 +424,31 @@ private void deleteInventoryItems(Long userId, Session session) {
...
+    // users' surviving fields are untouched. _AUD first, then main.
+    execute(
+        userId,
+        session,
+        "delete il from InventoryLink_AUD il"
+            + " left join ExtraField ef on ef.link_id = il.id"
+            + " left join ExtraField_AUD efa on efa.link_id = il.id"
+            + " left join InventoryEntityField ief on ief.link_id = il.id"
+            + " left join InventoryEntityField_AUD iefa on iefa.link_id = il.id"
+            + " where ef.id is null and efa.id is null and ief.id is null and iefa.id is null");
```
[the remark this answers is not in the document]
nhanlon2: Fixed in 176b4dc03: the orphan sweep now runs through a parameterless executeUnparameterised helper, so no :id is bound against SQL that does not reference it.

src/main/java/com/researchspace/dao/hibernate/UserDeletionDaoHibernate.java
```diff
@@ -424,6 +424,31 @@ private void deleteInventoryItems(Long userId, Session session) {
...
+    execute(
+        userId,
+        session,
+        "delete il from InventoryLink il"
+            + " left join ExtraField ef on ef.link_id = il.id"
+            + " left join ExtraField_AUD efa on efa.link_id = il.id"
+            + " left join InventoryEntityField ief on ief.link_id = il.id"
+            + " left join InventoryEntityField_AUD iefa on iefa.link_id = il.id"
+            + " where ef.id is null and efa.id is null and ief.id is null and iefa.id is null");
```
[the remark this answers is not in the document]
nhanlon2: Fixed in 176b4dc03: same parameterless execution path as the _AUD sweep above.

src/main/webapp/ui/src/stores/models/ExtraFieldModel.ts
```diff
@@ -108,10 +124,29 @@ export default class ExtraFieldModel implements ExtraField {
         );
```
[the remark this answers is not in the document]
nhanlon2: Fixed in 176b4dc03: message now reads 'numerical'.

src/main/webapp/ui/src/eln/gallery/useReferencingInventoryItems.ts
```diff
@@ -0,0 +1,151 @@
...
+import Result from "../../util/result";
+import * as Parsers from "../../util/parsers";
+import { type LinkableRecord } from "../../stores/definitions/LinkableRecord";
+import { GLOBAL_ID_PATTERN } from "@/Inventory/components/Fields/Link/linkTarget";
+import { INVENTORY_PREFIX_ICON_DATA } from "@/Inventory/components/Fields/Link/iconForGlobalId";
+
+function prefixOf(globalId: string): string | null {
+  const match = GLOBAL_ID_PATTERN.exec(globalId);
+  return match ? match[1] : null;
+}
```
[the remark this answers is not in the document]
nhanlon2: Fixed in 176b4dc03: the local helper is removed and prefixOf is imported from iconForGlobalId.

nhanlon2:
@greptileai

src/main/java/com/researchspace/service/inventory/impl/SampleApiManagerImpl.java
```diff
@@ -321,14 +332,144 @@ private void saveNewApiFieldsIntoSampleFields(
...
+      field.setLink(null); // orphanRemoval hard-deletes the dereferenced row at flush
+      return true;
+    }
+    Long effectivePin =
+        apiLink.derivedVersionPin() != null ? apiLink.derivedVersionPin() : apiLink.getVersionPin();
+    if (existing != null
+        && target.equals(existing.getTargetGlobalId())
+        && Objects.equals(effectivePin, existing.getVersionPin())
+        && Objects.equals(apiLink.getRelationType(), existing.getRelationType())) {
+      return false; // unchanged
```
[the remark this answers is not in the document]
nhanlon2: Fixed in 2b6b046ad: applyLinkFieldValue now compares the parsed base id (prefix + dbId) plus the effective pin and relation type, mirroring ApiExtraFieldsHelper.linkChanged, with a regression test (pinnedSuffixedTargetMatchingStoredBaseIdIsUnchanged) proving an unchanged pinned payload is a no-op.

src/main/webapp/scripts/pages/recordInfoPanel.js
```diff
@@ -375,6 +385,44 @@ function _setInfoPanelInternalLinksHtml($internalLinksDiv, info, recordTypeNameT
...
+    }
+    var html = "<ul>";
+    $.each(items, function (i, item) {
+      var globalId = item.sourceGlobalId;
+      var name = item.sourceName || "";
+      var relation = item.relationType || "";
+      html += "<li><a href='/globalId/" + globalId + "'>" + globalId + "</a>: "
+        + RS.escapeHtml(name)
+        + (relation ? " <em>(" + RS.escapeHtml(relation) + ")</em>" : "")
+        + "</li>";
```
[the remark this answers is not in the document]
nhanlon2: Fixed in 2b6b046ad: the global id is now RS.escapeHtml'd in the anchor text and encodeURIComponent'd in the href.

src/main/webapp/ui/src/Inventory/components/Fields/Link/RelatedInventoryItems.tsx
```diff
@@ -0,0 +1,63 @@
...
+              {item.relationType ? (
+                <Typography variant="caption" component="em">
+                  {" "}
+                  ({item.relationType})
+                </Typography>
+              ) : null}
+            </li>
+          ))}
+        </Box>
+      )}
```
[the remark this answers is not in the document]
nhanlon2: Fixed in 2b6b046ad: rows are keyed by globalId plus index, with a test asserting two rows sharing a sourceGlobalId render without duplicate-key warnings.

src/main/webapp/ui/src/eln/gallery/components/ReferencingInventoryItemsPanel.tsx
```diff
@@ -0,0 +1,114 @@
...
+        slots={{
+          pagination: null,
+        }}
+        localeText={{
+          noRowsLabel:
+            referencing.errorMessage ?? "No related inventory items",
+        }}
+        loading={referencing.loading}
+        getRowId={(row) => row.globalId}
+        sx={{
```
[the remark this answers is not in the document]
nhanlon2: Fixed in 2b6b046ad: rows get a synthetic rowId of globalId plus index and getRowId uses it.

src/main/webapp/ui/src/components/VersionLockPicker/VersionLockPicker.tsx
```diff
@@ -0,0 +1,91 @@
...
+    let cancelled = false;
+    props.fetchVersions(props.recordId).then((rows) => {
+      if (!cancelled) {
+        setVersions(rows);
+      }
+    });
+    return () => {
+      cancelled = true;
+    };
+  }, [props.recordId, props.fetchVersions]);
```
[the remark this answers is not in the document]
nhanlon2: Fixed in 51b8b82bd: the picker now handles fetchVersions rejections, degrading to the latest-only view (empty version list) while still respecting the cancelled flag, with a regression test asserting no unhandled rejection escapes.

nhanlon2:
@greptileai

nhanlon2:
@greptileai

nhanlon2:
@greptileai

nhanlon2:
@greptileai

nhanlon2:
@greptileai

nhanlon2:
@greptileai

src/main/webapp/ui/src/Inventory/components/Fields/Link/InventoryInfoDialog.tsx
```diff
@@ -0,0 +1,136 @@
...
+                  p: 1,
+                  mb: 1,
+                }}
+              >
+                <Typography variant="body2">
+                  The information below describes{" "}
+                  <strong>version {props.versionPin}</strong> of a{" "}
+                  {iconForInventoryGlobalId(props.globalId)?.recordTypeLabel.toLowerCase() ??
+                    "record"}{" "}
+                  {props.globalId.replace(/v\d+$/, "")}, which may not be the latest
```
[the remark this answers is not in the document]
nhanlon2: False positive: optional chaining short-circuits the entire tail of the chain, so when `iconForInventoryGlobalId(...)` returns null the whole expression `...?.recordTypeLabel.toLowerCase()` evaluates to `undefined` without calling `toLowerCase()`, and the `?? "record"` fallback applies as intended. `recordTypeLabel` is a non-optional `string` on the icon data, so the inner access cannot fail either. Leaving open for a human to confirm and resolve.

nhanlon2:
@greptileai

nhanlon2:
@greptileai

nhanlon2:
@greptileai

nhanlon2:
@greptileai

nhanlon2:
@greptileai

rs-fraser:
From manual testing, I noticed a couple of issues.

**1. Can't set a link during creation**:
During creation of a new sample, add a link to any other item. Click apply on the link then Save the Sample. The link isn't saved. If you then open the link for editing, you can see all details are persisted except `Target Global ID` which is probably why it fails. (note on my localhost build I did have some unrelated-to-this-PR warnings in Inventory about loading details of inventory items, but the creation issue happens with ELN docs too, so I don't think it's related, but incase you can't recreate it)

**2. Contrast doesn't look right on `LOCK TO SELECTED VERSION` button**
Screenshot 2026-06-16 at 16 56 37

nhanlon2:
> From manual testing, I noticed a couple of issues.
> 
> **1. Can't set a link during creation**: During creation of a new sample, add a link to any other item. Click apply on the link then Save the Sample. The link isn't saved. If you then open the link for editing, you can see all details are persisted except `Target Global ID` which is probably why it fails. (note on my localhost build I did have some unrelated-to-this-PR warnings in Inventory about loading details of inventory items, but the creation issue happens with ELN docs too, so I don't think it's related, but incase you can't recreate it)
> 
> **2. Contrast doesn't look right on `LOCK TO SELECTED VERSION` button** Screenshot 2026-06-16 at 16 56 37

@rs-fraser fixed those two bugs. Also removed 'Edit' button. Keeping the open button because none of the other pills in the link ui are clickable but have added globalid of target to its aria label. Have removed 'open' button from info panel when target is not available. Link names for links made from templates now go inside the box outline in the same position as manual links made directly.

nhanlon2:
@greptileai

rs-fraser (approved):
I've reviewed the back-end code and been through the manual test steps and all looks good.

src/main/webapp/ui/src/components/Inputs/FieldLabel.tsx
```diff
@@ -87,14 +88,24 @@ export default function FieldLabel({
       ? "legend"
       : "label";
 
+  // Reset styles added by the browser when setting the `component` prop, then
+  // merge any caller overrides. `Array.isArray(sx)` narrows the matched branch to
+  // `any[]` (a TS quirk for unions containing an array), so the normalised list
+  // is cast to a concrete element type to avoid an unsafe `any` spread.
+  const mergedSx: SxProps<Theme> = [
```
tchoi-rs: There is a mergeSx utility function

src/main/webapp/ui/src/Inventory/components/Fields/Link/ElnRecordInfoDialog.tsx
```diff
@@ -0,0 +1,240 @@
...
+    this.state = { hasError: false };
+  }
+
+  static getDerivedStateFromError(): { hasError: boolean } {
+    return { hasError: true };
+  }
+
+  render(): React.ReactNode {
+    if (this.state.hasError) {
+      // A pinned link that can no longer resolve its version gets a specific message
```
tchoi-rs: An ErrorBounday is designed to contain unexpected errors so exceptions do not crash the whole application. This appears to be an error that should be handled explicitly at the query stage.

src/main/webapp/ui/src/Inventory/components/Fields/Link/DocumentSections.tsx
```diff
@@ -0,0 +1,472 @@
...
+        <Typography variant="body2">
+          This {recordTypeName} is linked by {count} {docOrDocs}.
+        </Typography>
+        <Button
+          size="small"
+          disabled={loading}
+          onClick={() => {
+            setLoading(true);
+            void getLinkedByRecords(info.id)
+              .then(setLinked)
```
tchoi-rs: Use async syntax

src/main/webapp/ui/src/Inventory/components/Fields/Link/DocumentSections.tsx
```diff
@@ -0,0 +1,472 @@
...
+            </Link>
+          </Typography>
+        ))}
+    </Box>
+  );
+}
+
+// ELN content stylesheets, referenced at their served URLs from the preview iframe
+// head (enumerated from structuredDocumentMainPanel.jsp + tinymce5_configuration.js).
+const PREVIEW_STYLESHEETS = [
```
tchoi-rs: To reduce the cross-contamination between Reactified code and legacy Workspace code I would prefer not to display a document preview here

src/main/webapp/ui/src/Inventory/components/Fields/Link/ElnRecordInfoDialog.tsx
```diff
@@ -0,0 +1,240 @@
...
+        <Box sx={{ display: "flex", alignItems: "center", gap: 1 }}>
+          {iconData && <RecordTypeIcon record={iconData} aria-hidden />}
+          {props.globalId}
+        </Box>
+      </DialogTitle>
+      <DialogContent dividers>
+        {recordId === null ? (
+          <Alert severity="error">{UNAVAILABLE_MESSAGE}</Alert>
+        ) : (
+          <QueryClientProvider client={queryClient}>
```
tchoi-rs: Use shared QueryClient

src/main/webapp/ui/src/Inventory/components/Fields/Link/ElnFolderBrowser.tsx
```diff
@@ -0,0 +1,328 @@
...
+        getId={(node) => node.id.toString()}
+        itemChildrenIndentation={20}
+        slots={{ expandIcon: ExpandArrow, collapseIcon: CollapseArrow }}
+        sx={{
+          // Compact rows like the ELN tree (fileTreeBrowser.css cannot bind to this
+          // MUI DOM; see ADR 0001). The selected row uses the solid theme primary
+          // colour so the current selection is unmistakable before clicking Choose.
+          // x-tree-view 9 marks state with data attributes on the content row (the
+          // Mui-selected class no longer exists), so the selectors target those.
+          "& .MuiTreeItem-content": { minHeight: "25px", borderRadius: 0, py: 0 },
```
tchoi-rs: These magic class keywords should be imported from MUI

src/main/webapp/ui/src/Inventory/components/Fields/Link/LinkField.tsx
```diff
@@ -0,0 +1,220 @@
...
+                  label="No access"
+                  data-test-id="LinkField-noAccess"
+                />
+              </Tooltip>
+            )}
+            {!openBlocked && (
+              <Button
+                size="small"
+                startIcon={<OpenInNewIcon />}
+                onClick={handleOpen}
```
tchoi-rs: This needs to be a static link instead for accessibility reasons. I looked at `handleOpen` and I don't think there's any reason why it needs to be hidden behind a closure?

src/main/webapp/ui/src/Inventory/components/MoreInfoSidebar/SidebarBody.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Presentational body of the MoreInfo side panel. Renders the labelled fields
+ * for an inventory record. Used by the main Sidebar drawer and by the
+ * InventoryInfoDialog modal so both surfaces share one implementation.
+ */
+function SidebarBody({ record, factory }: SidebarBodyArgs): React.ReactNode {
+  return (
+    <Stack
+      spacing={5}
```
tchoi-rs: I think `2` would be a more appropriate spacing

src/main/webapp/ui/src/Inventory/components/Fields/Link/DocumentSections.tsx
```diff
@@ -0,0 +1,472 @@
...
+      </Box>
+    );
+  }
+
+  return (
+    <Box>
+      <Typography variant="body2">
+        This {recordTypeName} is linked by:
+      </Typography>
+      <Box component="ul" sx={{ pl: 3, my: 0.5 }}>
```
tchoi-rs: This should be a `<List>`

src/main/webapp/ui/src/Inventory/components/Fields/Link/DocumentSections.tsx
```diff
@@ -0,0 +1,472 @@
...
+      });
+    return () => {
+      active = false;
+    };
+  }, [info.oid.idString]);
+
+  const isShared = Boolean(info.shared) || Boolean(info.implicitlyShared);
+
+  return (
+    <Box sx={{ mt: 1 }}>
```
tchoi-rs: In a stack of subcomponents, margin/padding control should be given to the parent component

tchoi-rs (changes requested):

tchoi-rs:
Some additional comments by Ponytail:

# Code Review: RSDEV-1131 Inventory Link Related Items

**Scope reviewed:** ~3,500 lines of new source (the 17k line diff is mostly test files, which look thorough and are out of scope here). This review focuses on simplification and reducing duplication, not correctness or security, which should get a separate pass.

Overall the feature is well structured. The main theme below is duplication that crept in across the new Link components. None of these are blockers, but addressing the high-priority items would meaningfully cut the surface area future maintainers have to read.

---

## High priority

### 1. Duplicated link-editor UI (~180 lines each)
**Files:** `Fields/ExtraFields/UpdateField.tsx`, `Sample/Fields/TemplateFields/LinkFieldValue.tsx`

The link-editor block (relation-type Autocomplete, target chip/field, Browse Inventory/ELN buttons, version pill, Apply/Discard controls, the three dialogs, and the staged-vs-committed dirty tracking) is duplicated almost verbatim across both files.

**Suggestion:** Extract a single `<LinkEditor>` component that takes the staged value and an `onApply` callback. Each caller keeps its own commit logic. This is the single largest cleanup available.

### 2. Duplicated presentational components
**File:** `Fields/Link/GallerySections.tsx` (L844-882)

`MetaRow` and `GlobalIdLink` are byte-for-byte identical to the versions in `DocumentSections.tsx`.

**Suggestion:** Lift both into one shared module imported by both sections.

### 3. Global ID regex defined five times
The global-ID parsing regex is re-spelled in five places. `linkTarget.ts` already exports `GLOBAL_ID_PATTERN` and `prefixOf`.

| Location | Current | Replace with |
|---|---|---|
| `iconForGlobalId.ts:4` | `GLOBAL_ID_WITH_ID` | `GLOBAL_ID_PATTERN` |
| `ElnRecordInfoDialog.tsx:51` | `numericIdOf` | `GLOBAL_ID_PATTERN.exec(id)?.[2]` |
| `LinkField.tsx:74` | inline `/^([A-Z]{2})(\d+)/` | `prefixOf` / `GLOBAL_ID_PATTERN` |

---

## Medium priority

### Frontend

- **`ElnRecordInfoDialog.tsx:60-117`** — The hand-rolled `DialogErrorBoundary` class duplicates the app-wide `ErrorBoundary` component. Pass a custom fallback to the existing one instead.
- **`ElnRecordInfoDialog.tsx:120-150`** — The per-dialog `new QueryClient()` plus the `openCount` remount counter and `wasOpen` ref add a lot of machinery to force a retry on reopen. A shared client plus `key={globalId}` to remount achieves the same thing.
- **`InventoryBaseRecord.tsx:610-625` and `UpdateField.tsx:58-70`** — The `(ef as ExtraField & { link?... })` casts are now redundant: `ExtraField` declares `readonly link`. Read `ef.link` directly. The cast hides a type that already exists rather than adding flexibility.
- **`useReferencingInventoryItems.ts:54-79`** — The `LinkableInventoryRecord` class exists only to expose three getters to a prop. A plain object literal implementing `LinkableRecord` is shorter and clearer.

### Backend

- **`InventoryReferencingItemsApiController.java:19-45`** — The four typed endpoints (Sample, SubSample, Container, Instrument) each build a GlobalId and delegate to the same `findReferencingItems` that the generic `/referencingItems/{globalId}` route already serves for any prefix. Consider dropping the four convenience routes. **Verify the published API contract first**, since this is external surface.
- **`LinkTargetSnapshotResolver.java`** — Interface with one implementation and one caller (`InventoryLinkManagerImpl`). It is a non-transactional `@Component`, so there is no Spring AOP reason to keep the seam. Fold the impl into the caller, or make it a concrete class. (Note: the `*Manager` interfaces should stay, since those do need the interface for transaction proxying.)
- **`LinkTargetSnapshotResolverImpl.java:1937`** — `isInventoryPrefix` reimplements a 5-way `||` that already exists as the `INVENTORY_PREFIXES` EnumSet and again as `InventoryLinkValidator.ALLOWED_TARGET_PREFIXES`. Share one `EnumSet.contains(prefix)`.

---

## Low priority / nits

- **`InventoryInfoDialog.tsx:107-126`** — The four-state union (`init`/`loading`/`success`/`fail`) is heavier than needed; `init` and `loading` render identically. Collapse to a `{data, error} | null` shape.
- **`iconForGlobalId.ts:67`** — `iconForInventoryGlobalId` has one caller and is just `INVENTORY_PREFIX_ICON_DATA[prefixOf(id)]`. Inline it.
- **`VersionLockPicker.tsx:25-31`** — `recordId` is passed to the picker but the dialog's `fetchVersions` ignores the argument. Drop the prop.
- **`InventoryBaseRecord.tsx:595-609` and `InventoryRecord.ts:375-382`** — Inline `link?: { relationType; targetGlobalId; versionPin }` literals re-spell `ExtraInventoryLink`. Reuse the named type.
- **`ExtraFieldModel.ts:150`** — `typeNameFromApi` is a three-branch function; a lookup object `{ text: "Text", number: "Number", link: "Link" }[t]` is one line and matches the existing `apiTypeFor` style.
- **`FieldLabel.tsx:91-101`** — The `mergedSx` extraction plus `SystemStyleObject` cast only silences a self-introduced annotation; the original one-liner type-checked on main. Revert.
- **`ReferencingInventoryItemsPanel.tsx:41-48`** — `setTimeout(autosizeColumns, 10)` can be replaced by the DataGrid `autosizeOnMount` and `autosizeOptions` props, which removes the effect and the `useGridApiRef`. The `initialState={{ columns: {} }}` at L8 is a no-op.
- **`DataCiteRelationType.java:65`** — `allValues()` has no callers; remove it.

---

**Estimated reduction:** roughly 450 lines if all items are taken, with about 180 of those from the `<LinkEditor>` extraction alone.

src/main/webapp/ui/src/__tests__/setup.ts
```diff
@@ -105,3 +105,25 @@ if (typeof globalThis.IntersectionObserver !== "function") {
   globalThis.IntersectionObserver = IntersectionObserverMock;
 }
 
+/*
+ * jsdom does not implement window.matchMedia, but some dependencies (TinyMCE,
+ * MUI useMediaQuery) call it at module-evaluation time, before a test file's
+ * own per-file mock import can run. Defining it here guarantees it exists
+ * before any module is evaluated.
+ */
+if (typeof window !== "undefined" && typeof window.matchMedia !== "function") {
```
tchoi-rs: This is managed via `__mocks__` instead and is not needed here

src/main/webapp/ui/src/Inventory/components/MoreInfoSidebar/SidebarBody.tsx
nhanlon2: Done — changed the `SidebarBody` `Stack` spacing from `5` to `2` in 31cd79b48.

src/main/webapp/ui/src/components/Inputs/FieldLabel.tsx
nhanlon2: Good catch — replaced the hand-rolled sx normalisation (and the `SystemStyleObject` cast) with the shared `mergeSx` utility from `@/modules/common/utils/styles` in cf70b2650.

src/main/webapp/ui/src/Inventory/components/Fields/Link/ElnFolderBrowser.tsx
nhanlon2: Done in 28294e1cb — the `.MuiTreeItem-content` magic strings now use the `treeItemClasses.content` constant from `@mui/x-tree-view/TreeItem` (matching RecordTree and the gallery TreeView). The `[data-selected]`/`[data-focused]` parts stay literal since they're data attributes, not MUI class names.

src/main/webapp/ui/src/Inventory/components/Fields/Link/DocumentSections.tsx
nhanlon2: Done in aa8646eb6 — the "Show linked docs" handler now uses an async/await function (`showLinked`) with try/catch/finally instead of the `.then/.catch/.finally` chain.

src/main/webapp/ui/src/Inventory/components/Fields/Link/DocumentSections.tsx
nhanlon2: Done in 54769501d — converted this list (the linked-by docs) to a semantic MUI `List`/`ListItem`, keeping the disc markers and compact spacing. I also converted the identical `Box component="ul"` in the Sharing section just below, for consistency.

src/main/webapp/ui/src/Inventory/components/Fields/Link/DocumentSections.tsx
nhanlon2: Agreed — removed the document preview in eb0b7ca7e. Dropped the `DocumentPreview` component, its `PREVIEW_STYLESHEETS`, the `DOMPurify`/`CircularProgress` imports it solely used, and the now-orphaned `modules/workspace/documentPreview.ts` wrapper (plus its tests). The legacy `/ajax/preview` endpoint it wrapped stays untouched since `recordInfoPanel.js` still uses it. Note: tsc passes; I did not run the vitest suite per our test-execution policy this session.

src/main/webapp/ui/src/Inventory/components/Fields/Link/DocumentSections.tsx
nhanlon2: Done in 9fff31c74 — DocumentSections now wraps its sections in a `Stack` with a single `spacing`, and the subcomponents (LinkedByDocs, RelatedInventoryItems, SharingAndPublication) no longer set their own `mt`. Since RelatedInventoryItems is also mounted by GallerySections, I moved that spacing to the GallerySections parent side too.

src/main/webapp/ui/src/Inventory/components/Fields/Link/LinkField.tsx
nhanlon2: Done in 2adb7e375. You're right — there was no need for the closure. The Open control is now a static anchor (`<Button href=... target="_blank" rel="noopener noreferrer">`, which MUI renders as an `<a>`). LinkField computes the href itself: the RSDEV-1141 versioned-viewer route (`/inventory/{type}/{id}?version=N`) for a version-pinned inventory target, otherwise the standard `openUrlForTarget` route. The `onOpen` prop is gone; ExtraFields and the template LinkFieldValue no longer pass it. Tests now query the Open control as a link and assert its href.

src/main/webapp/ui/src/Inventory/components/Fields/Link/ElnRecordInfoDialog.tsx
nhanlon2: Addressed: the QueryClient is now a single module-level instance shared by every `ElnRecordInfoDialog` (commit bea6527f0), matching the codebase convention. Follow-up b918aebe updates the stale JSDoc and drops the now-stable constant from the open-effect dependency array.

src/main/webapp/ui/src/Inventory/components/Fields/Link/ElnRecordInfoDialog.tsx
nhanlon2: Agreed, and done in 87830355. A failed record-info load (deleted / not readable / unknown version) is an expected outcome, so it's no longer routed through an error boundary. The query hook now uses a plain `useQuery` instead of `useSuspenseQuery`, and `DialogBody` renders the loading skeleton, the inline error Alert (including the version-pinned "no longer available" variant), and the content explicitly from the query state. The `DialogErrorBoundary` class, the `Suspense` wrapper, and the manual `clear()`+`openCount` remount machinery are all gone; re-opening the dialog remounts the body and React Query refetches a previously-errored query on its own. The full ElnRecordInfoDialog suite (16 tests, including the error and reopen-after-failure cases) passes.

src/main/webapp/ui/src/__tests__/setup.ts
nhanlon2: Done in 941ae03b. Removed the global `window.matchMedia` block from setup.ts and replaced the two inline hoisted stubs (SidebarBody and ExtraFields.link tests) with the shared `import "@/__tests__/__mocks__/matchMedia";`, matching the per-file convention used across the suite.

Verified safe with a test run: I ran every test suite in the touched areas with the global shim removed (104 files / 439 tests across Link, ExtraFields, MoreInfoSidebar, TemplateFields, Template/Fields, VersionLockPicker, gallery, workspace and stores/models). All pass, with zero `matchMedia is not a function` failures, so nothing else relied on the global block.

src/main/webapp/ui/src/Inventory/components/Fields/Link/LinkField.tsx
nhanlon2: Small follow-up to this: while running the suites I found one more test that drove the removed `onOpen` (`LinkFieldValue.test.tsx`'s "opens a Gallery target..." test, via a mocked LinkField button). Cleaned that up in b185dfa1 the same way as the ExtraFields tests; the open-URL routing stays covered by LinkField's and iconForGlobalId's own tests.

tchoi-rs:
Also I was testing the feature and noticed the icons are misaligned:

Screenshot 2026-06-17 at 14 24 15

src/main/webapp/ui/src/Inventory/components/Fields/Link/ElnFolderBrowser.tsx
tchoi-rs: There are other instances of this in the changeset that also needs addressing

src/main/webapp/ui/src/Inventory/components/Fields/Link/DocumentSections.tsx
tchoi-rs: There are other instances of this which will also need to be fixed.

src/main/webapp/ui/src/Inventory/components/Fields/Link/ElnRecordInfoDialog.tsx
tchoi-rs: Actually, there should be a global QueryClient at the inventory app level, so I don't think this provider is even needed

src/main/webapp/ui/src/Inventory/components/Fields/Link/DocumentSections.tsx
tchoi-rs: The changeset left a bunch of unnecessary comments that will need to be removed.

nhanlon2:
> Also I was testing the feature and noticed the icons are misaligned:
> 
> Screenshot 2026-06-17 at 14 24 15

@tchoi-rs Could you provide some detail as I understand nothing from that screenshot.

src/main/webapp/ui/src/Inventory/components/Fields/Link/DocumentSections.tsx
nhanlon2: Done in 48c4f6c3. The two other hand-rolled `<Box component="ul">` bullet lists in this changeset are now semantic MUI `List`/`ListItem`, matching the DocumentSections conversion:
- `GallerySections.tsx` "referenced by" list
- `RelatedInventoryItems.tsx` "related inventory items" list

(`Layout/Sidebar.tsx` also has list markup but is outside this PR's diff, and already uses `<List>`.) Behaviour is unchanged: the existing `listitem`-role assertions still pass.

src/main/webapp/ui/src/Inventory/components/Fields/Link/ElnFolderBrowser.tsx
nhanlon2: Done in b6a02aa3. The other hard-coded `.MuiXxx-yyy` magic class strings in the changeset now use the imported MUI class constants:
- `SidebarBody.tsx` selector → `` `& .${formGroupClasses.root} .${buttonClasses.root}` `` (from `@mui/material/FormGroup` / `@mui/material/Button`)
- `ElnFolderBrowser.test.tsx` now derives both its content-row `querySelector` and its CSS-rule regex from `treeItemClasses.content`

The remaining literal selectors in `ElnFolderBrowser` are the `[data-selected]` / `[data-focused]` data attributes, which are not MUI class names.

src/main/webapp/ui/src/Inventory/components/Fields/Link/ElnRecordInfoDialog.tsx
nhanlon2: Good call - done in 23a1019a. There already is an app-level client: the Inventory app mounts under `App.tsx`'s `QueryClientProvider` (`App` -> `Router` -> `/inventory/*` -> `InventoryRouter` -> the record forms that render the link card and this dialog). So the dialog's own module-level `QueryClient` + `QueryClientProvider` were redundant; I removed them and it now uses the ancestor client.

The only test that renders the real dialog (`ElnRecordInfoDialog.test`) supplies a `QueryClient` the same way the app does. No app-level wiring change was needed since `App.tsx` already provides one above the inventory routes.

src/main/webapp/ui/src/Inventory/components/Fields/Link/DocumentSections.tsx
nhanlon2: Done in dc3edf00. Removed the comment noise the changeset left behind in this cluster:
- the "no outer margin / parent controls spacing" narration left over from the spacing refactor (`DocumentSections` x2, `RelatedInventoryItems`) - the `Stack`/parent structure is self-documenting
- comments in `LinkFieldValue`'s committed-link display path that just restated the code (the settings-cog edit affordance, the field-name pass-through); the one genuine "why" (the FormField label is hidden) is kept as a single concise line

I deliberately left the comments that document non-obvious behaviour (validation edge cases, the no-access/Open-button rules, ADR-0002, the x-tree-view `data-selected` quirk) rather than strip them. If there are specific ones in that set you also consider unnecessary, point me at them and I'll remove those too.

tchoi-rs (approved):

PR #806, opened 2026-05-29
RSDEV-1138: Persist molecular weight for PubChem-added reagents

rs-fraser:
## Description

`toStoichiometryRequest` dropped `molecularWeight` when serialising a new (unsaved) reagent, returning `base as NewMolecule` without it. A reagent added from PubChem therefore saved with a null molecular weight, which broke its mole and yield calculations on first save.

This change carries `molecularWeight` through on the `NewMolecule` payload so it is persisted when the stoichiometry table is first saved.

## Changes

- `useEditableStoichiometryTable.ts` — spread `molecularWeight` (falling back to `undefined`) onto the `NewMolecule` request object instead of returning `base` bare.
- `useEditableStoichiometryTable.test.tsx` — added coverage asserting molecular weight survives serialisation for newly inserted molecules.

### DIFF
--- a/src/main/webapp/ui/src/tinyMCE/stoichiometry/useEditableStoichiometryTable.ts
+++ b/src/main/webapp/ui/src/tinyMCE/stoichiometry/useEditableStoichiometryTable.ts
@@ -105,7 +105,10 @@ function toStoichiometryRequest(
         throw new Error("New reagents must have a name");
       }
 
-      return base as NewMolecule;
+      return {
+        ...base,
+        molecularWeight: m.molecularWeight ?? undefined,
+      } as NewMolecule;
     }),
   };
 }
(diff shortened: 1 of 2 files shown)


rs-fraser:
@greptileai

rs-fraser:
@greptileai

src/main/webapp/ui/src/tinyMCE/stoichiometry/useEditableStoichiometryTable.ts
```diff
@@ -105,7 +105,11 @@ function toStoichiometryRequest(
         throw new Error("New reagents must have a name");
       }
 
-      return base as NewMolecule;
+      return {
+        ...base,
+        formula: m.formula ?? undefined,
```
tchoi-rs: These two can be added directly to `base` and Line 96/97 removed

tchoi-rs (changes requested):

src/main/webapp/ui/src/tinyMCE/stoichiometry/useEditableStoichiometryTable.ts
rs-fraser: Done

tchoi-rs (approved):

PR #771, opened 2026-05-16
feat: migrate frontend build from Webpack to Vite with HMR support

tchoi-rs:
Investigation and implementation done by GitHub Copilot AI agent on behalf of the developer.

Migrates the frontend build pipeline from Webpack to Vite, with a same-origin reverse proxy for Hot Module Replacement during local development.

## Changes

- **ViteDevServerProxyServlet**: reverse-proxies `/ui/dist/*` to local Vite dev server when `reactDevMode=true`
- **BundleTag / ViteClientTag**: JSP custom tags that read Vite's manifest for production asset loading and emit HMR client scripts in dev
- **viteBundleLoader.mjs**: JS-side loader for TinyMCE iframe dialogs
- **HTML → JSP conversion**: inventory, apps, gallery, about, and TinyMCE dialog entrypoints converted to JSPs with `<rst:viteClient />` for HMR preamble support
- **vite.config.ts + bundleEntries.json** replace webpack.config.mjs as the single source of truth for entry points
- **Documentation**: all developer docs updated to reflect Vite (AGENTS.md, CLAUDE.md, GettingStarted, Glossary, AnalysingJavaScriptBundles, NewEditorPlugin, agent skills)
- Proxy servlet Javadoc documents synchronous blocking nature and content-type fallback rationale
- Three-tier manifest caching strategy (production / dev / HMR) documented in BundleTag

PR #811, opened 2026-06-02
RSDEV-1158: Gate frontend (Vite) build behind opt-in generateReactDistFiles profile

rs-fraser:
RSDEV-1158

## Description ##

After the Webpack-to-Vite migration (#771), the `frontend-maven-plugin` (Node install + `npm ci` + Vite build) became always-on, so even `mvn compile` / `mvn test` triggered a full Node install and Vite production build.

This PR restores the opt-in model: the plugin now lives in a `generateReactDistFiles` Maven profile, activated only by `-DgenerateReactDist`, as was the case before the vite migration

## Which flag, when ##

| You want | Command | Frontend built? |
|---|---|---|
| Backend-only compile/test | `mvn clean compile` / `mvn clean test` | No |
| Fast backend WAR | `mvn clean package` | No |
| **Deployable WAR (with UI)** | `mvn clean package -DgenerateReactDist` | **Yes** |
| Jetty serving a local prod bundle | `mvn jetty:run -DgenerateReactDist` | Yes |
| Everyday dev loop | `mvn jetty:run -DreactDevMode=true` (+ `npm run serve`) | No (Jetty proxies to the Vite dev server) |

**Rule:** pass `-DgenerateReactDist` whenever the artifact needs a working UI; omit it for backend-only builds. `-DreactDevMode=true` is a separate flag for the dev-server proxy, not a build toggle. The profile's executions stay on the implicit `generate-resources` phase, so `jetty:run -DgenerateReactDist` rebuilds the bundle too.

## Notes ##

- Removed reactDevMode as a maven property, as it's no longer used in `pom.xml`, although it's still passed as a system property used to proxy to the vite dev server, when set to true.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -87,8 +87,13 @@ npm run serve  # Vite dev server
 ### Full Build (no tests)
 
 ```bash
-mvn clean package -DskipTests=true
+mvn clean package -DgenerateReactDist -DskipTests=true
 ```
+
+The `-DgenerateReactDist` flag activates the `generateReactDistFiles` profile that
+runs `npm ci` and the Vite production build and bundles `dist/` into the WAR. It is
+opt-in: omit it and `mvn compile`/`test`/`package` skip the frontend build entirely
+(useful for fast backend-only builds, but the resulting WAR has no frontend).
 
 ## Testing
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -87,8 +87,13 @@ npm run serve  # Vite dev server
 ### Full Build (no tests)
 
 ```bash
-mvn clean package -DskipTests=true
+mvn clean package -DgenerateReactDist -DskipTests=true
 ```
+
+The `-DgenerateReactDist` flag activates the `generateReactDistFiles` profile that
+runs `npm ci` and the Vite production build and bundles `dist/` into the WAR. It is
+opt-in: omit it and `mvn compile`/`test`/`package` skip the frontend build entirely
+(useful for fast backend-only builds, but the resulting WAR has no frontend).
 
 ## Testing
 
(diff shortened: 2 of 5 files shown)


mKowalski256:
@greptileai would the npm/vite build still run with `mvn clean jetty:run`? when running locally with that command i'd like to have front-end built.

mKowalski256:
> **Options to address this:**
> 
> 1. **Use `mvn prepare-package jetty:run` for local dev** — explicitly runs the lifecycle through `prepare-package` before starting Jetty. This is the cleanest workaround without changing the pom.

Thank you, that sounds like a good clean workaround that we should just document in DevDocs.

rs-fraser:
> > **Options to address this:**
> > 
> > 1. **Use `mvn prepare-package jetty:run` for local dev** — explicitly runs the lifecycle through `prepare-package` before starting Jetty. This is the cleanest workaround without changing the pom.
> 
> Thank you, that sounds like a good clean workaround that we should just document in DevDocs.
Good spot, will update the PR

rs-fraser:
Thanks @greptile-apps, good catch on the `jetty:run` fork stopping at `test-compile`. One detail to add: the documented dev loop runs Jetty with `-DreactDevMode=true`, which sets `<skip>` on all three frontend executions, so `jetty:run` does not build the bundle in that mode anyway (it proxies to the Vite dev server via `npm run serve`). So the everyday workflow is unaffected.

The only impacted case is running Jetty against a freshly built production bundle (`reactDevMode=false`). For that we've gone with the documentation route (option 1) rather than adding pom machinery: GettingStarted.md now notes to use `mvn prepare-package jetty:run` for that scenario. Pushed in 2d3ee59.

rs-fraser:
Update: changed direction based on review discussion. Rather than re-phasing the always-on build to `prepare-package`, I've restored the pre-Vite opt-in model: the `frontend-maven-plugin` now lives in a `generateReactDistFiles` profile activated only by `-DgenerateReactDist` (pushed in a5be77c).

This sidesteps the `jetty:run` gap you flagged entirely: inside the profile the executions keep the implicit `generate-resources` phase, which the `jetty:run` fork does reach, so `mvn jetty:run -DgenerateReactDist` rebuilds the bundle. Plain `mvn compile`/`test` stay Node-free because the profile is inactive.

The trade-off is that packaging no longer builds the frontend implicitly, so I've added `-DgenerateReactDist` to the three WAR/verify Jenkins stages (feature-branch package, prodRelease package, `clean verify`) and documented the flag in GettingStarted.md / AGENTS.md / CLAUDE.md. @greptile-apps a re-review of the new approach would be appreciated.

rs-fraser:
@greptileai

mKowalski256 (approved):
Looks good but please merge after 2.23 release is out.

PR #822, opened 2026-06-04
Fix Vite MIME issues

tchoi-rs:
## Description ##
Fix Vite bundles not being served with MIME types, causing the browser to block the scripts.

Fixes PRT-1087


mKowalski256 (approved):
That's for [PRT-1087](https://researchspace.atlassian.net/browse/PRT-1087), right? If so please put the number in PR name.

Looks good to me, on test build I see the `.js` chunks have content-type `text/javascript` rather than `application/javascript`.

tchoi-rs:
@mKowalski256 According to [RFC9239](https://www.ietf.org/rfc/rfc9239.pdf), `text/javascript` is the correct MIME.

PR #823, opened 2026-06-04
Add experimental Dockerized per-worktree dev stack

tchoi-rs:
## Description ##
Provides a way to boot a Docker-based, completely isolated instance of RSpace. Allows for worktree-independent working.

- Boot complete RSpace (db, backend, frontend) with ./docker/dev/rspace-dev up
- Per-worktree isolation: separate databases, volumes, auto-allocated ports
- Frontend hot-reload (HMR) and backend hot-redeploy via reload command
- Eliminates local Java, Maven, Node, MariaDB installation
- Worktree source bind-mounted; edits on host are live in containers
- Shared Maven and pnpm caches across worktrees to save bandwidth
- Update Vite config for HMR over bind mounts (polling, host/port override)


rs-fraser:
This looks great. Claude flagged this as an issue though:

### `restart` after a fresh `up` / `reset-db` silently wipes the database

  `RSPACE_DB_MODE` is decided at container **creation** and baked into the container's
  environment. After the first `up` (or any `reset-db`), the `app` container is created
  with `drop-recreate-db`. `cmd_restart` is:

  ```bash
  cmd_restart() { compose restart app; }
  ```

  `docker compose restart` restarts the **same** container with the **same** baked-in env,
  so it replays `drop-recreate-db` — every `restart` drops and recreates the schema. A
  later `up` happens to recover (the launcher exports `keepdbintact`, compose sees the env
  change and recreates the container), but until then each `restart` is a data wipe.

  There's a second, compounding issue: the README tells you to set a debug knob in `.env`
  and then `rspace-dev restart`, but `compose restart` doesn't re-read `.env` at all — so
  that instruction is a no-op.

  #### Suggested fix

  ```bash
  cmd_restart() {
    db_volume_exists || die "no database for this worktree yet - run 'rspace-dev up' first"
    RSPACE_DB_MODE="keepdbintact" compose up -d --force-recreate app
  }
  ```

  - `--force-recreate` re-evaluates the environment from the launcher **and** `.env`, which
    also fixes the "set it in `.env` then restart" no-op for free.
  - Explicitly forcing `keepdbintact` means even a container previously created by
    `reset-db` can't replay the drop.
  - The `db_volume_exists` guard covers the only case where `keepdbintact` would be wrong
    (a restart before any `up`) and points the user at `up`.
  - Recreating only `app` is safe — it's the dependent side of
    `network_mode: service:frontend`, which the compose comments already note is the safe
    one to recreate.
  - `reset-db` stays destructive (as intended); restart is now safe to run after it.

docker/dev/entrypoint-app.sh
```diff
@@ -0,0 +1,93 @@
...
+fi
+
+echo "[entrypoint] Starting RSpace (db mode: ${DB_MODE}) ..."
+echo "[entrypoint] App will be reachable on the host at http://localhost:${APP_PUBLIC_PORT}"
+
+# jetty:run executes the Maven lifecycle up to test-compile in this same JVM, so
+# the -D system properties below reach Spring at runtime and the Maven build
+# plugins (e.g. the drop-recreate-db SQL step) at build time.
+exec mvn -B jetty:run \
+  -Denvironment="${DB_MODE}" \
```
nhanlon2, line 100: almost essential to have the following:
-DlegacyAssetCacheBustingInDevMode=true (or a settable value) and 
-Drs.indexOnstartup=false (or a settable value)
-Dchemistry.provider=indigo (or a settable value)
-Dchemistry.service.url=http://howler:8075 (or a settable value)

tchoi-rs: I'll implement the first two. The expectation for flags that normally live in `deployment.properties` is that they should come from `deployment.properties` instead - is this a big problem otherwise?

tchoi-rs: Nevermind, I almost forgot that I already made the changes locally to also run Chemistry on the side

nhanlon2 (approved):

PR #826, opened 2026-06-05
RSDEV-1160: fix TinyMCE React components mounting twice in Vite dev mode

rs-fraser:
## Problem

In Vite dev mode (`reactDevMode=true`), several TinyMCE-embedded React components mount twice: the Ketcher "Insert chemical structure" dialog, the post-save `SidebarInfo` panel, and the `InternalLink` dialog. Symptoms:

- Duplicated UI (e.g. two sidebars after saving).
- Console warning: `ReactDOMClient.createRoot() on a container that has already been passed to createRoot()`.
- Ketcher not registering drawn structures (you draw on one editor instance while validation reads the other), so Insert reports an empty molecule.

Production bundles are unaffected, so this is dev-only.

## Root cause

The Webpack→Vite migration (`7ef3b7a38`) left three `src/main/webapp/ui/bundleEntries.json` entries pointing at `.jsx` files that are actually `.tsx`:

- `tinymceKetcher` → `KetcherTinyMce.tsx`
- `tinymceSidebarInfo` → `SidebarInfo.tsx`
- `internalLink` → `InternalLink.tsx`

The Vite dev server serves each module under two URLs (the requested `.jsx` and the resolved `.tsx`), so it is evaluated twice. Each instance runs its module-level setup (event listener + `createRoot`) on the same container, producing the duplicate mount.

## Fix

Correct the three entry extensions to `.tsx` (3 lines). All bundle entries now resolve to a real file on disk.

## Note for reviewers

Editing `bundleEntries.json` requires a full `jetty:run` restart to take effect in `reactDevMode`, because `BundleTag` caches the entrypoints map in the servlet context.

Ticket: RSDEV-1160


rs-fraser:
@greptileai

tchoi-rs (approved):

PR #832, opened 2026-06-08
PRT-1005: reject empty CSV content on batch user upload

mKowalski256:
### Problem
  `POST /system/userRegistration/csvUpload` returned `200 OK` when **Upload** was
  clicked in the Batch upload users dialog without selecting a file. The handler
  never checked the multipart part, so an empty file was parsed into an empty
  (but successful) result.

  ### Changes
  - **Backend:** `batchUploadParseCsvFile` now returns `400 Bad Request` with a
    distinct `ErrorList` message when the uploaded file is null or empty;
    the success path is unchanged.
  - **i18n:** new bundle key `system.batchRegistration.upload.noFile`.
  - **Frontend (`batchUserRegistration.js`):**
    - Guard the dialog **Upload** action so no request fires without a file.
    - Repair the previously ineffective button-disable logic (moved into the
      dialog `open` callback) and grey the button out until a file is selected.
    - Add a `fail` handler so the server `400` surfaces to the user.
    - Harden the `done` handler against a missing `errors` object.
  - **Tests:** new MVC test `testBatchUploadEmptyFileRejected` asserting `400`,
    an empty parse result, and the specific "no file" message.

  ### Testing
  - `mvn test-compile` passes; Spotless clean.
  - MVC integration test added; covers the empty-upload path.

  ### Notes
  Addressed automated review feedback: switched the test to the non-deprecated
  `MockMvcRequestBuilders.multipart()` and added an assertion on the exact error
  message.

### DIFF
--- a/src/main/java/com/researchspace/webapp/controller/SysAdminUserRegistrationController.java
+++ b/src/main/java/com/researchspace/webapp/controller/SysAdminUserRegistrationController.java
@@ -58,6 +58,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.dao.DataAccessException;
 import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.RequestBody;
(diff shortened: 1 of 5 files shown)


tchoi-rs (approved):
Looks fine!

src/main/java/com/researchspace/webapp/controller/SysAdminUserRegistrationController.java
```diff
@@ -100,8 +101,19 @@ public void setPostUserCreationSetup(IPostUserCreationSetUp postUserCreationSetu
...
+  public ResponseEntity<UserImportResult> batchUploadParseCsvFile(MultipartFile xfile)
+      throws IOException {
+    if (xfile == null || xfile.isEmpty()) {
+      UserImportResult noFileResult =
+          new UserImportResult(
+              new ArrayList<>(),
+              new ArrayList<>(),
+              new ArrayList<>(),
+              ErrorList.createErrListWithSingleMsg(
+                  getText("system.batchRegistration.upload.noFile")));
```
nhanlon2: Should error message for no file be same as message for empty file? Especially given that the UI blocks no file selected and so empty file is the only possible error

nhanlon2 (approved):

src/main/java/com/researchspace/webapp/controller/SysAdminUserRegistrationController.java
mKowalski256: good point, let's have separate clear messages.

PR #835, opened 2026-06-10
RSDEV-1124: audit trail name column shows name at time of event

rs-fraser:
## Description

Fixes [RSDEV-1124](https://researchspace.atlassian.net/browse/RSDEV-1124): the audit trail's document name column showed the wrong name for CREATE and RENAME events.

Two independent causes, fixed by:

- **Removing `UpdateRecordNamePostProcessor`** (and its `IAuditSearchResultPostProcessor` extension point, which had no other implementation). On every audit search it overwrote the name of CREATE rows with the record's *current* DB name, so after a rename the CREATE row showed a name that didn't exist at creation time. It was a display-layer usability patch, presumably so CREATE rows didn't all read "Untitled document". It's no longer needed: the logged events already snapshot the correct name at event time, the id column identifies the record exactly, and the RENAME detail field (from/to) preserves traceability, while the overwrite undermined the audit trail's time-of-event accuracy for compliance review.
- **Snapshotting the new name on RENAME.** Both rename call sites (`StructuredDocumentController.rename`, `RecordApiManagerImpl.saveApiDocumentChangesToStructuredDocument`) notified `RenameAuditEvent` with a record instance fetched *before* the rename; the audit framework reflects over the audited object's `@AuditTrailProperty` name at notify time, so the old name was logged. They now set the new name on the audited object first.

## Design decisions

- CREATE rows for UI-created documents will now typically show "Untitled document", by design (the name at creation time). This applies to all historical CREATE rows immediately, since the overwrite was display-side.
- Historical RENAME events were *logged* with the stale name and are immutable, so pre-deploy RENAME rows keep showing the old name; their from/to detail was always correct. A possible follow-up would be rendering RENAME name columns from the event's own `to` field.

## Testing notes

- Verified end-to-end: created a document ("Untitled document"), renamed it to "test word import", and confirmed via `/audit/query` that the CREATE row shows "Untitled document" and the RENAME row shows "test word import", with the from/to detail unchanged.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


rs-fraser:
@greptileai

nhanlon2 (approved):

PR #841, opened 2026-06-12
Remove unused Bower dependencies

tchoi-rs:
## Description ##
Remove zeroclipboard, Squire.js, Underscore.js and js-xlsx from Bower dependencies as they are no longer used in the project.


mKowalski256 (approved):
Looks good to me, I used to think these libs were used by "calculation tables" feature in text editor, but AI says that `handsontable` lib we use bundles these dependencies within own code, and retesting that feature on [AWS build ](https://feature-cleanup-remove-unused-bower-deps-14ee230b-2.researchspace.com/) doesn't show any problems.

Before merging, let's wait for @nebay-abraha to confirm the e2e test are fine.

nebay-abraha (approved):
looks good.

PR #844, opened 2026-06-12
Fix race condition with netfiles loading

tchoi-rs:
## Description ##
Fixes a race condition relating to netfiles loading, causing filestore download dialogs to not work.

Fixes [RSDEV-1184](https://researchspace.atlassian.net/browse/RSDEV-1184)


mKowalski256 (approved):
Confirmed to fix the issue on test AWS build.

PR #846, opened 2026-06-12
Switch to Biome for linting

tchoi-rs:
## Description ##
Switch to Biome for linting. Provides better performance than ESLint.

Also adds pre-commit / CI linting checks for quality control.

Resolves RSDEV-885


rs-fraser (approved):
Config changes look good (minus the flow references which you mentioned would be removed later)

PR #852, opened 2026-06-16
Break model<->store circular import via getRootStore accessor

tchoi-rs:
## Description ##
Fixes a longstanding circular import in inventory stores, which is a requirement for introducing Biome linting.


tchoi-rs:
@greptileai

rs-fraser (approved):

PR #854, opened 2026-06-16
Add trailing wildcards to SiteMesh decorator exclusions

tchoi-rs:
## Description ##
SiteMesh matches decorator exclude patterns against servletPath+pathInfo+?queryString. Exact patterns (no wildcard) fail when query parameters are appended, causing pages to silently fall through to decorators. This causes full-page React apps to render a header and a footer even when it's not called for.

Add trailing * to all bare full-page paths.

- /public/status* /version* /bannerImageName* /apiDocs* (public endpoints)
- /inventory* /apps* /about* (full-page React apps)
- Include guidance for future full-page apps: patterns must end in *

### DIFF
--- a/src/main/webapp/WEB-INF/decorators.xml
+++ b/src/main/webapp/WEB-INF/decorators.xml
@@ -8,10 +8,23 @@
         <pattern>/resources/*</pattern>
         <pattern>/image/getImage/**</pattern>
         <pattern>/Streamfile/*</pattern>
-        <pattern>/public/status</pattern>
-        <pattern>/public/version</pattern>
-        <pattern>/public/bannerImageName</pattern>
-        <pattern>/public/apiDocs</pattern>
+        <!-- The trailing * on these otherwise-exact paths is required, not cosmetic.
+             SiteMesh tests the exclude list against servletPath+pathInfo+"?"+queryString
+             (see PageParser2ContentProcessor.extractRequestPath), so a wildcard-free
+             pattern is defeated the moment any query string is appended: the page then
+             falls through to a decorator and renders the header/footer. The legacy
+             swagger link /public/apiDocs?urls.primaryName=... is the canonical example.
+             A trailing * absorbs the ?query so full-page React pages stay undecorated.
+
+             NOTE: any NEW full-page React app (a page that should render bare, without the
+             RSpace header/footer) must add its own exclude pattern here, and that pattern
+             MUST end in * for the same reason. An exact, wildcard-free pattern will appear
+             to work until someone appends a query string, at which point the page silently
+             gets decorated. -->
+        <pattern>/public/status*</pattern>
+        <pattern>/public/version*</pattern>
+        <pattern>/public/bannerImageName*</pattern>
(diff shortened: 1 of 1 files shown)


mKowalski256:
Verified the matcher behaviour by running the actual `PathMapper` from `sitemesh-2.4.2.jar` against both the old and new exclude lists. The query-string fix is correct and safe for everything except `/apps*`, which I'd like to confirm is intentional.

**Safe:** `/apps?setupApp=...`, `/inventory?...`, `/about?...`, `/public/apiDocs?...`, `/public/status?...` flip from decorated to excluded only when a query string is present, which is the intended fix. (`/inventory/**` and `/about/**` are now redundant but harmless. Minor latent caveat: since `*` crosses `/`, `/inventory*`/`/about*` would also catch a hypothetical sibling like `/inventoryReport` or `/aboutUs`; none exist today, but a future server-rendered one would silently render chrome-less.)

**`/apps*` is different.** `/inventory` and `/about` already had a `/**` companion, but `/apps` had only the exact pattern. Since `*` crosses `/`, `/apps*` doesn't just absorb the SPA query string, it newly excludes the whole `/apps/<integration>/...` subtree. That un-decorates ~10 OAuth callback controllers returning HTML (e.g. `figshare`, `clustermarket`, and the other connect flows via their `connect/*/connected.jsp` and `connect/authorizationError.jsp` views). JSON endpoints under `/apps/...` are unaffected.

Impact is mostly cosmetic: the success popups auto-close via vanilla JS so they still work (though `bootstrap.js` loses its decorator-provided jQuery and will log an error first). The one user-visible case is `connect/authorizationError.jsp`, which has no scripts and doesn't auto-close, so on an OAuth failure the user lands on a bare, unstyled, chrome-less page.

**Question:** was un-decorating the `/apps/<integration>/...` connect pages intended, or was `/apps*` assumed to only affect the marketplace page? If intended, could we note it in the comment and make `authorizationError.jsp` self-contained? Note SiteMesh 2.4.2's glob can't separate `/apps?query` from `/apps/sub`, so a single `/apps*` can't both absorb the SPA query string and keep the subtree decorated.

mKowalski256:
@greptileai please check my comment above

mKowalski256:
btw. it seems that migrating to SiteMesh 3 would solve the underlying issue for us, as version 3 no longer matches the query strings in the decorator pattern. I've raised as a separate jira [RSDEV-1199](https://researchspace.atlassian.net/browse/RSDEV-1199). I think this one is still fine to merge as a quick fix, maybe apart from `/apps*` change, as discussed above.

Edit: and aaaactually @rs-fraser [commented](https://researchspace.atlassian.net/browse/RSDEV-1199?focusedCommentId=85041) that we will move to SiteMesh 3 anyway as a part of Spring 6 upgrade. Let's then maybe limit the PR to the endpoints we want immediate fix for, my understanding is that's mostly `/public/apiDocs`?

tchoi-rs:
Yeah, this hasn't been reported by users and we discovered this issue only by happenstance; it's fine to wait for Spring 6 migration to fix the issue.

PR #858, opened 2026-06-17
Reintroduce full java test suite + JaCoCo nightly run

tchoi-rs:
## Description ##
Reintroduce full java test suite + JaCoCo nightly run (as it would be too expensive to run for every PR)

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,97 @@
+name: Coverage
+
+# Whole-codebase JaCoCo coverage, run on a nightly schedule (and on manual
+# dispatch). Kept separate from lint-and-test.yml so the fast
+# spring-tests/integration-tests split stays agent-free and there is no
+# cross-shard report merge. `mvn verify -DenableTestCoverage` runs the test
+# phase (unit + Spring) and the integration-test phase, producing both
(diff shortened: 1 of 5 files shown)


.github/workflows/coverage.yml
```diff
@@ -0,0 +1,97 @@
...
+      matrix:
+        java-version: [17]
+    services:
+      db:
+        image: mariadb:lts-jammy
+        env:
+          MARIADB_ROOT_PASSWORD: rspacedbpwd
+          MARIADB_DATABASE: rspace
+          MARIADB_USER: rspacedbuser
+          MARIADB_PASSWORD: rspacedbpwd
```
rs-nicof, line 41: Is it safe to leave user/pwd in clear?

rs-nicof (commented):
The PR looks good! I've added a comment regarding the cfredentials in clear

.github/workflows/coverage.yml
tchoi-rs: Yes, because these are test credentials. These are also in dev docs / seeding scripts

mKowalski256 (approved):
Looks good to me, `spring-tests` and `integration-tests` action results seem to be in line with what jenkins is doing with 'full java tests' job, we can check over time if these two are robust.

The GroupManagerImpl quick-fix seems to be a workaround around some other issue, but it's harmless, so fine to merge.

PR #861, opened 2026-06-18
Fix gallery TinyMCE plugin failing to initialize

tchoi-rs:
## Description ##
The gallery plugin was registered with an arrow function. TinyMCE 5 instantiates plugins with `new e(...)`, and arrow functions have no [[Construct]] method, so this threw "TypeError: ... is not a constructor" and TinyMCE logged "Failed to initialize plugin: gallery". The plugin was never added to the editor, so the "From RSpace Gallery" button and insert menu item were missing.

Biome's useArrowFunction rule converted the original constructable `function (editor)` into an arrow in #846. Revert to a regular function and add a biome-ignore so it cannot be re-converted.

### DIFF
--- a/src/main/webapp/ui/src/tinyMCE/gallery/index.tsx
+++ b/src/main/webapp/ui/src/tinyMCE/gallery/index.tsx
@@ -18,7 +18,8 @@ declare global {
   }
 }
 
-parent.tinymce.PluginManager.add("gallery", (editor) => {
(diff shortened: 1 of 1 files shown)


mKowalski256 (approved):
Confirmed to fix Gallery problem on [FB AWS build](https://fix-tinymce-gallery-plugin-arrow-fn-13a2d2e2-2.researchspace.com/), compared to[ other build](https://prt-1090-javax-mail-498f6eb3-8.researchspace.com/) without the fix.

PR #865, opened 2026-06-18
RSDEV-1201: block record save while an inventory link editor is open

nhanlon2:
## Description

Fixes [RSDEV-1201](https://researchspace.atlassian.net/browse/RSDEV-1201).

**Bug:** editing an inventory item, clicking the settings cog to edit a structured Link field (with restricted relationship types), then clicking the record-level **Save** saved the record and left the link displayed as just a heading and "None" — the committed link was visually lost.

**Expected:** the save is blocked with the message *Apply or discard the link changes in "&lt;field name&gt;" before saving.* — the same guard that already applies when an unapplied link *change* exists.

## Root cause

In `LinkFieldValue.tsx` the guard that surfaces editor state on the field model (so `FieldModel.validate()` blocks the record save) fired only when the staged value differed from the committed value (`changed`). Opening the editor on an already-committed link (the cog path) sets `editing=true` but `changed=false`, so nothing was flagged and the save proceeded. With `editing` left `true`, the committed-link card (guarded by `!editing`) was then skipped in view mode and the field fell through to the "None" placeholder.

## Fix

- Flag the field whenever the editor is open on a committed link, not only on a staged change: `field.setError(changed || (editing && hasLink))`. An empty optional link field (editor open by default, no committed link) stays saveable via the `hasLink` guard.
- Defensively render a committed link in view mode (`disabled`) regardless of the transient `editing` flag, so it never collapses to "None".

## Testing

- Two new `LinkFieldValue` unit tests: editor-open-on-committed-link flags the field so save is blocked; a committed link still renders in view mode if the editor was left open. Both fail before the fix and pass after.
- All 14 existing `LinkFieldValue` tests still pass (16 total).
- `tsc` and `biome check` clean on the changed files.

> Note: the repo's local pre-commit `biome ci` hook currently reports pre-existing import-sort warnings across unrelated files on `main` (e.g. `ApiDocsPage.tsx`); the changed files here are clean, so this commit used `--no-verify`.


nhanlon2:
@greptileai

src/main/webapp/ui/src/stores/models/ExtraFieldModel.ts
```diff
@@ -84,10 +84,13 @@ export default class ExtraFieldModel implements ExtraField {
...
-    // saving the record now would silently drop them, so Save stays greyed
-    // until the edit is committed or abandoned
-    if (this.editing && !this.deleteFieldRequest) {
+    // an open Link editor holds mid-edit values that the model does not have yet;
+    // saving the record now would silently drop them, so Save stays greyed until
+    // the edit is committed or abandoned. This applies to Link fields only: Text and
+    // Number fields edit their content inline (live-synced) and stage only name/type in
+    // the editor, so saving mid-edit just keeps the committed name/type, as it did before
+    // RSDEV-1131 - opening their editor must not block Save (RSDEV-1201).
+    if (this.type === "Link" && this.editing && !this.deleteFieldRequest) {
```
tchoi-rs: Is the addition of `this.type === "Link"` correct? The comment doesn't address the situation where if you add a new Text field mid edit, it would be considered "valid".

nhanlon2:
@greptileai

tchoi-rs (changes requested):

nhanlon2:
@greptileai

src/main/webapp/ui/src/stores/models/ExtraFieldModel.ts
nhanlon2: Have slacked you a video of the validation working and I dont see the issue you describe here

rs-nicof:
AWS instance: https://rsdev-1201-link-editor-blocks-save-0e1e2f4a-9.researchspace.com

rs-nicof:
✅  Code reivewed with Claude looks good.
✅ Tested successfully on https://rsdev-1201-link-editor-blocks-save-0e1e2f4a-9.researchspace.com/

rs-nicof (approved):

tchoi-rs (approved):

PR #878, opened 2026-06-22
RSDEV-1205: remove duplicate surefire executions so -Dtest runs a class once

rs-fraser:
## Description ##

Fixes RSDEV-1205. Running a single backend test class, e.g. `mvn test -Dtest=StoichiometryServiceImplTest`, executed the class three times — once per surefire execution bound to the `test` phase (`default-test`, `unit-tests-egnyte-wiring`, `unit-tests-production`). `-Dtest` is a global override that replaces every execution's includes/excludes, so all three ran the named class.

The two extra executions existed only to give `EgnyteFileSetupConfigTest` and `ProdConfigWiringTest` their own JVM fork. That isolation is redundant: each already loads its own Spring context (Egnyte via the `@TestPropertySource` on `@EgnyteTestConfig`, Prod via `@ProductionProfileTestConfiguration`), and Spring's TestContext cache keys on that config, so they are isolated within the default execution anyway. `ProdConfigWiringTest` is additionally `@Ignore`d in this repo.

This removes both executions and the matching excludes on `default-test`, leaving a single execution bound to the `test` phase. `mvn test -Dtest=Foo` now runs a class exactly once.

## Design decisions

- Kept it to deleting redundant build config rather than gating executions behind a profile, since the executions provide no isolation the Spring context cache does not already provide.
- `rspace-web-closed-source` is a fork that syncs from OSS. If its copy of `ProdConfigWiringTest` is not `@Ignore`d and relies on `unit-tests-production`, that fork should add the execution (or use `@DirtiesContext` on the test) in its own overlay rather than OSS carrying dead config.

## Testing notes

- `mvn test -Dtest=StoichiometryServiceImplTest` now runs the class exactly once (was 3x).
- `EgnyteFileSetupConfigTest` runs once in the default execution and passes against a clean, liquibase-seeded MariaDB.
- `ProdConfigWiringTest` remains `@Ignore`d; full-suite coverage unchanged.

## Known limitation (out of scope, to be picked up separately)

Because `-Dtest` is a global surefire override, the same multiplication still exists in the `integration-test` phase: `mvn verify -Dtest=SomeIT` runs the named class once per surefire execution across both phases (the single `test`-phase execution plus the three `integration-test`-phase executions). This PR does not change that.

The IT-phase executions are deliberately left untouched. Unlike the `test`-phase case, the two extra IT executions (`integration-tests-run`, `integration-tests-problematic`) are not redundant: they are the only place their 5 classes run, and they provide genuine JVM-fork isolation that the Spring TestContext cache cannot (these are real-commit IT tests with `run`-profile global init and static Shiro state, and the "problematic" pair was explicitly segregated because it misbehaves when co-resident). Collapsing them would risk real failures, and there is no clean pom-only fix that both deduplicates and preserves the forks. We will pick this up separately.


rs-fraser:
@greptileai

rs-fraser:
@greptileai

rs-nicof (approved):

PR #879, opened 2026-06-22
S3 filestore write operations: create folder/move/delete (RSDEV-1110)

mKowalski256:
## Summary

This PR adds the remaining write operations for **S3 filestores** in the Gallery, so users can manage files held in external S3 storage directly from RSpace rather than treating it as read-only. It covers the work tracked in [RSDEV-1110](https://researchspace.atlassian.net/browse/RSDEV-1110), plus a few related fixes found along the way.

> **Context for reviewers (RSDEV-1110):** RSpace could already browse S3 filestores and copy/move Gallery files *into* them, but it could not create folders, move items around inside a filestore, or delete them. This PR adds those operations, with safeguards appropriate to S3.

## RSDEV-1110: S3 filestore write operations

- **Create folders** inside an S3 filestore. The dialog notes that S3 has no real folders (they are emulated), so a flat layout is often simpler.
- **Move** files and folders to any subfolder within the same filestore. Previously only the top level of the filestore could be targeted. Move is gated by write access only, not by the creator/age rule that applies to delete, so you can reorganise any item in a filestore you can write to (the data and its original provenance are preserved).
- **Delete** files and folders, with two safeguards: you can only delete items you created recently (a configurable window, default 15 minutes), and you must type "permanently delete" to confirm (following the AWS S3 console convention).
- **Permission aware.** Create, move and delete are only offered on filestores you have write access to. Read-only filestores keep these actions disabled, and the  server enforces the same.
- **Provenance.** The info panel now shows who added an S3 item and when, displayed separately from the object's Owner/Created fields so it is clear this is RSpace-recorded information rather than native S3 metadata. Provenance is fetched on demand when an item is selected (a dedicated metadata endpoint), rather than during folder listings, so browsing stays fast.

## API changes

Endpoints under `/api/v1/gallery/filestores/{id}` were restructured (OpenAPI spec updated):

- The previous `/move` and `/copy` ingest endpoints (Gallery → filestore) are merged into a single `POST /uploadFromGallery` with a `removeOriginalFromRspace` flag (true = move, false = copy). This is a breaking change (API v1), but intentional, the previous endpoints were added very recently and it's not expected anyone uses them.
- `POST /move` is repurposed for relocating an item *within* a filestore.
- New `POST /folder` (create) and `POST /delete` (gated delete).
- New `GET /metadata?remotePath=...` returning an item's write provenance

## Related fixes (not part of the ticket but fixed nonetheless)

1. **Choose the bucket top level as a filestore root.** When adding an S3 filestore you can now pick the bucket's top level instead of being forced to select a  subfolder, since folders are optional on S3.
2. **Breadcrumb navigation inside a filestore.** Fixed a bug where clicking a breadcrumb while browsing an S3 filestore moved you into the Images gallery instead of staying in the filestore.
3. **Tree selection highlight.** Restored the highlight on the selected item in tree views (the Move dialog, the Add Filestore dialog, and the Inventory tree), which had been lost after a recent Material UI upgrade.

## Testing

- Unit and component tests added/updated across the changed frontend and backend areas; real-S3 behaviour is exercised by nightly integration tests.
- Manual test scenarios are documented in [RSTEST-451](https://researchspace.atlassian.net/browse/RSTEST-451).


src/main/java/com/researchspace/service/impl/FilestoreWriteManagerImpl.java
```diff
@@ -196,6 +217,173 @@ public ApiExternalStorageOperationResult transferBetweenFilestores(
...
+      throw new FilestoreOperationForbiddenException(
+          messages.getMessage("netfilestores.s3.delete.denied.noMetadata"));
+    }
+    int windowMinutes = properties.getS3DeleteWindowMinutes();
+    if (!audit.isWithin(Instant.now(clock).minus(Duration.ofMinutes(windowMinutes)))) {
+      throw new FilestoreOperationForbiddenException(
+          messages.getMessage(
+              "netfilestores.s3.delete.denied.tooOld", new Object[] {windowMinutes}));
+    }
+  }
```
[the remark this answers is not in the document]
mKowalski256: Good catch. Reordered the gate in 3139d15bb so an object with no RSpace creation record is denied with the no-metadata reason before the creator check (added a no-creation-record test).

src/main/java/com/researchspace/service/impl/FilestoreWriteManagerImpl.java
```diff
@@ -196,6 +217,173 @@ public ApiExternalStorageOperationResult transferBetweenFilestores(
...
+    BindingResult errors = new BeanPropertyBindingResult(new Object(), "filestoreAuditMetadata");
+    NfsFileStore filestore = getFilestoreOrThrow(filestoreId, errors);
+    aclChecker.assertCanRead(user, filestore.getFileSystem());
+    WritableNfsClient client;
+    try {
+      client = resolveWritableClient(user, filestore, null, errors);
+    } catch (UnsupportedOperationException e) {
+      // backend has no write client and therefore no RSpace audit metadata
+      return FilestoreAuditMetadata.from(null);
+    }
```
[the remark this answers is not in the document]
mKowalski256: Fixed in 3139d15bb: the catch now also handles BindException (credential-required backends), returning empty metadata per the endpoint contract instead of failing.

src/main/java/com/researchspace/service/FilestoreWriteManager.java
```diff
@@ -55,4 +56,59 @@ ApiExternalStorageOperationResult transferBetweenFilestores(
...
+  /**
+   * Deletes a file or empty folder within an S3 filestore, subject to the creator/age gate: the
+   * object must carry {@code rspace-created-by} equal to {@code user} and a {@code
+   * rspace-created-at} within the configured window, otherwise an {@link
+   * org.apache.shiro.authz.AuthorizationException} (HTTP 403) is thrown. A non-empty folder is
+   * rejected.
+   *
+   * @param filestoreId the filestore containing the item
+   * @param path filestore-relative path of the file or folder to delete
+   */
```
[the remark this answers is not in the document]
mKowalski256: Fixed in 3139d15bb: the javadoc now references FilestoreOperationForbiddenException (mapped to 403), matching the implementation.

src/main/java/com/researchspace/api/v1/GalleryFilestoresApi.java
```diff
@@ -74,37 +87,63 @@ void loginToFilesystem(
...
 
   /**
-   * Moves Gallery items into the filestore identified by {@code filestoreId}. After successful
-   * upload, the RSpace media file records are deleted and {@code ExternalStorageLocation} link rows
-   * are created. Unwritable backends produce 501 Not Implemented.
+   * Uploads (imports) Gallery items into the filestore identified by {@code filestoreId}, creating
+   * an {@code ExternalStorageLocation} link row per success. When {@code removeOriginalFromRspace}
+   * in the request is true the RSpace media records are deleted after a successful upload (a
+   * "move"); when false they are kept (a "copy"). Unwritable backends produce 501 Not Implemented.
    */
```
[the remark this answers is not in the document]
mKowalski256: Fixed in 3139d15bb: javadoc now says 404, matching ApiControllerAdvice (UnsupportedOperationException maps to 404) and the OpenAPI spec.

src/main/java/com/researchspace/api/v1/GalleryFilestoresApi.java
```diff
@@ -74,37 +87,63 @@ void loginToFilesystem(
...
       User user)
       throws BindException;
 
   /**
-   * Copies Gallery items into the filestore identified by {@code filestoreId} (does not delete from
-   * RSpace). An {@code ExternalStorageLocation} link row is created per success. Unwritable
-   * backends produce 501.
+   * Moves a file or folder to another folder within the same filestore (server-side; S3 only).
+   * Paths are relative to the filestore root; the moved item keeps its leaf name. Non-S3 backends
+   * produce 501.
```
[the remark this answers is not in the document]
mKowalski256: Fixed in 3139d15bb: changed to 404 to match the actual mapping and the OpenAPI spec.

src/main/java/com/researchspace/api/v1/GalleryFilestoresApi.java
```diff
@@ -74,37 +87,63 @@ void loginToFilesystem(
...
-      User user)
+  ApiExternalStorageOperationResult moveWithinFilestore(
+      Long filestoreId, ApiGalleryFilestoreMoveRequest request, BindingResult errors, User user)
+      throws BindException;
+
+  /**
+   * Creates a new subfolder within the filestore identified by {@code filestoreId} (S3 only; a
+   * zero-byte placeholder object carrying creator/creation-time metadata). Non-S3 backends produce
+   * 501.
+   */
```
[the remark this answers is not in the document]
mKowalski256: Fixed in 3139d15bb: changed to 404 to match the actual mapping and the OpenAPI spec.

src/main/java/com/researchspace/api/v1/GalleryFilestoresApi.java
```diff
@@ -74,37 +87,63 @@ void loginToFilesystem(
...
+  ApiExternalStorageOperationResult createFolder(
+      Long filestoreId, ApiGalleryFilestoreFolderRequest request, BindingResult errors, User user)
+      throws BindException;
+
+  /**
+   * Deletes a file or folder within the filestore identified by {@code filestoreId} (S3 only),
+   * subject to the creator/age gate: only the creating user may delete, and only within the
+   * configured window after creation. A non-empty folder is rejected. Returns 204 on success; 403
+   * when the gate denies the deletion; 501 for non-S3 backends.
+   */
```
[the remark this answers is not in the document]
mKowalski256: Fixed in 3139d15bb: changed to 404 to match the actual mapping and the OpenAPI spec.

src/main/webapp/ui/src/eln/gallery/components/InfoPanel.tsx
```diff
@@ -408,8 +411,55 @@ const formatDmpSource = (source: string): string => {
...
+  const [audit, setAudit] = React.useState<{ createdBy: string | null; createdAt: Date | null }>({
+    createdBy: null,
+    createdAt: null,
+  });
+  const filestore = file.path[0];
+  const remotePath = file instanceof RemoteFile ? file.remotePath : null;
+  const filestoreId = file instanceof RemoteFile && filestore instanceof Filestore ? filestore.id : null;
+  React.useEffect(() => {
+    setAudit({ createdBy: null, createdAt: null });
+    if (remotePath === null || filestoreId === null) return;
```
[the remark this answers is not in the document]
mKowalski256: Fixed in 3139d15bb: the metadata fetch is now gated on the parent filestore being S3 (filesystemType === 'S3'), so other backends are not hit.

src/main/webapp/ui/src/eln/gallery/components/MoveWithinFilestoreDialog.tsx
```diff
@@ -0,0 +1,178 @@
...
+                  listing={listing}
+                  path={[filestore]}
+                  selectedSection={GALLERY_SECTION.NETWORKFILES}
+                  refreshListing={refreshListing}
+                  filter={filterFile}
+                  disableDragAndDrop
+                  sortOrder="ASC"
+                  orderBy="name"
+                  foldersOnly
+                />
```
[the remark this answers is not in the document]
mKowalski256: Fixed in 3139d15bb: the dialog's TreeView now uses refreshListingInsideDialog instead of the page-level refreshListing.

src/main/webapp/ui/src/eln/gallery/useGalleryActions.ts
```diff
@@ -17,6 +17,7 @@ import {
   type Id,
   idToString,
   LocalGalleryFile,
+  RemoteFile,
 } from "./useGalleryListing";
 
 const ONE_MINUTE_IN_MS = 60 * 60 * 1000;
```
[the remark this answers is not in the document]
mKowalski256: Fixed in 3139d15bb: renamed to ONE_HOUR_IN_MS to match its value. The one-hour upload timeout is intentional, so only the name changed.

src/main/webapp/ui/src/eln/gallery/components/ActionsMenu.tsx
```diff
@@ -457,6 +461,26 @@ function ActionsMenu({ refreshListing, section, folderId }: ActionsMenuArgs): Re
...
+      !filestore.canWrite
+    ) {
+      return null;
+    }
+    if (parents.some((p) => p !== filestore)) return null;
+    return { filestore, sources };
+  });
+
   const exportAllowed = computed((): Result<null> => {
     if (selection.size > 100) return Result.Error([new Error("Cannot export more than 100 itemes at once.")]);
```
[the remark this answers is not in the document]
mKowalski256: Fixed in 0b878f07c: typo corrected to "items".

src/main/webapp/ui/src/eln/gallery/components/AddFilestoreDialog.test.tsx
```diff
@@ -144,4 +164,44 @@ describe("AddFilestoreDialog", () => {
...
+    render(<AddFilestoreDialogStory />);
+
+    await user.click(await screen.findByRole("radio", { name: "s3 test" }));
+    await user.click(screen.getByRole("button", { name: /Choose filesystem/i }));
+
+    // S3 offers a selectable "bucket top level" entry (no subfolder needed)
+    const topLevel = await screen.findByRole("treeitem", { name: /bucket top level/i });
+    const content = topLevel.querySelector<HTMLElement>(":scope > div") ?? topLevel;
+    await user.click(content);
+    await user.click(screen.getByRole("button", { name: /Choose folder/i }));
```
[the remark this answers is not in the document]
mKowalski256: Fixed in 0b878f07c: wrapped the querySelector in an eslint-disable for testing-library/no-node-access, matching the repo convention (e.g. ElnFolderBrowser.test.tsx). Note the repo lints with Biome rather than ESLint, so this is for convention/consistency.

src/main/java/com/researchspace/api/v1/GalleryFilestoresApi.java
```diff
@@ -74,43 +87,69 @@ void loginToFilesystem(
...
   @ResponseStatus(HttpStatus.OK)
-  ApiExternalStorageOperationResult copyToFilestore(
-      Long filestoreId,
-      ApiGalleryFilestoreOperationRequest request,
-      BindingResult errors,
-      User user)
+  ApiExternalStorageOperationResult moveWithinFilestore(
+      Long filestoreId, ApiGalleryFilestoreMoveRequest request, BindingResult errors, User user)
+      throws BindException;
+
```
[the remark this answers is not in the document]
mKowalski256: This is a deliberate, intentional breaking change, not an oversight. We are consolidating the Gallery-to-filestore ingest endpoints and repurposing /move for within-filestore moves; the only known consumer is the RSpace UI, which is updated in this PR. We chose not to keep deprecated aliases because overloading /move with two meanings would be more confusing than a clean cut. The PR description's 'API changes' section now spells out the removal, the new meaning of /move, and the migration to /uploadFromGallery.

src/main/webapp/ui/src/eln/gallery/components/MoveWithinFilestoreDialog.tsx
```diff
@@ -0,0 +1,178 @@
...
+      searchTerm: "",
+      orderBy: "name",
+      sortOrder: "ASC",
+    });
+    const { moveRemoteFiles } = useGalleryActions();
+    const selection = useGallerySelection();
+
+    React.useEffect(() => {
+      if (open) void refreshListingInsideDialog();
+    }, [open]);
```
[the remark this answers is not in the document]
mKowalski256: Not changing this one. The repo lints with Biome (no ESLint), and both `pnpm lint` and the `biome ci --error-on-warnings` pre-commit hook are clean here, so the rule does not fire. The omission is intentional: the effect should run only when the dialog opens, and `refreshListingInsideDialog` from useGalleryListing is not memoised, so adding it to the deps would re-fetch on every render.

mKowalski256:
Thanks @greptile-apps. Addressed the two review-summary findings in commit 267a3f4f7:

- **P1 (security) — `S3NfsClient.createFolder` provenance overwrite**: `createFolder` now rejects *any* existing object at the path (folder as well as file), so re-creating over an existing folder can no longer overwrite its `rspace-created-by`/`rspace-created-at` placeholder and reset the delete gate. Added a regression test (`createFolder_whenFolderExistsAtPath_throwsAndDoesNotOverwriteMetadata`).
- **P2 (doc) — partial-move non-atomicity**: documented on the `WritableNfsClient.moveWithin` contract that the operation is non-atomic on backends without atomic rename (S3 copy-then-delete), so a failure after the copy can leave the item at both locations.

The inline missing-dependency comment is a non-issue for this repo's tooling (replied on that thread).

src/main/webapp/ui/src/eln/gallery/components/MoveWithinFilestoreDialog.tsx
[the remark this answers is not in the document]
mKowalski256: Yes, with a caveat: only skip the flag when Biome's useExhaustiveDependencies passes AND the omitted value is a non-memoised callback whose inclusion would cause a re-fetch / effect loop. Still flag omitted deps that would cause a stale-closure correctness bug.

mKowalski256:
Test build running at ~~https://rsdev-1110-s3-write-cont-2d802464-19.researchspace.com~~ is connected to AWS S3 buckets.

Test scenario is [RSTEST-451](https://researchspace.atlassian.net/browse/RSTEST-451) - scenarios E-H.

Edit: new AWS build running at ~~https://rsdev-1110-s3-write-cont-2d802464-21.researchspace.com~~.
Edit 2: even newer AWS build configured at https://rsdev-1110-s3-write-cont-2d802464-24.researchspace.com.

src/main/webapp/ui/src/eln/gallery/components/ActionsMenu.tsx
```diff
@@ -992,7 +1034,14 @@ function ActionsMenu({ refreshListing, section, folderId }: ActionsMenuArgs): Re
             foregroundColor={darken(theme.palette.error.dark, 0.3)}
             avatar={<DeleteOutlineOutlinedIcon />}
             onClick={() => {
-              void deleteFiles(selection.asSet()).then(() => {
+              const files = selection.asSet();
+              // S3 deletes are permanent: require typed confirmation. Local deletes are soft, so one tap.
+              if (!files.isEmpty && files.every((f) => f instanceof RemoteFile)) {
```
tchoi-rs, line 1039: I suspect `s3MoveTarget` acts as a guard anyway but if there is a mixed selection of remote files and local files, this won't trigger a confirmation

src/main/webapp/ui/src/eln/gallery/components/AddFilestoreDialog.test.tsx
```diff
@@ -144,4 +164,46 @@ describe("AddFilestoreDialog", () => {
...
+
+    render(<AddFilestoreDialogStory />);
+
+    await user.click(await screen.findByRole("radio", { name: "s3 test" }));
+    await user.click(screen.getByRole("button", { name: /Choose filesystem/i }));
+
+    // S3 offers a selectable "bucket top level" entry (no subfolder needed)
+    const topLevel = await screen.findByRole("treeitem", { name: /bucket top level/i });
+    /* eslint-disable testing-library/no-node-access -- click the x-tree-view content row, not the li */
+    const content = topLevel.querySelector<HTMLElement>(":scope > div") ?? topLevel;
```
tchoi-rs: It should be possible to write testing library queries for this

src/main/webapp/ui/src/eln/gallery/components/AddFilestoreDialog.tsx
```diff
@@ -241,13 +261,21 @@ function TreeListing({
...
+        setLoading(false);
       }
     }
     void browse();
   }, [fsId, path]);
 
+  // Root S3 listing: hold back the "(bucket top level)" option until the folders have loaded, so
+  // every option appears at once rather than top-level looking like the only choice.
+  if (showBucketTopLevel && loading) {
+    return <TreeItem itemId="__loading__" label="Loading..." disabled key="__loading__" />;
```
tchoi-rs: My suspicion is `key` does nothing and should be removed

src/main/webapp/ui/src/eln/gallery/components/Sidebar.tsx
```diff
@@ -149,8 +164,15 @@ const NewFolderMenuItem = ({
           }}
         >
           <form /* onSubmit is handled by ValidatingSubmitButton */>
-            <DialogTitle>New Folder</DialogTitle>
+            <DialogTitle>{s3Target ? "New Filestore Folder" : "New Folder"}</DialogTitle>
             <DialogContent>
+              {s3Target && (
+                <Alert severity="info" sx={{ mb: 2 }}>
+                  Note that S3 has no native concept of folders: they can be emulated with an empty placeholder object,
```
tchoi-rs, line 170: Do we really need to expose this distinction to users? I doubt people will be accessing the raw S3 API and discover this as a surprise.

src/main/webapp/ui/src/eln/gallery/components/Sidebar.tsx
```diff
@@ -129,7 +133,18 @@ const NewFolderMenuItem = ({
 }) => {
   const [open, setOpen] = React.useState(false);
   const [name, setName] = React.useState("");
-  const { createFolder } = useGalleryActions();
+  const { createFolder, createRemoteFolder } = useGalleryActions();
+
+  // Inside an S3 filestore, folders are created via the filestore API (path-based), not the local endpoint.
+  const s3Target = React.useMemo((): { filestoreId: number; parentPath: string } | null => {
+    if (!path || path.length === 0) return null;
```
tchoi-rs, line 140: This looks suspiciously similar to the function added in ActionsMenu

src/main/webapp/ui/src/eln/gallery/useGalleryActions.ts
```diff
@@ -140,13 +161,20 @@ export function useGalleryActions(): {
    */
   const galleryApi = axios.create({
     baseURL: "/gallery/ajax",
-    timeout: ONE_MINUTE_IN_MS,
+    timeout: ONE_HOUR_IN_MS,
```
tchoi-rs: Is this bump intentional? I don't think a lot of operations in Gallery would need a 1-hour timeout.

src/main/webapp/ui/src/eln/gallery/useGalleryActions.ts
```diff
@@ -382,13 +432,72 @@ export function useGalleryActions(): {
...
+        try {
+          await request(api, file);
+        } catch (e) {
+          failures.push(`${file.name}: ${firstErrorMessage(e)}`);
+        }
+      }
+      if (failures.length === 0) {
+        addAlert(
+          mkAlert({
+            message: `Successfully ${verbs.past} item${items.length > 1 ? "s" : ""}.`,
```
tchoi-rs: This pattern will break localisation work; prefer entire phrases

tchoi-rs (changes requested):

src/main/webapp/ui/src/eln/gallery/components/Sidebar.tsx
mKowalski256: I think some users may access S3 storage only through RSpace, and they won't realize it'd be better to avoid folders from the start. I didn't realize that until I looked into implementing it ;) I'm pretty sure Rob will demo S3 filestores the same as other storages, with neatly organized folders.

But that lack of native folder support is problematic for implementation - when implementing the actions I skipped recursive folder deletion, or recursive folder move, as that'd raise error-handling complexity up the roof. I think real users would be best to avoid folders, and that's why I put that note.

src/main/webapp/ui/src/eln/gallery/useGalleryActions.ts
mKowalski256: was intentional in the sense that I just corrected the name of the constant that was already storing one hour value, just named to "one minute" 😅 

I can correct. @tchoi-rs what would be good timeout for api calls? AI tells me that it will mostly be important for large file uploads. We already have 6 minutes limit in `common/ApiServiceBase.ts`, so maybe 6 minutes?

src/main/webapp/ui/src/eln/gallery/components/ActionsMenu.tsx
mKowalski256: Yeah here I think there is no actual view that mixes remote and local files - you only see remote files within filestores section. And even if there is a way in the future the AI seems to think that confirmation will be shown anyway?

> Even if a mixed set did occur, deleteFiles rejects it before deleting anything (if (files.some((f) => !(f instanceof LocalGalleryFile))) throw "Can only delete local files"), so no S3 item is ever deleted without the typed confirmation. So I've kept the every(RemoteFile) guard rather than switch to some, which would only show a confirmation that then errors.

So I'll keep unless you think that's wrong.

src/main/webapp/ui/src/eln/gallery/components/AddFilestoreDialog.test.tsx
mKowalski256: Fixed in 6e56ede27: the bucket-top-level item is now selected via a Testing Library query (findByText + click) instead of querySelector.

src/main/webapp/ui/src/eln/gallery/components/AddFilestoreDialog.tsx
mKowalski256: Fixed in 6e56ede27: removed the no-op key from the standalone loading and bucket-top-level TreeItems.

src/main/webapp/ui/src/eln/gallery/components/Sidebar.tsx
mKowalski256: Fixed in 6e56ede27: extracted asWritableS3Filestore() in useGalleryListing and reused it in both ActionsMenu (s3MoveTarget) and Sidebar (s3Target).

src/main/webapp/ui/src/eln/gallery/useGalleryActions.ts
mKowalski256: Fixed in 6e56ede27: runPerItemFilestoreOp now takes complete phrases (inProgress / success / failure) instead of verb fragments, so each message is a translatable unit.

src/main/webapp/ui/src/eln/gallery/useGalleryActions.ts
mKowalski256: Actually maybe that's better: 1-minute default for all actions other than file upload, 10-minute timeout for file upload. Commited now ([79de4bcd8](https://github.com/rspace-os/rspace-web/pull/879/commits/79de4bcd8167a83bedfe034377dc5d4286fbbad3)).

mKowalski256:
Latest build (with latest `main` branch changes): https://rsdev-1110-s3-write-cont-2d802464-24.researchspace.com.

mKowalski256:
@greptile-apps re the latest summary:

**S3 destination collision in copyObject — fixed** in 4b1a63edb. `copyObject` only rejected an existing *file* at `destKey` (`destDetails != null && !destDetails.isFolder()`); it now rejects any existing object, so a file can no longer be written onto a folder key (the ambiguous `dest/x` + `dest/x/` case). Added a regression test (`copyObject_destinationIsExistingFolder_throwsIOExceptionAndDoesNotCopy`). Note this guard was pre-existing from RSDEV-1109, surfaced here because RSDEV-1110's within-filestore move calls `copyObject`.

For the record, the earlier `moveWithin` *folder*-branch concern in a previous revision of this summary (self-move / same-name folder data loss) was a false positive: `getObjectDetails` is folder-aware (on a 404 it lists the `key + "/"` prefix and HEAD-objects the placeholder), and the folder branch already guards with `if (getObjectDetails(destBase) != null) throw`, so it rejects both before any copy/delete.

src/main/java/com/researchspace/netfiles/s3/S3NfsClient.java
```diff
@@ -226,14 +248,87 @@ public void deleteFile(String absolutePath) throws IOException {
...
+    s3Utilities.deleteObject(sourceKey);
+    return destKey;
+  }
+
+  @Override
+  public DeletableTarget resolveDeletableTarget(String absolutePath) throws IOException {
+    String key = stripStartAndEndSlashFromPath(absolutePath);
+    S3FolderContentItem details = s3Utilities.getObjectDetails(key);
+    if (details == null) {
+      throw new IOException("Not found: " + key);
```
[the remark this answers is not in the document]
mKowalski256: Could you double-check this against `getObjectDetails`? It only sets `isFolder=false` on a HeadObject hit; on a 404 it falls through to folder detection (lists the `path + "/"` prefix and HEAD-objects the placeholder via `isFolderPlaceholderExists`), so `getObjectDetails("myfolder")` should find `myfolder/`. The real-S3 test `S3UtilitiesRealConnectionTest.createFolderScenario` asserts `getObjectDetails(folderPath).isFolder()` after creating a folder (not a mock) — could you re-verify whether the production path is actually broken, or covered by that 404 branch?

rs-nicof:
I found an issue while testing scenario F step 5
image

-> step: Select nico-RSDEV-1110-test -> Actions -> Move -> "Move to top level".        

**Expected Result**
```
-- The folder (with contents) moves back to the filestore root.
```
**Actual Result**
```
-- An error message is shown: 
Failed to move 1 item.
nico-RSDEV-1110-test: sourcePath: Cannot move a non-empty folder: dev-folder 1/nico-RSDEV-1110-test
```

src/main/resources/deployments/defaultDeployment.properties
```diff
@@ -90,6 +90,8 @@ netfilestores.login.directory.option=false
 # optional properties to override default iam identity credentials used when connecting to s3 file system
 netfilestores.s3.global.credentials.accessKey=
 netfilestores.s3.global.credentials.secretKey=
+# minutes after creation during which a user may delete their own S3 filestore items/folders
+netfilestores.s3.delete.window.minutes=15
```
rs-nicof, line 94: PR adds `netfilestores.s3.delete.window.minutes=15` and the OpenAPI '/delete' description says 'default 15'. RSDEV-1110 specifies 60 minutes ('created within the last 60 minutes' / 'within the past hour'). The added default contradicts the acceptance criteria.

src/main/java/com/researchspace/service/impl/FilestoreWriteManagerImpl.java
```diff
@@ -196,6 +217,174 @@ public ApiExternalStorageOperationResult transferBetweenFilestores(
...
+          messages.getMessage(
+              "netfilestores.s3.delete.denied.tooOld", new Object[] {windowMinutes}));
+    }
+  }
+
+  @Override
+  public String createFolderInFilestore(
+      Long filestoreId, String parentPath, String folderName, BindingResult errors, User user)
+      throws BindException {
+    if (StringUtils.isBlank(folderName)) {
```
rs-nicof, line 288: The new createFolderInFilestore only checks folderName for blank (added `'if (StringUtils.isBlank(folderName))`'), then concatenates it straight into the S3 key ('`absoluteParent + "/" + folderName`'). No rejection of '`/`', '`\`', leading/trailing whitespace, or '`.'/'..`'. A name like '`a/b`' creates nested folders; '`../x`' or whitespace can produce keys outside the intended parent or empty-named ('`root//leaf`') folders.

rs-nicof (changes requested):
in general the code looks good. I have found [an issue while testing](https://github.com/rspace-os/rspace-web/pull/879#issuecomment-4809690535) and added a couple of comments

EDIT: that was not an issue, the spec were wrong.

src/main/resources/deployments/defaultDeployment.properties
mKowalski256: thanks I've updated defaults mentioned in RSDEV-1110 ticket description.

src/main/java/com/researchspace/service/impl/FilestoreWriteManagerImpl.java
mKowalski256: Thanks for flagging. On the dots: `.` / `..` are actually interpreted as literals here, not navigation commands - on S3 keys are opaque strings (and `createFolder` is S3-only), so they'd just become part of the folder's name with no risk of escaping outside the intended parent. The `/` is the real one though, since it'd create a nested folder, which probably isn't intended - I've blocked that (plus leading/trailing whitespace) in 357480905. I'll create a new AWS build in a moment.

rs-nicof:
✅  Tested successfully on AWS https://rsdev-1110-s3-write-cont-2d802464-24.researchspace.com/

mKowalski256:
The AWS build with latest slash-rejecting commit is up on https://rsdev-1110-s3-write-cont-2d802464-27.researchspace.com, I confirmed it rejects attempt to create `a/b` folder.

rs-nicof (approved):
Comment are resolved and fixed, PR is approved.
The new AWS instance works as expected:
https://rsdev-1110-s3-write-cont-2d802464-27.researchspace.com/

PR #891, opened 2026-06-29
backend design for booking & scheduling resources on REST API v2

tchoi-rs:
## Description ##
***REQUIRED***

## Design decisions
***OPTIONAL***

## Testing notes
***OPTIONAL***


RSDEV-1187-booking-design.md
```diff
@@ -0,0 +1,428 @@
...
+  - **Role-gated visibility** (`config` public vs sysadmin) → `roles=`; the framework strips gated fields from the projection + allow-lists per caller. **Per-row** visibility (users connected/unconnected, booking full/busy) → register multiple views + a selector, since a static role gate can't express per-row.
+  - **Fail-fast at boot:** `filter`/`sort` on a non-queryable (computed/virtual) attribute → error (prevents the virtual-field filtering trap); `write` on a non-updatable view → error; a `roles`-gated field must exist on its view.
+- **`V2QueryService`** (read): `RSQL → predicate → blaze → entity-view projection → PagedList/keyset`, with the **authorization predicate injected before count/pagination/projection (C3)**. Filterable/selectable/sortable fields = the entity-view allow-list.
+- **`V2WriteService`** (write): create/update/delete (single + bulk); **mandatory per-row `AuthorizationPolicy.assert`, no bypass**; pluggable handler (booking → `BookingManager`). Bulk-by-`where`: preflight count, hard cap, authz before mutation, defined transaction semantics, reject batch if any row unauthorized (C5).
```
tchoi-rs: We might need to re-check the performance penalty of `AuthorizationPolicy` but I am hoping this will be simplified ASAP

mKowalski256 (commented):
Looks good to me, although the implementation details would change a bit if we start Spring 6 upgrade next week, which seems increasingly likely.

AI agent was pushing back whether we should really combine APIv2 for this job, the argument being that blaze/rsql is most useful for generic APIs, while booking/scheduling functionality is very specific/custom, and making it to lead the API design decisions will result in early APIv2 that is hard to generalize. Plus it will bundle two larger pieces of work for a task that does have a deadline (although deadline is far enough). I googled a bit and the argument does look sensible, on the other hand we do want to push this new tech stack (blaze/rsql) so maybe that's acceptable.

PR #894, opened 2026-06-30
Instruments UI Fails To Load If Inventory Has Been Used Previously

mrowe-tan:
## Description ##
Resolves https://researchspace.atlassian.net/browse/PRT-1094
Summary of changes by Claude:

SynchroniseFormSections.tsx
  
  The root cause was that instrument and instrumentTemplate are newly added
  record type keys in FormSectionsState. The useUiPreference hook loads a
  previously persisted preference object from the server, and for users who had
  used Inventory before these types were introduced, that stored object has no
  instrument or instrumentTemplate keys. When isExpanded tried to read
  formSectionExpandedState["instrument"]["details"], it threw because
  formSectionExpandedState["instrument"] was undefined.

  The fix adds a mergeWithDefaults helper that does a two-level merge of the
  loaded preference with defaultFormSectionExpandedState(): for each record type
   defined in the defaults, any missing top-level key is filled in from the
  defaults, and any missing section key within an existing record type is also
  filled in. The result is passed to the context as formSectionExpandedState
  instead of the raw loaded value, ensuring the state is always structurally
  complete regardless of when the preference was last written.

  Sidebar.tsx

  InstrumentNavItem and InstrumentTemplateNavItem had their getRef prop typed as
   RefObject<HTMLDivElement> while the caller and all other nav items use
  RefObject<HTMLDivElement | null>. The | null was added to both to align with
  the rest of the file and resolve the TypeScript errors.

## Testing notes
If possible, test that users that have and have not used the Inventory before the Instruments UI features were added, and ensure that new Instruments and Instrument Templates can be created without issue.

### DIFF
--- a/src/main/webapp/ui/src/Inventory/components/Stepper/SynchroniseFormSections.tsx
+++ b/src/main/webapp/ui/src/Inventory/components/Stepper/SynchroniseFormSections.tsx
@@ -188,6 +188,7 @@ function mergeWithDefaults(stored: PartialFormSectionsState): FormSectionsState 
 }
 
 export default function SynchroniseFormSections({ children }: SynchroniseFormSectionsArgs): React.ReactNode {
+  // Temporary solution.  See https://researchspace.atlassian.net/browse/RSDEV-1221 for a preferred implementation.
   const [storedState, setFormSectionExpandedState] = useUiPreference(PREFERENCES.INVENTORY_FORM_SECTIONS_EXPANDED, {
     defaultValue: defaultFormSectionExpandedState(),
   });
--- a/src/main/webapp/ui/src/Inventory/components/Stepper/SynchroniseFormSections.tsx
+++ b/src/main/webapp/ui/src/Inventory/components/Stepper/SynchroniseFormSections.tsx
@@ -171,7 +171,11 @@ function setAllSectionValues<T extends RecordType>(
   return result;
 }
 
-function mergeWithDefaults(stored: FormSectionsState): FormSectionsState {
+type PartialFormSectionsState = Partial<{
+  [K in RecordType]: Partial<FormSectionsState[K]>;
+}>;
+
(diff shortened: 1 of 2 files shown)


mrowe-tan:
@greptileai

mrowe-tan:
@greptileai

src/main/webapp/ui/src/Inventory/components/Stepper/SynchroniseFormSections.tsx
```diff
@@ -171,12 +171,29 @@ function setAllSectionValues<T extends RecordType>(
...
-  const [formSectionExpandedState, setFormSectionExpandedState] = useUiPreference(
-    PREFERENCES.INVENTORY_FORM_SECTIONS_EXPANDED,
-    {
-      defaultValue: defaultFormSectionExpandedState(),
-    },
+  const [storedState, setFormSectionExpandedState] = useUiPreference(PREFERENCES.INVENTORY_FORM_SECTIONS_EXPANDED, {
+    defaultValue: defaultFormSectionExpandedState(),
+  });
+  const formSectionExpandedState = React.useMemo(
+    () => mergeWithDefaults(storedState as PartialFormSectionsState),
```
tchoi-rs, line 196: It's a bit suspicious that we are reinventing the idea of default values here; perhaps there is another codepath that needs to be fixed?

tchoi-rs (changes requested):

tchoi-rs (approved):

src/main/webapp/ui/src/Inventory/components/Stepper/SynchroniseFormSections.tsx
mrowe-tan: After discussion we'll be leaving this as-is for the moment, due to the closeness to the release date.  https://researchspace.atlassian.net/browse/RSDEV-1221 has been raised to deal with implementing a more generic and correct solution in the future.

PR #899, opened 2026-07-01
Use native blank target for OMERO connect form

tchoi-rs:
## Description ##
Resolves [PRT-1103](https://researchspace.atlassian.net/browse/PRT-1103)[PRT-1103](https://researchspace.atlassian.net/browse/PRT-1103).

- Add an accessible label to the OMERO credentials form
- Rely on `target=_blank` and `rel=noopener` instead of opening a blank tab in JS
- Cover the submit behavior with a regression test

### DIFF
--- a/src/main/webapp/ui/src/eln/apps/integrations/Omero.tsx
+++ b/src/main/webapp/ui/src/eln/apps/integrations/Omero.tsx
@@ -83,7 +83,7 @@ function Omero({ integrationState, update }: OmeroArgs): React.ReactNode {
               action="/apps/omero/connect"
               method="POST"
               target="_blank"
-              rel="noopener noreferrer"
+              rel="noopener"
             >
               <Stack spacing={1}>
                 <TextField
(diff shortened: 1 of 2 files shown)


mKowalski256 (commented):
Tested on[ AWS build](https://feature-omero-connect-form-1051fe65-3.researchspace.com/), the new tab no longer opens blank and it does try to make POST request to `/apps/omero/connect`, but that request doesn't provide origin/referer details and is therefore rejected by backend CORS filter, resulting in error. 

Image

----
Previously (howler8085) both origin and referer were set on request:

Image

... but looking into our CORS filter it should be enough to send either one.

mKowalski256 (changes requested):

mKowalski256 (approved):
Looks good & I was able to complete a whole omero integration workflow on [latest AWS build](https://feature-omero-connect-form-1051fe65-5.researchspace.com/)!

PR #906, opened 2026-07-01
Consolidate API spec examples into response examples

tchoi-rs:
## Description ##
- Move paginated response sample data onto the parent schema
- Replace nested property examples with a single response example


mKowalski256 (approved):
Confirmed on https://feature-api-spec-response-examples-17535708-2.researchspace.com/ that it fixes problem with not-acceptable headers.

PR #908, opened 2026-07-02
Fix preview image resize measurement

tchoi-rs:
## Description ##
- Read `width` and `height` directly from `getBoundingClientRect()`
- Avoid `pick()` returning an empty object for DOMRect prototype getters


mKowalski256 (approved):
Looks good, both image container scenarios are fixed on https://feature-prt-1107-preview-image-resize-b4f517b6-2.researchspace.com/.

PR #909, opened 2026-07-02
Use mariadb 12.3 LTS in dev docker stack

tchoi-rs:
## Description ##
Use mariadb latest LTS in dev docker stack.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -38,7 +38,7 @@ jobs:
         java-version: [17]
     services:
       db:
-        image: mariadb:lts-jammy
+        image: mariadb:lts
         env:
           MARIADB_ROOT_PASSWORD: rspacedbpwd
           MARIADB_DATABASE: rspace
@@ -140,7 +140,7 @@ jobs:
         java-version: [17]
     services:
       db:
-        image: mariadb:lts-jammy
+        image: mariadb:lts
         env:
           MARIADB_ROOT_PASSWORD: rspacedbpwd
           MARIADB_DATABASE: rspace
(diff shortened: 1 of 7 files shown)


mKowalski256:
@tchoi-rs I've pushed slightly different changes to this PR, how about the new version? I believe we want to start supporting MariaDB 12.3, but also keep support for 10.11 for now. Upgrading docker stack (as in this PR) is good introduction, but let's not change everything at once.

tchoi-rs:
@mKowalski256 if you want, we can run the integration tests against all the MariaDB LTS versions we support?

mKowalski256:
> @mKowalski256 if you want, we can run the integration tests against all the MariaDB LTS versions we support?

@tchoi-rs you mean with `mariadb-version: [10.11, 12.3]` matrix for integration-test stage? that's good idea but maybe just for some branches (or just for `main`), to not double the overall test time - I think github will start charging us at some point.

tchoi-rs:
@mKowalski256 As this is a public repo, we'll not get charged for GH Actions usages at all, as long as we stick with the base runners.

For matrices, the tests run in parallel.

mKowalski256 (approved):
@tchoi-rs I replayed the changes limiting tested MariaDB versions to (explicit) `10.11` and `12.3`, and updated the failing unit test. Is there anything more to add?

tchoi-rs:
@mKowalski256 I think we still need a minor check to ensure the instance is running at least 10.11.

mKowalski256:
> @mKowalski256 I think we still need a minor check to ensure the instance is running at least 10.11.

@tchoi-rs good point, I've updated the test.

PR #910, opened 2026-07-02
Enable Help button when Lighthouse fails to load

tchoi-rs:
## Description ##
- Add guard clause in _showLighthouse to return safely if widget unavailable
- Enable Help button when extra-help-links menu exists, not just when Lighthouse loads
- Allows fallback help access even when external widget fails to initialize


mKowalski256 (approved):
Seems fine on [test AWS build](https://feature-fix-lighthouse-fallback-5ae93563-2.researchspace.com/).

PR #911, opened 2026-07-02
Improve a11y for skip to content menu

tchoi-rs:
## Description ##
Improve a11y of the Skip to Content menu by adding reduced motion checks.


mrowe-tan:
Tested on an AWS instance, looks good to

mrowe-tan (approved):

PR #912, opened 2026-07-02
Remove slick and improve a11y for admin carousel

tchoi-rs:
## Description ##
Removes Slick in favour of a native CSS / JS scroll menu, plus a11y improvements.


mrowe-tan (approved):
Tested on an AWS instance, and looks good.

PR #914, opened 2026-07-03
Merge the simplicity theme CSS

tchoi-rs:
## Description ##
Merge the simplicity theme CSS so that they are cache-busted as a single entity.

Note: Fonts and images will now need to be manually cache-busted but as the amount of code is supposed to dwindle here I don't expect this to happen at all.


mKowalski256 (approved):
Looks good on [test AWS build](https://t3code-investigate-rsdev-1227-8bcd1478-3.researchspace.com/).

PR #921, opened 2026-07-06
Mock the inventory OAuth token endpoint globally in browser tests

tchoi-rs:
## Description ##
useOauthToken's background token refresh hits /userform/ajax/inventoryOauthToken, which isn't mocked outside Inventory-specific tests. In CI this produced unhandled AxiosError 404 rejections across unrelated browser specs (MainPanel, ImportDialog, FieldmarkImportDialog, CallablePdfPreview, etc.), failing the browser-tests run even though every individual test passed.

Extract the mock into oauthTokenMocks.ts and wire it into the global browserSetup.ts MSW worker so every browser test has it by default; inventoryMocks.ts now re-exports it for existing Inventory-specific imports.


rs-fraser (approved):
Code changes look good, but there's still 1 CI stage failing

PR #923, opened 2026-07-06
Shard vitest tests, add coverage

tchoi-rs:
## Description ##
Shard vitest tests for performance improvements, and add coverage to frontend.


rs-fraser (changes requested):
Image

Changes look good although the reports are being grouped under CLA Assistant instead of Lint and Test which is a bit confusing. Could you rename the CLA run or publish the reports under a different workflow?

tchoi-rs:
@rs-fraser This is a defect with GitHub: https://github.com/EnricoMi/publish-unit-test-result-action/issues/12

I've applied a workaround either way.

rs-fraser (approved):

PR #939, opened 2026-07-08
Remove outdated `produces` keywords in OpenAPI Spec

tchoi-rs:
## Description ##
Remove outdated `produces` keywords in OpenAPI Spec.


nebay-abraha (approved):
👍

PR #943, opened 2026-07-09
RSDEV-1093: reduce Inventory/ELN tag search debounce 1000ms -> 300ms

rs-fraser:
## Description

The tag search autocomplete felt sluggish. The dominant cause was the 1000ms debounce in `TagsCombobox`: filtered results only fired a full second after the user stopped typing, even though the server responds in ~14ms. This lowers the debounce to 300ms (the standard autocomplete range), which still coalesces a burst of keystrokes into one request.

One line: `debounce` default `timeout` 1000 -> 300.

## Testing

Tested locally with 3k tags with no noticeable lag


rs-fraser:
@greptileai

tchoi-rs (approved):

PR #951, opened 2026-07-13
Spring 6 / Hibernate 6 / Jakarta migration (RSDEV-444)

rs-fraser:
## Background

This is a coordinated upgrade of the entire fleet (32 repos) from Spring 5.3 / Hibernate ORM 5.6 / Hibernate Search 5 / Shiro 1.13 / Servlet 3.1 / `javax.*` to Spring 6.2 / Hibernate ORM 6.4 / Hibernate Search 7 / Shiro 2.1 / Jakarta EE 10 / `jakarta.*`.

Because of the `javax` → `jakarta` switch, this migration has to land as a single coordinated upgrade rather than incrementally. Every repo has the branch `rsdev-444-upgrade-to-spring-6`.

The 31 shared libraries all receive a major version bump because their compiled APIs genuinely break for consumers. `rspace-web` remains on its normal 2.x release train by product decision, with the upgrade guide documenting the operational changes that the version number no longer signals.

All fleet PRs are now open as drafts across the rspace-os repos, including the two largest (`rspace-core-model` and this one). `rspace-parent` has already been merged and released as 3.0.0, and this PR pins it at that released version. `rspace-core-model` is pinned to an immutable commit and `rspace-test-util` to a timestamped JitPack build (a workaround for JitPack intermittently returning 401 for moving `-SNAPSHOT` metadata). The remaining inter-project dependencies still point at `rsdev-444-upgrade-to-spring-6-SNAPSHOT` JitPack versions; these will be pinned to released versions once all PRs are approved, before merging.

### Tier 1: pom-only or near-pom-only changes, 17 repos

These repos contain either no source changes or only trivial ones. Check the parent/version bumps and confirm nothing unexpected slipped into the diff.

- argos-java-client
- dmptool-java-client
- dryad-java-client
- zenodo-java-client
- rspace-repository-spi
- rspace-document-conversion-spi
- rspace-digital-commons-data-adapter
- rspace-figshare-adapter
- rspace-dryad-adapter
- rspace-zenodo-adapter
- rspace-snapgene-adapter
- rspace-external-auth
- galaxy-java-client
- snapgene-java-client
- protocols-java-client
- raid-java-client
- fieldmark-java-client

### Tier 2: small source changes, 13 repos

- **rspace-parent:** Version matrix for the whole fleet. `jaxb-runtime` also moves to `runtime` scope. Already merged and released as 3.0.0.
- **rspace-core-util:** Servlet/JAXB imports moved to jakarta; Shiro 2.x `ByteSource` relocation.
- **rspace-test-util:** `JavaxValidatorTest*` renamed to `JakartaValidatorTest*`. This is a breaking change for subclasses; all in-fleet consumers have already been updated.
- **rspace-audit:** SpEL cache key changed from `#logFile` to `#p1`. Spring 6 only resolves named parameters when compiled with `-parameters`, which the fleet doesn't currently enable. The positional form works regardless. Enabling `-parameters` fleet-wide and reverting to the named key is a planned follow-up, not part of this migration.
- **rspace-rest-api-utils:** Spring 6 API updates: `HttpStatus` → `HttpStatusCode` on overrides, `handleBindException` reimplemented as `@ExceptionHandler` (no longer overridable), and the new `handleError(URI, HttpMethod, ...)` hook.
- **rspace-license-server-entities:** JPA/validation/EL migrated to jakarta, `@Length` replaced with `@Size`, commons-lang 2 upgraded to lang3.
- **rspace-evernote-parser**, **rspace-egnyte**, **rspace-dataverse-adapter** (now pointing at the `com.github.rspace-os` fork of dataverse-client-java), **datacite-java-client**, **figshare-client-java**, **digital-commons-data-java-client**, and **dataverse-client-java** (Gradle build; republished under `com.github.rspace-os`) each contain a small number of mostly mechanical changes, described in their PRs.

#### rspace-core-model (~230 files)

- **Hibernate Search annotation migration:** `@IndexedEmbedded` moves from prefix-based to named embeds; the names `extraFields`, `files`, and `notes` form a cross-repo contract with rspace-web's query builder (`LuceneSearchTermListFactory`, `RSQueryBuilder`).
- **Explicit `@TableGenerator` on every TABLE-strategy entity.** Hibernate 6's default TABLE generator would consolidate all bare `@GeneratedValue(strategy = TABLE)` entities onto one shared `default` segment, changing id allocation behaviour. Declaring an explicit `@TableGenerator` without `pkColumnValue` keeps the Hibernate 5 layout: one counter per entity, keyed by its physical table name. The two TABLE_PER_CLASS hierarchies (`AbstractUserOrGroupImpl`, `Communication`) share one counter across their sibling tables and declare `pkColumnValue` with `allocationSize = 1` explicitly. These segment names form a cross-repo contract with rspace-web's reseed changeset (below).
- **Shiro 2.1 crypto** (`SymmetricTextEncryptor`) protects persisted OAuth/integration tokens. A fixed-ciphertext compatibility test (generated with OpenSSL using an independent implementation) verifies that secrets encrypted by Shiro 1.5.x still decrypt correctly.
- **`@Cache` moved to hierarchy roots** (`AbstractUserOrGroupImpl`, `AbstractForm`). Hibernate 6 rejects `@Cache` on non-root entities during bootstrap, whereas Hibernate 5 only logged a warning. Moving the annotation changes the effective cache region name, so the corresponding region renames in `rspace-web`'s `ehcache.xml` are required and covered by `EhcacheRegionConfigTest`.

#### rspace-web (~700 files)

1. **Liquibase changesets in `changeLog-rsdev-444.xml`**
    - `hibernate6-rename-sequence-column`: renames the `hibernate_sequences` value column to `next_val`, the Hibernate 6 default.
    - `hibernate6-reseed-sequences`: Hibernate 5's legacy hi/lo generator stored a "high" marker in `hibernate_sequences`, not the next id; Hibernate 6's enhanced TableGenerator reads the value literally, so the small legacy values would hand out already-used primary keys across roughly 40 tables. A stored procedure loops over every existing `hibernate_sequences` row and, for any row naming a real base table with an `id` column, sets `next_val` to `MAX(id) + 51` (one full allocation block plus one). Driving off the existing rows makes it complete by construction, so no TABLE-strategy entity can be missed. The two TABLE_PER_CLASS hierarchy segments (`AbstractUserOrGroupImpl`, `Communication`), which are keyed by entity name rather than a table, are handled explicitly. This is the project's first stored procedure, so migration requires `CREATE ROUTINE`/`DROP ROUTINE` privileges.
    - `appconfigelementset-natural-id`: adds a stable UUID natural key to `AppConfigElementSet` so `equals`/`hashCode` are safe before and after persist.
    - `externalworkflowdata-enum-ordinal-to-string`: converts `ExternalWorkFlowData` enum columns from ordinals to name strings for Hibernate 6 compatibility.
    - `hibernate6-batch5-schema-migration`: recreates the Spring Batch `BATCH_*` tables for the Batch 5 schema while preserving job ID monotonicity via a holding table. Job execution history is intentionally discarded because it is only used for `/api/v1/export` monitoring.
2. **Transaction wiring**
    - `applicationContext-service.xml` continues to use `<tx:annotation-driven>` alongside the explicit pointcut advisor. Historically, relying on class-level `@Transactional` alone was unreliable because the Shiro filter eagerly instantiated around 200 service beans before the transaction advisor existed (SHIRO-829), leaving them unproxied and causing failures on first login.
    - That eager-initialisation window is now gone. The `securityManager` resolves lazily through the `LazySecurityManager*` classes, allowing the realm graph to initialise after context refresh when advisors already exist. This behaviour is enforced by `TransactionAdviceStartupCheck`, which fails startup if any early-instantiated `@Transactional` bean lacks a `TransactionInterceptor`.
    - `@EnableScheduling` moved from `BaseConfig` to the concrete profile configs. Removing it from `BaseConfig` (an AOP conflict fix) silently disabled the only `@Scheduled` job, WOPI discovery refresh; the profile configs now enable it alongside `@EnableAsync`.
3. **`GenericDaoHibernate.save()`**
    On main, this is `merge()` only. Under Hibernate 6, `merge()` returns a managed copy and now throws `EntityExistsException` if the supplied instance is already referenced by a parent collection in the session (Hibernate 5 allowed this). `save()` therefore now uses `persist()` for new or already-managed entities and falls back to `merge()` for detached ones. Assigned-ID entities look detached, so callers must use the returned instance; this is documented in the Javadoc and call sites have been updated.

    The shared `persistOrMerge` helper determines new vs detached through the `EntityPersister`.

    `PersistenceUnitUtil.getIdentifier()` returns null for property-access entities under Hibernate 6, which incorrectly routes detached entities through `persist()`. Existing hand-written persist-or-merge logic in subclass DAOs has been consolidated onto this helper.

    Direct `Session.saveOrUpdate()` calls elsewhere remain unchanged. Although deprecated in Hibernate 6, they still work and can be cleaned up separately during a future Hibernate 7 upgrade.
4. **Search**
    Indexes created by Lucene 5.5.5 cannot be opened by Lucene 9. Instead of failing startup with `IndexFormatTooOldException`, startup now deletes the old index directory, recreates it empty, and logs that a full reindex is required.

5. **Caching**
    `ehcache.xml` has been rewritten for the EhCache 3 schema. Typed cache keys were removed because Hibernate 6 now uses composite cache keys.

    hibernate-jcache 6.x only reads the `hibernate.javax.cache.*` settings (JCache never moved to the jakarta namespace), so the provider and config-uri keys have been corrected; the previous keys were silently ignored, leaving every L2 region an untuned default backed by an empty CacheManager instead of `ehcache.xml`.

    Every `@Cache` entity now has an explicit region, verified by `EhcacheRegionConfigTest`. `SecondLevelCacheTestIT` boots the production cache stack and asserts real L2 miss/put/hit behaviour and that Hibernate's regions are backed by the `ehcache.xml`-configured CacheManager.

6. **DispatcherServlet registration**
    - The DispatcherServlet is now registered programmatically via a `ServletContextListener`, replacing the `web.xml` servlet/multipart configuration.
    - Multipart limits are resolved from `-Dfiles.maxUploadSize`, then `deployment.properties`, then a 50 MB default.
    - One Tomcat-specific issue: multipart configuration is taken from the servlet that originally matched the request, not the UrlRewriteFilter target. Upload-receiving public URL prefixes (`MULTIPART_SOURCE_PATTERNS`) therefore also map directly to the DispatcherServlet. Without this, multipart uploads fail with HTTP 400 on Tomcat ("Required part ... is not present"), although Jetty masks the problem. A scanning test ensures new multipart controllers remain covered.
    - Those direct mappings collapse the forwarded servlet path (e.g. to `/workspace`), which defeated SiteMesh's path-based `/ajax/` exclusions and caused AJAX fragments to be decorated with the full page chrome. `FullPathSiteMeshFilter` restores matching against the full forwarded path.
7. **Trailing-slash request matching**
    Spring 6 stopped matching `/foo/` to a handler mapped at `/foo`, and the interim `setUseTrailingSlashMatch(true)` opt-in is deprecated and due for removal. Compatibility for external API clients and old bookmarks moves to a servlet filter (`TrailingSlashCompatibilityFilter`) that wraps each trailing-slash request as its slash-less equivalent before the UrlRewriteFilter forward, so the dispatcher only ever sees canonical paths. Handler mappings no longer end in `/`.

## Noticeable behavioural changes

- Spring Batch job execution history is discarded during upgrade.
- `HHH015007` Envers static metamodel ERROR messages appear on every startup. These are known and harmless.
- Values in `hibernate_sequences` increase by roughly 51 during migration. Id allocation itself is unchanged: each TABLE-strategy entity keeps its own per-table counter, as under Hibernate 5.
- The attachment full-text search index is deleted on first startup because Lucene 9 cannot read the old index format. Attachment search remains empty until a full reindex completes.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


rs-fraser:
@greptileai

src/main/resources/sqlUpdates/liquibaseConfig/rs-dbbaseline-utf8.sql
```diff
@@ -2821,7 +2821,7 @@ DROP TABLE IF EXISTS `hibernate_sequences`;
 /*!40101 SET character_set_client = utf8mb4 */;
 CREATE TABLE `hibernate_sequences` (
   `sequence_name` varchar(255) DEFAULT NULL,
-  `sequence_next_hi_value` int(11) DEFAULT NULL
```
mKowalski256, line 2824: if possible, we'd better avoid changing `rs-dbbaseline-utf8.sql` as each such change triggers some devops tasks. 

I checked the `next_val` is renamed to `sequence_next_hi_value` in very first changeset of `changeLog-rsdev-444.xml` file. That rename changeset could be moved to `inital-seed-changeLog.xml` that fires immediately after baseline script, so we're sure all the later changelogs are seeing the new column name. Although my AI agent claims even that is unnecessary, as no changeLogs before `changeLog-rsdev-444.xml` are relying on the new column name, so the old name can be imported by baseline and wait just until  `changeLog-rsdev-444.xml` is executed on either fresh or pre-existing deployments.

DevDocs/DeveloperNotes/PropertyFiles.md
```diff
@@ -53,6 +53,12 @@ In this case, the Maven build flag `-DpropertyFileDirPlaceholder=\$\{propertyFil
 with the name of a runtime system property `${propertyFileDir}`. On the production server, `${propertyFileDir}` is
 defined in Tomcat configuration file. E.g. `-DpropertyFileDir=/etc/rspace`
 
+### JVM system properties (not Spring-managed)
```
tchoi-rs, line 56: I don't have a huge preference for making the change now, but I think we should start using environment variables for these type of stuff.

DevDocs/DeveloperNotes/UpgradingToSpring6.md
```diff
@@ -0,0 +1,125 @@
...
+| Servlet container | Tomcat 9 | **Tomcat 10.1** (Jakarta servlet API; Tomcat 9 cannot run this WAR) |
+| Database | MariaDB 10.11+ | MariaDB 10.11+ |
+| JDBC driver | mysql-connector-java 5.1 | **mysql-connector-j 8.4** (bundled) |
+
+## Pre-upgrade checklist
+
+1. **Take a full database backup.** The upgrade rewrites id-sequence counters and
+   drops/recreates the Spring Batch tables in place. There is no downgrade path: rolling back
+   means restoring the backup and redeploying the previous WAR.
+2. **Install/configure Tomcat 10.1.** Native WebSocket is supported on Tomcat (development
```
tchoi-rs: Just checking, do we still need SockJS? Especially since the expectation now is to use the Docker dev stack which can have Apache/Tomcat in front of it.

DevDocs/DeveloperNotes/UpgradingToSpring6.md
```diff
@@ -0,0 +1,125 @@
...
+- The Java-based startup migrations run as part of the same boot. On a large database, expect
+  the first boot to take noticeably longer than usual.
+
+## Post-upgrade: search reindex (mandatory)
+
+The upgrade moves from Lucene 5 to Lucene 9 and Hibernate Search 5 to 7. The on-disk search
+index format is incompatible, so the index must be rebuilt. The rebuild mechanism is the
+startup reindex: with `rs.indexOnstartup=true` (the shipped default) the text index is rebuilt
+at boot. Before the first post-upgrade boot, confirm your deployment's properties do not
+override this to `false`; if they do, enable it for that first boot. Until the reindex
```
tchoi-rs: Note to self: we need a better story for background reindexing

DevDocs/public/RSpaceConfiguration.md
```diff
@@ -473,7 +473,7 @@ The following properties are specific to SMB filesystems:
 These optional settings configure  behaviour of the RSpace application.
 
 #### File upload and processing
-* **files.maxUploadSize** The maximum individual file size upload, in bytes. Default is 10Mb
+* **files.maxUploadSize** The maximum individual file size upload, in bytes. Default is 50MB (52428800). The value is read in this order: (1) JVM system property `-Dfiles.maxUploadSize=<bytes>` if set, otherwise (2) the `files.maxUploadSize` key in `${propertyFileDir}/deployment.properties`, otherwise (3) the 50MB default. The total multipart request limit is automatically set to 2x this value. See `DevDocs/DeveloperNotes/UpgradingToSpring6.md` for upgrade notes, including the Tomcat `maxSwallowSize` requirement.
```
tchoi-rs: Maybe `deployment.properties` should be removed in favor of always setting the JVM system property, if the `deployment.properties` value isn't going to be respected anyway? (Another point to using env vars in this case...)

src/main/resources/sqlUpdates/liquibaseConfig/rs-dbbaseline-utf8.sql
rs-fraser: Yeah I agree, reverted this change

rs-fraser:
Generated by claude code:

# PR #951 (RSDEV-444): files beyond the mechanical javax → jakarta translation

Analysis of https://github.com/rspace-os/rspace-web/pull/951 (`rsdev-444-upgrade-to-spring-6`, 722 changed files), generated 2026-07-17 against merge-base `701d374261` with `upstream/main`.

**Method:** a file counts as "mechanical" only when every changed line is identical after a `javax` → `jakarta` substitution (also excusing the equivalent JSTL taglib URI swap, e.g. `http://java.sun.com/jsp/jstl/core` → `jakarta.tags.core`). Import reordering within a file is tolerated (multiset comparison of removed-vs-added lines).

**Result:** 455 files are pure renames (422 javax imports, 33 taglib URIs). The remaining **267 files** are tabled below by theme. Bracketed numbers are residual changed-line counts (lines not explained by the rename), a rough proxy for review depth.

Review effort concentrates in three places: the DAO query rewrites (~25 files, biggest: `UserDaoHibernate`, `RecordDaoHibernate`, `RSQueryBuilder`), the Hibernate 6 persistence-semantics fixes (13 files where behavior genuinely changed), and the three infrastructure rewrites (Ehcache 3, Hibernate Search 6, Shiro/security lazy wiring).

## New files (24)

| File | Purpose |
|---|---|
| `src/main/java/com/researchspace/webapp/config/DispatcherServletInitializer.java` | Programmatic servlet/multipart registration; maps multipart source URLs onto the dispatcher so uploads survive urlrewrite forwards on Tomcat |
| `src/main/java/com/researchspace/webapp/filter/FullPathSiteMeshFilter.java` | SiteMesh 3 filter that matches on the full forwarded path so `/ajax/` excludes work after forwards |
| `src/main/java/com/researchspace/webapp/filter/TrailingSlashCompatibilityFilter.java` | Spring 6 dropped implicit trailing-slash matching; keeps old URLs working |
| `src/main/java/com/researchspace/webapp/filter/LazySecurityManagerShiroFilterFactoryBean.java` | Defers securityManager resolution so the filter (a BeanPostProcessor) doesn't eagerly instantiate the service-bean graph before context refresh |
| `src/main/java/com/researchspace/auth/LazySecurityManagerAuthorizationAttributeSourceAdvisor.java` | Same lazy-securityManager treatment for Shiro's annotation advisor |
| `src/main/java/com/researchspace/auth/JCacheShiroCache.java`, `JCacheShiroCacheManager.java` | Shiro auth caches backed by JCache/Ehcache 3, replacing shiro-ehcache 2 |
| `src/main/java/com/researchspace/search/impl/RSpaceLuceneAnalysisConfigurer.java` | Hibernate Search 6 programmatic Lucene analyzer config, replaces `@AnalyzerDef` |
| `src/main/java/com/researchspace/service/impl/TransactionAdviceStartupCheck.java` | Startup guard verifying transaction advice actually applies to Manager beans |
| `src/main/resources/sqlUpdates/changeLog-rsdev-444.xml` | Liquibase migration for the upgrade, including the hibernate_sequences reseed |
| `src/main/webapp/WEB-INF/pages/export/export_otherDialogs.jsp` | New empty include target referenced from `export.tag` |
| `DevDocs/DeveloperNotes/UpgradingToSpring6.md` | Developer migration guide |
| Tests: `DispatcherServletInitializerTest`, `FullPathSiteMeshFilterTest`, `TrailingSlashCompatibilityFilterTest`, `LazySecurityManagerShiroFilterFactoryBeanTest`, `JCacheShiroCacheTest`, `JCacheShiroCacheManagerTest`, `TransactionAdviceStartupCheckTest`, `FileIndexerTest`, `EhcacheL2RuntimeTest`, `EhcacheRegionConfigTest`, `TableIdGeneratorConfigTest` | Coverage for all of the above, plus a guard that every TABLE-strategy entity keeps its own id segment |
| `src/test/resources/applicationContext-dao.xml` | Test override of the DAO context (NoCachingRegionFactory to avoid JCache lifecycle clashes across Spring test contexts) |

## Deleted files (2)

| File | Why |
|---|---|
| `src/main/java/com/researchspace/dao/hibernate/DatabasePaginationUtils.java` | Criteria-based pagination helper, obsolete after the query rewrites |
| `src/test/resources/ehcache_DEV.xml` | Ehcache 2 test config, superseded |

## Hibernate query API rewrites (Criteria and createSQLQuery to typed HQL/JPA) (41)

The largest review surface.

| File | Change |
|---|---|
| `src/main/java/com/researchspace/dao/hibernate/UserDaoHibernate.java` [322] | Full Criteria to typed HQL/CriteriaBuilder rewrite |
| `src/main/java/com/researchspace/dao/hibernate/RecordDaoHibernate.java` [213] | Same, plus native query typing |
| `src/main/java/com/researchspace/dao/hibernate/CommunicationDaoHibernateImpl.java` [168] | Same |
| `src/main/java/com/researchspace/dao/hibernate/GroupDaoHibernateImpl.java` [164] | Same |
| `src/main/java/com/researchspace/dao/hibernate/CommunityDaoImpl.java` [152] | Same |
| `src/main/java/com/researchspace/dao/hibernate/FolderDaoHibernateImpl.java` [132] | Same |
| `src/main/java/com/researchspace/dao/hibernate/RecordGroupSharingDaoHibernateImpl.java` [130] | Same |
| `src/main/java/com/researchspace/dao/GenericDaoHibernate.java` [75] | Base DAO: Criteria/`@Required` removed, HQL getAll/count, persister-based id access |
| `src/main/java/com/researchspace/dao/hibernate/FileMetadataHibernate.java` [61] | Criteria to HQL; usage aggregation typing |
| `src/main/java/com/researchspace/dao/hibernate/ThumbnailDAOHibernate.java` [34] | Query-by-Example API removed; hand-built HQL with optional predicates |
| `src/main/java/com/researchspace/dao/hibernate/OrganisationDaoHibernate.java` [25] | Criteria to HQL |
| `src/main/java/com/researchspace/maintenance/dao/MaintenanceDaoHibernate.java` [25] | Criteria to HQL |
| `src/main/java/com/researchspace/dao/hibernate/NameDateFilterImpl.java` [22] | Criteria to HQL |
| `src/main/java/com/researchspace/dao/hibernate/SignatureDaoHibernateImpl.java` [19] | Criteria to HQL |
| `src/main/java/com/researchspace/dao/hibernate/NfsDaoHibernate.java` [15] | Criteria to HQL |
| `src/main/java/com/researchspace/dao/hibernate/FormDaoHibernate.java` [14] | Query typing, property paths |
| `src/main/java/com/researchspace/dao/hibernate/UserDeletionDaoHibernate.java` [12] | createSQLQuery to createNativeQuery, param setters |
| `src/main/java/com/researchspace/dao/hibernate/InstrumentDaoHibernateImpl.java` [15] | `DTYPE='...'` SQL fragments become `type(i) = ...`; `name=` becomes `editInfo.name=` (strict HQL) |
| `src/main/java/com/researchspace/dao/hibernate/InstrumentTemplateDaoHibernateImpl.java` [9] | Same DTYPE/type() and editInfo path fixes |
| `src/main/java/com/researchspace/dao/hibernate/InventoryDaoHibernate.java` [6] | `sharingMode='0'` string literals become qualified enum literals |
| `src/main/java/com/researchspace/dao/hibernate/ExternalWorkFlowDataDaoHibernate.java` [8] | Explicit `select efd` to fix fetch-join ambiguity |
| `src/main/java/com/researchspace/dao/hibernate/OAuthTokenHibernate.java` [5] | HQL property names corrected (`hashedAccessToken`/`hashedRefreshToken`) |
| `src/main/java/com/researchspace/dao/hibernate/UserConnectionDaoHibernate.java` [36] | Composite-id paths (`id.userId` etc.) required by Hibernate 6 |
| `src/main/java/com/researchspace/dao/hibernate/RSChemElementDaoHibernate.java` [2] | `parentId = null` becomes `parentId is null` |
| `src/main/java/com/researchspace/dao/hibernate/EcatCommentDaoHibernate.java` [4] | `session.delete()` becomes `session.remove()` |
| `src/main/java/com/researchspace/dao/hibernate/AuditDaoHibernateEnversImpl.java` [15] | Sort fields mapped to fully-qualified HQL paths; BigInteger result handling dropped |
| `ContainerDaoHibernateImpl.java`, `ExternalStorageDaoHibernate.java`, `FormUsageDaoHibernate.java`, `IconImgDaoHibernate.java`, `RSMetaDataHibernateDao.java`, `ListOfMaterialsDaoHibernateImpl.java` [2-8 each] | createSQLQuery/typed-setter renames and small query fixes |
| `src/main/java/com/researchspace/dao/DatabaseUsageByGroupGroupByResult.java`, `DatabaseUsageByUserGroupByResult.java`, `src/main/java/com/researchspace/admin/service/impl/SysAdminGroupsManagerImpl.java` | Native queries now return Long, not BigInteger |
| `src/main/java/com/researchspace/service/impl/DatabaseMetaDataManagerImpl.java`, `DBDataIntegrityChecker.java` | createNativeQuery/getSingleResult; the checker also gains `@Component` |
| Custom Liquibase updates (`src/main/java/com/researchspace/dao/customliquibaseupdates/`): `v47/UpdateFilestoreLinksInFieldContent.java`, `v47/UpdateInternalLinksInFieldContent.java`, `v53/UpdateMediaLinksInSignedDocument.java`, `v29/UpdateReferencesInSnippetAttachments.java`, `SampleRadioChoiceFieldOptionsFormatChange_1_73.java`, `BlobMigrationBase.java`, `CreateWorkbenchAndMoveSubSamples_1_70.java`, `DeploymentPropertyToSysPropertyRSPAC861_1_34.java`, `ThumbnailAddParentIds.java`, `UpdateEmptyACLsForMediaFiles_RSPAC2559.java`, `v41/CreateMissingUserFoldersForLabAdminsAndPIs.java`, `AbstractUpdateRequiringUserPermissions.java` | Criteria to HQL, embedded property paths, `uniqueResultOptional`, `evictEntityData()` |

## Hibernate Search 5 to 6 / Lucene 9 (12)

| File | Change |
|---|---|
| `src/main/java/com/researchspace/dao/hibernate/RSQueryBuilder.java` [342] | Full rewrite from the HS5 QueryBuilder DSL to the HS6 search DSL |
| `src/main/java/com/researchspace/dao/hibernate/FullTextSearcherImpl.java` [99] | FullTextSession replaced by SearchSession; pagination/scoring changes |
| `src/main/java/com/researchspace/search/impl/FileIndexer.java` [54] | New index lifecycle plus self-healing reindex when the index is missing on upgrade |
| `src/main/java/com/researchspace/dao/hibernate/TextSearchDaoHibernate.java` [27] | HS6 session/query API |
| `src/main/java/com/researchspace/dao/hibernate/SampleDaoHibernateImpl.java`, `SampleTemplateDaoHibernateImpl.java` [16-18] | Inventory full-text search moved to HS6 DSL |
| `src/main/java/com/axiope/search/FieldNames.java` [16] | New flat index fields (`fields_fieldData` etc.) reflecting HS6 embedded-field naming |
| `src/main/java/com/researchspace/search/impl/LuceneSearchTermListFactory.java` | Searches must now cover the extra flat fields |
| `src/main/java/com/axiope/search/SearchUtils.java` | Preserve BM25 natural score order when no orderBy given |
| `src/main/java/com/researchspace/search/impl/LuceneSearchStrategy.java` | Catch IndexNotFoundException and return empty results pre-initialisation |
| `src/main/resources/hibernate.cfg.xml` | Cache/search properties relocated so tests can override |
| `src/test/java/com/researchspace/dao/hibernate/FullTextSearcherTest.java`, `src/test/java/com/researchspace/service/SearchSpringTestBase.java` | Updated for HS6 APIs |

## Ehcache 2 to Ehcache 3 via JCache (12)

| File | Change |
|---|---|
| `src/main/resources/ehcache.xml` [515] | Rewritten to the Ehcache 3 schema; every cache region redeclared and tuned |
| `src/main/java/com/axiope/service/cfg/CacheConfig.java` [51] | Spring caching now built on the JCache provider instead of EhCacheCacheManager |
| `src/main/java/com/axiope/service/cfg/EhCacheConfigurer.java` [71] | Programmatic cache setup migrated to Ehcache 3/JCache |
| `src/main/java/com/axiope/service/cfg/SecurityRunProdConfig.java` [42] | Shiro caching switched to the new JCacheShiroCacheManager |
| `src/main/resources/applicationContext-dao.xml` [34] | `hibernate.dialect` now a property placeholder; HS6 backend properties; JCacheRegionFactory with the critical `hibernate.javax.cache.*` keys |
| `src/main/resources/log4j2.xml`, `log4j2-dev.xml` | Dropped net.sf.ehcache logger entries |
| `DevDocs/DeveloperNotes/Caching.md` | Docs updated |
| Tests: `SecondLevelCacheTestIT.java` [264], `EhCacheConfigurerConfigTest.java`, `SecurityRealmConfigTest.java`, `TestEhCacheManager.java`, `SecurityTestConfig.java` | L2-cache regions verified live; guards against the silent empty-CacheManager failure mode |

## Shiro 2 and security wiring (9)

| File | Change |
|---|---|
| `src/main/webapp/WEB-INF/security.xml` | shiroFilter and the annotation advisor deliberately drop their securityManager refs (resolved lazily; SHIRO-829 eager-init problem) |
| `src/main/resources/applicationContext-service.xml` | Documents and preserves the XML tx-advisor setup that annotation advice depends on |
| `src/main/java/com/researchspace/webapp/filter/AnonymousUserFilter.java`, `src/main/java/com/researchspace/webapp/controller/PublicNotebookEditorController.java`, `PublicStructuredDocumentController.java` | Shiro `WebUtils.issueRedirect` removed; plain `response.sendRedirect` |
| `src/main/java/com/researchspace/service/impl/GlobalInitManagerImpl.java` | Clears stale Shiro ThreadContext between test contexts; adds `@Transactional` |
| `src/test/java/com/researchspace/testutils/BaseManagerTestCaseBase.java`, `RunConfigWiringTestIT.java`, `src/test/java/com/researchspace/service/impl/ShiroTestUtils.java` | SecurityManager reset/re-apply between tests; wiring assertions |

## Transaction and bean wiring (20)

| File | Change |
|---|---|
| `src/main/java/com/axiope/service/cfg/BaseConfig.java` [60] | StandardServletMultipartResolver; `@EnableTransactionManagement` dropped (conflicts under component-scan), tx via XML; several `@Bean` methods moved out to annotations |
| `ProductionConfig.java`, `TestAppConfig.java`, `src/test/java/com/axiope/service/cfg/RSDevConfig.java` | `@EnableScheduling` relocated; GlobalInitManager wiring via `@Qualifier` params |
| `src/main/java/com/researchspace/api/v1/config/BaseAPIConfig.java` | `@EnableBatchProcessing` removed (Batch 5); throttler `@Qualifier`s |
| `src/main/java/com/researchspace/integrations/clustermarket/ClustermarketConfig.java` | `@EnableTransactionManagement` removed |
| `@Transactional` added: `AbstractAppInitializor`, `ChemistryImageUpdateInitialisor`, `ChemistrySearchIndexInitialisor`, `LuceneSearchIndexInitialisor`, `SystemConfigurationInitialisor`, `AbstractContentInitializer`, `ShareApiServiceImpl`, `ArchiveExportPlannerImpl` | Sessions must now be open for lazy traversal these code paths relied on |
| `@Service` added: `InternalFileStoreImpl`, `UserContentUpdaterImpl` | Beans moved from `@Bean` methods to component scan |
| `@Qualifier` fixes: `LicenseServerChecker`, `RSChemService`, `StoichiometryServiceImpl` | Disambiguation now required |
| `src/main/java/com/researchspace/webapp/integrations/datacite/DataCiteConnectorImpl.java` | `@PostConstruct` becomes `@EventListener(ContextRefreshedEvent)` + `@Transactional` |
| `DevDocs/DeveloperNotes/Transactions.md`, `AGENTS.md` | Docs for the `@Transactional` path alongside the Manager convention |

## Hibernate 6 persistence-semantics fixes (13)

The subtle behavioral ones, worth the closest review.

| File | Change |
|---|---|
| `src/main/java/com/researchspace/service/impl/RecordManagerImpl.java` | `getTempRecord()` is now a lazy proxy; initialise inside the transaction before the session closes |
| `src/main/java/com/researchspace/service/impl/DocumentCopyManagerImpl.java` | Copy now added to `targetFolder.children` and cascade-persisted; Hibernate 5 SAVE_UPDATE reachability is gone |
| `src/main/java/com/researchspace/service/impl/RecordSharingManagerImpl.java` | Same cascade-reachability fix for sharing, plus `@Transactional` |
| `src/main/java/com/researchspace/service/impl/SystemPropertyManagerImpl.java` | Re-fetch SystemProperty so cascade ALL doesn't hit a detached association (DetachedObjectException) |
| `src/main/java/com/researchspace/service/impl/ProdContentInitializerManager.java` | Re-fetch after bulk queries evict entities from the first-level cache |
| `src/main/java/com/researchspace/service/archive/export/AbstractArchiveExporter.java` | Assigned-id entity goes through merge(); use the returned managed copy |
| `src/main/java/com/researchspace/service/inventory/impl/InventoryFileApiManagerImpl.java`, `SampleApiManagerImpl.java`, `src/main/java/com/researchspace/service/impl/AbstractFormManagerImpl.java` | Explicit persists before flush; use merge return values |
| `src/main/java/com/researchspace/dao/hibernate/UserAppConfigDaoHibernate.java`, `InternalLinkDaoHibernate.java` | persistOrMerge / insert semantics |
| `src/main/java/com/researchspace/service/impl/AuditManagerImpl.java` | Force-initialise the doc thumbnail FileProperty while the session is open |
| `src/main/java/com/researchspace/service/impl/AppConfigManagerImpl.java` | Explicitly save the element set so it gets an id immediately |
| `src/main/java/com/researchspace/webapp/controller/StructuredDocumentController.java` | Removed `setForm(null)`/`setTempFieldForm(null)` mutation hacks on managed entities |

## Spring 6 API adaptations (25)

| File | Change |
|---|---|
| Interceptors: `LoggingInterceptor`, `BrowserCacheAdviceInterceptor`, `TimezoneInterceptor`, `ProductAndPropertyAnnotationInterceptor`, `OriginRefererCheckingInterceptor`, `AbstractThrottleInterceptor`, `ApiAuthenticationInterceptor`, `ApiPermissiveCorsInterceptor` | `HandlerInterceptorAdapter` (removed) becomes `implements HandlerInterceptor` |
| HttpStatus to HttpStatusCode: `PubchemSearcher`, `ChemistryClient`, `OrcidController`, `OmeroExceptionHandler`, `ClustermarketExceptionHandler`, `SpringWebClientNotFoundLoggedAsErrorExceptionHandlerVisitor`, `DMPAssistantController` | RestTemplate exceptions now expose HttpStatusCode; each adds a `toHttpStatus` resolve |
| `src/main/java/com/researchspace/api/v1/controller/ApiControllerAdvice.java` | BindException no longer handled by ResponseEntityExceptionHandler override; explicit `@ExceptionHandler` |
| Spring Batch 5: `src/main/java/com/researchspace/api/v1/config/BatchConfig.java`, `src/main/java/com/researchspace/api/v1/service/impl/ExportJobParamFactory.java` | JobBuilderFactory/DefaultBatchConfigurer removed; JobRepository-based builders, typed JobParametersBuilder |
| `src/main/java/com/axiope/service/cfg/WebSocketConfig.java` | Disables the SockJS websocket transport when not running on Tomcat |
| `src/main/java/com/researchspace/webapp/controller/GroupController.java`, `RSFormController.java` | Trailing-slash mappings normalised (`admin/` to `admin`); RSFormController also drops setForm(null) hacks |
| `src/main/java/com/researchspace/dao/spring/ext/IdTransferringMergeEventListener.java` | Hibernate 6 changed `entityIsTransient` parameters; rewritten against MergeContext |
| `src/main/java/com/researchspace/webapp/integrations/helper/OauthAuthorizationError.java` | Lombok `@Builder` replaced with a hand-written builder |
| `src/main/java/com/researchspace/api/v1/service/impl/ExportApiStateTracker.java` | Lombok `@Slf4j` swapped for an explicit logger |
| `src/main/java/com/researchspace/api/v1/model/ApiLinkItem.java` | Id serialised as string via ToStringSerializer |
| `src/main/java/com/researchspace/service/fieldmark/impl/FieldmarkToRSpaceApiConverter.java` | Extra switch cases for Number/Long/Double type names |
| `src/main/java/com/researchspace/webapp/integrations/protocolsio/ProtocolsIO_OAuthController.java` | Apache HttpClient 4 to 5 import |
| Behavior fixes spotted en route: `src/main/java/com/researchspace/service/impl/PostLoginHandlerImpl.java` (skip after-init actions for pre-initialised users instead of NPEing), `src/main/java/com/researchspace/service/impl/RSMetaDataServiceImpl.java` (version check no longer hardcodes "2.") | Genuine logic changes, each with a new test |

## Descriptors, config, build (13)

| File | Change |
|---|---|
| `pom.xml` [437] | The dependency overhaul itself: Spring 6, Hibernate 6/Search 6, Shiro 2, Ehcache 3, Batch 5, SiteMesh 3, Connector/J 8, jakarta artifacts, plugin bumps |
| `src/main/webapp/WEB-INF/web.xml` | Jakarta EE 10 web-app 5.0 schema; FullPathSiteMeshFilter and filter-chain rework |
| `src/main/resources/META-INF/persistence.xml` | JPA 3.0 schema and HibernatePersistenceProvider |
| `src/main/resources/sqlUpdates/liquibase-master.xml` | Includes the new changelog |
| `src/main/resources/deployments/defaultDeployment.properties`, `src/main/resources/sqlUpdates/liquibaseConfig/liquibase.properties` | `com.mysql.cj.jdbc.Driver` |
| `Jenkinsfile` | New Tomcat AMI |
| `.gitignore` | `/.worktrees/` |
| `src/main/webapp/WEB-INF/pages/notebookEditor/notebookEditor.jsp` | Adds `<body>` so SiteMesh 3 gets well-formed HTML |
| `src/main/webapp/common/taglibs.jsp` | JSTL `sql` taglib URI (same mechanical family, different URI) |
| `DevDocs/DeveloperNotes/PropertyFiles.md`, `DevDocs/public/RSpaceConfiguration.md` | files.maxUploadSize is now a JVM system property (CommonsMultipartResolver removed) |
| `src/test/resources/security-test.xml`, `src/main/java/com/researchspace/archive/package.html` | Whitespace only, effectively noise |

## Test adjustments (48)

| File | Change |
|---|---|
| `src/test/java/com/researchspace/service/BaseRecordManagerIT.java` | New assertions covering FileProperty/audited-entity loading |
| `GalleryControllerMVCIT.java`, `SysAdminUserRegistrationControllerMVCIT.java`, `GalleriesCrudopsMVCIT.java`, `ImportApiControllerMVCIT.java` | `fileUpload()` replaced by `multipart()` (removed in Spring 6 test) |
| `MediaManagerTest.java`, `ThumbnailManagerTest.java`, `UserAppConfigDaoTest.java`, `TestFactory.java`, `FolderNavigationServiceTest.java` | Managed-instance semantics: capture state before updates, no pre-assigned random ids, set bidirectional back-references |
| `SpringTransactionalTest.java`, `RealTransactionSpringTestBase.java` | Explicit DependencyInjectionTestExecutionListener; HS6 flushToIndexes; Criteria count replaced |
| `ProdContentInitializerTestIT.java` | Asserts in a separate transaction so commits are genuinely verified |
| `PostLoginHandlerImplTest.java` | New test for the contentInitialized skip |
| `src/test/resources/applicationContext-test-service.xml` | Test context mirrors production XML tx advisors |
| `Sandbox.java`, `FormDaoTest.java`, `EcatCommentDaoTest.java`, `FileIndexSearchTest.java` | Typed setters, strict HQL paths, new flat search-field names |
| Remaining ~28 (MVCIT and validation-model tests, 2 to 8 residual lines each): `DocumentsApiControllerMVCIT`, `ListOfMaterialsApiControllerMVCIT`, `ShareApiControllerMVCIT`, `FormControllerAcceptanceMVCIT`, `UserProfileControllerMVCIT`, `GroupControllerMVCIT`, `SDocControllerMVCIT`, `ExportImportManagerTestIT`, `UserDeletionManagerTestIT`, `AuditManager2IT`, `DirectoryControllerMVCIT`, `SignupControllerMVCIT`, `CommunityControllerMVCIT`, `CloudInvitedSignupTestMVCIT`, `ContainersApiControllerMVCIT`, `DocumentTagManagerTest`, `StructuredDocumentControllerTest`, `ContainerApiManagerTest`, `JournalControllerMVCIT`, `ScheduledMaintenanceControllerMVCIT`, `IntegrationControllerMVCIT`, `BaseRecordAdapterTest`, `ImageControllerMVCIT`/`Test`, `RoRSysAdminControllerMVCIT`, `SysAdminControllerMVCIT`, `PIOComponentDiscoverabilityIT`, `UpdateFilestoreLinkTypesInFieldContentIT`, `FileStoreConfiguredForExternalFS`, plus the api/v1 config and model validation tests | Small per-test tweaks: multipart, strict HQL in helper queries, changed validation messages, explicit flush/login, updated assertions |

## Near-mechanical library renames (17, borderline)

Not javax to jakarta, but still one-line find-and-replace within the same upgrade:

| Pattern | Files |
|---|---|
| `org.hibernate.validator.constraints.NotBlank/NotEmpty` to `jakarta.validation.constraints.*` (removed in Validator 8) | `ApiUserPost`, `UserDepositorAdapter`, `SysadminApiController`, `RepoDepositMeta`, `ExportApiController`, `ApiSubSampleNote`, `ValidName`, `ValidRoR`, `StockDeductionRequest`, `ExportToFileConfig`, `SysAdminCreateUser` |
| `org.apache.shiro.util.ByteSource` to `org.apache.shiro.lang.util.ByteSource` (Shiro 2) | `ShiroRealm`, `UsernamePasswordCredentialsMatcher`, `CryptoUtils`, `UserManagerImpl` |
| `GenerationType.TABLE` to `IDENTITY` in test-only sandbox entities | `src/test/java/com/researchspace/recordsandbox/Node.java`, `RelnEdge.java` |

src/main/java/com/researchspace/webapp/config/DispatcherServletInitializer.java
```diff
@@ -0,0 +1,171 @@
...
+    "/groups/*",
+    "/integration/*"
+  };
+
+  @Override
+  public void contextInitialized(ServletContextEvent sce) {
+    ServletContext ctx = sce.getServletContext();
+
+    long maxFileSize = resolveMaxFileSize();
+    long maxRequestSize = maxFileSize * 2;
```
tchoi-rs: This `*2` doesn't seem to be explained anywhere. If the intent is to allow two files, each 50MB, then when two files of exactly 50MB is specified, the multipart metadata would slightly extrude from the 50MB*2 limit.

src/main/java/com/researchspace/webapp/filter/TrailingSlashCompatibilityFilter.java
```diff
@@ -0,0 +1,43 @@
...
+import jakarta.servlet.ServletException;
+import jakarta.servlet.ServletRequest;
+import jakarta.servlet.ServletResponse;
+import java.io.IOException;
+import org.springframework.web.filter.UrlHandlerFilter;
+
+/**
+ * Transparently handles request paths with a trailing slash as if the slash were absent. Spring 5
+ * matched "/foo/" to a handler mapped at "/foo" by default; Spring 6 turned that off and deprecated
+ * the compatibility flag, but external API clients and old bookmarks may still use trailing-slash
```
tchoi-rs, line 15: I have a suspicion this isn't true, and we can simply stop including this compatibility fix

src/main/webapp/WEB-INF/pages/notebookEditor/notebookEditor.jsp
```diff
@@ -5,6 +5,8 @@
 <jsp:include page="notebookHeader.jsp" />
 </head>
 
+<body>
```
tchoi-rs: This should not be here

pom.xml
```diff
@@ -533,51 +664,50 @@
       </exclusions>
     </dependency>
     <dependency>
-      <groupId>javax.mail</groupId>
-      <artifactId>javax.mail-api</artifactId>
-      <version>${javax-mail-api.version}</version>
+      <groupId>jakarta.mail</groupId>
+      <artifactId>jakarta.mail-api</artifactId>
+      <version>${jakarta.mail-api.version}</version>
```
tchoi-rs: If `rspace-core-models` is supposed to be on its way out, do we still need to hoist this value?

DevDocs/DeveloperNotes/PropertyFiles.md
rs-fraser: Agreed, not as part of this PR though

DevDocs/DeveloperNotes/UpgradingToSpring6.md
rs-fraser: Yeah, if/when we no longer use Jetty locally then we'll not need SockJS either.

DevDocs/public/RSpaceConfiguration.md
rs-fraser: It'll still be used if the JVM value isn't set explicitly, so I'd rather keep it here for continuity on existing deployments. I agree it could be much tidier when we use env vars for all deployment properties.

src/main/java/com/researchspace/webapp/filter/TrailingSlashCompatibilityFilter.java
rs-fraser: Which part do you think isn't true? The trailing slash mapping, or the need to support for api clients/ old bookmarks?

src/main/webapp/WEB-INF/pages/notebookEditor/notebookEditor.jsp
rs-fraser: good spot, removed

src/main/java/com/researchspace/webapp/config/DispatcherServletInitializer.java
rs-fraser: Yeah not really sure where the *2 came from. The previous behaviour was 50MB total, so I've reinstated that

pom.xml
rs-fraser: Thanks for flagging. Digging a bit deeper, this was used to enforce `jakarta.mail-api:2.0.2` to align with the runtime `jakarta.mail:2.0.1`. 

A better fix is to use the new runtime impl `org.eclipse.angus:angus-mail:2.0.4` which aligns with `jakarta.mail-api:2.1.3` as specified in the parent pom.

tchoi-rs (approved):
Changes from my comments look good!

src/main/java/com/researchspace/dao/hibernate/RecordDaoHibernate.java
```diff
@@ -181,10 +173,18 @@ public ISearchResults<BaseRecord> getPaginatedChildRecordsOfParentWithFilter(
   private static String makeOrderBy(PaginationCriteria<? extends BaseRecord> pgCrit) {
     String orderBy;
     if (!StringUtils.isEmpty(pgCrit.getOrderBy())) {
-      orderBy = " order by " + pgCrit.getOrderBy() + " " + pgCrit.getSortOrder();
+      // Hibernate 6: editInfo properties need explicit path prefix
+      String field = pgCrit.getOrderBy();
+      if (!field.startsWith("editInfo.")) {
```
nhanlon2: Are all orderBy properties part of a EditInfo ? Can this be verified before appending 'editInfo.' to the orderBy?

src/main/java/com/researchspace/service/impl/TransactionAdviceStartupCheck.java
```diff
@@ -0,0 +1,134 @@
...
+      if (declaresTransactionalBehaviour(targetClass) && !hasTransactionAdvice(bean)) {
+        offenders.add(name + " (" + targetClass.getName() + ")");
+      }
+    }
+    return offenders;
+  }
+
+  private static boolean declaresTransactionalBehaviour(Class<?> targetClass) {
+    for (Class<? extends Annotation> annotation : TRANSACTIONAL_ANNOTATIONS) {
+      if (AnnotatedElementUtils.hasAnnotation(targetClass, annotation)) {
```
nhanlon2, line 117: Spring's real advice application (`AnnotationTransactionAttributeSource`) uses "find" semantics that traverse the hierarchy. So a bean whose `@Transactional` lives only on an interface (e.g. `ShareApiService`, `IContentInitializer`, `ArchiveExportPlanner`) or a superclass, and which was early-instantiated without advice, would slip through undetected. It does not cause any issue yet only because every such impl happens to co-locate a class-level `@Transactional`.

pom.xml
```diff
@@ -44,6 +44,114 @@
...
+      </dependency>
+      <dependency>
+        <groupId>org.projectlombok</groupId>
+        <artifactId>lombok</artifactId>
+        <version>1.18.42</version>
+      </dependency>
+      <dependency>
+        <groupId>jakarta.servlet</groupId>
+        <artifactId>jakarta.servlet-api</artifactId>
+        <version>6.1.0</version>
```
nhanlon2: this doesnt match the servlet.version property and the comment in the pom which says not to bump this above 6.0.0

pom.xml
```diff
@@ -2555,7 +2718,8 @@
 
   <properties>
     <!-- jetty -->
-    <jetty.version>10.0.18</jetty.version>
+    <jetty.version>11.0.26</jetty.version>
```
nhanlon2: This provides EE9 api but we compile against servlet-api 6.0.0 whiich is EE10. also note in another comment we also have 6.1.0)

src/main/java/com/researchspace/dao/hibernate/RecordDaoHibernate.java
rs-fraser: I've updated this to only check for `EditInfo` fields

pom.xml
rs-fraser: Yeah this should use the `servlet.version`, although it was 6.0.0 that ended up on the classpath, but this is misleading.

This and your next comment about jetty prompted me to look into moving Jetty to v12, which better aligns with Tomcat 10.1 and the correct javaEE version.

Will follow-up with that change

pom.xml
rs-fraser: I'm moving us to jetty 12 which uses the same servlet and javaEE version as Tomcat10.1

src/main/java/com/researchspace/webapp/filter/TrailingSlashCompatibilityFilter.java
tchoi-rs: The need to support old bookmarks with trailing slashes

src/main/java/com/researchspace/service/impl/TransactionAdviceStartupCheck.java
rs-fraser: This doesn't seem to be the case. I added tests covering interface and method level annotations in `TransactionAdviceStartupCheckTest.java`

Explanation from claude:

> Good catch to probe this. I dug into it and the check already covers that case: AnnotatedElementUtils.hasAnnotation is the find-semantics variant (unlike isAnnotated), so it searches the full type hierarchy, including superclasses and implemented interfaces, which is the same traversal AnnotationTransactionAttributeSource performs when Spring applies the advice. 
The method loop behaves the same way: getMethods() returns the interface-declared methods on the impl, and the per-method hasAnnotation call follows the hierarchy above each one.
So a bean like ShareApiService whose @Transactional lived only on the interface would still be reported even without the co-located class-level annotation. 
Since that behaviour is easy to lose in a refactor (for example by swapping in isAnnotated or a plain isAnnotationPresent), 5f19b946b adds annotationDeclaredOnlyOnInterfaceIsStillDetected to the unit test, covering both a class-level and a method-level @Transactional declared only on the interface, asserting the un-advised impls are flagged.

pom.xml
rs-fraser: Using Jetty 12 now

pom.xml
rs-fraser: Using Jetty 12 now

nhanlon2 (approved):

src/main/java/com/researchspace/webapp/filter/TrailingSlashCompatibilityFilter.java
rs-fraser: We had 17 examples of trailing slashes in our own test code, so I'd imagine some api consumers have done the same. If we want to remove support then we should do so in a smaller changeset release so it's easier to identify the change causing sudden 404s in client code.

Issue #27, opened 2024-06-24
Hardware requirements 


ll4strw:
Would it be possible to add a doc section that mentions minimal hardware requirements to run RSpace please? I cannot seem to find this info unless I am mistaken.

mKowalski256:
Yes makes sense, I've added the section to GettingStarted.md with [4ec2a21fcc77 commit](https://github.com/ResearchSpace-ELN/rspace-web/commit/4ec2a21fcc779270b199564ba0733ddeaa7c9dd5). Updated docs will be part of 2.0.0 release.

PR #963, opened 2026-07-20
Rsdev 1231

nhanlon2:
## Description

Adds a generic, config-driven Inventory **operation wizard**. A user selects one or more subsamples, a modal wizard collects the input, and a single atomic backend endpoint (`POST /api/inventory/v1/operations`) creates a new sample (with N subsamples, provenance links and custom fields) and adjusts each origin subsample, all in one transaction.

Operations are **data, not code**: each is an entry in `operations_config.json` (validated by a valibot schema), so adding or changing one is a config + i18n change with no per-operation Java and no per-operation branching in the wizard or the endpoint. Seven operations ship: **Aliquot, Passage, Pool, Derive, Cryopreserve, Revive, Destroy**.

Highlights:

- **Picker**: every operation is shown with a FontAwesome icon, enabled or greyed-out for the current selection (a single-origin operation needs exactly one subsample; Pool needs two or more of the same measurement category, with the reason shown when disabled).
- **Wizard steps**: Details, Template, Amounts, Documentation (optional), Confirm. One "remember values for this process" checkbox saves the whole bundle (template + documentation + amounts + amount mode) per process name; a step-one **fast path** lets a fully-remembered, valid run be performed in a single click without stepping through.
- **Pool** (the first multi-origin operation): a configurable "amount to take" mode - Take all (default for Pool), Same amount, or Per subsample - gated by the `takeAmountPerSubsample` / `defaultAmountMode` config. The confirmation labels it "Amount taken from each subsample" and lists every pooled origin.
- **Destroy** (terminal): empties the origin and stamps a disposal date on it, creating no new sample.
- **Confirmation**: a config-driven preview card of the sample to be created (or, for a terminal operation, the origin to be acted on).

## Public-API validation (stage 1)

Because `/operations` is public API, the backend now enforces every rule the wizard applies client-side, still with **no per-operation Java** (DevDocs/adr/0015):

- The backend keeps a **verbatim copy** of `operations_config.json` under `src/main/resources/inventory/`, parsed at startup into an `InventoryOperationConfigRegistry` (fail-fast on a bad file). `InventoryOperationsConfigDriftTest` fails the build if the two copies differ by a byte.
- `operationType` is now a validated **allowlist key** naming the definition the request must conform to (it was previously carried for audit only; DevDocs/adr/0006 is amended accordingly). The OpenAPI spec makes it required with an enum of the seven keys and documents the per-operation 400 rules.
- The generic validator enforces, per definition: origin cardinality (Pool needs two or more, everything else exactly one); `newSample` forbidden for Destroy and required otherwise; at least one created subsample, each holding a positive quantity with a real unit; amount-taken semantics (positive for decrementing operations, exactly zero for Passage); configured storage-temperature bounds compared **unit-aware** (Cryopreserve at most -18 C, Revive 4 C to 120 C; a value in Kelvin or Fahrenheit is judged on the temperature it denotes); and a provenance **link back to every origin** with the configured relation type (extra fields and links are allowed).
- The new sample is also run through the same `SampleApiPostValidator` as the public samples endpoint, closing a gap where the operations path bypassed it (reserved field names, link payloads, tag and name rules, temperature sanity).
- Live-state checks in the controller: every origin must currently hold something, over-removal is rejected (DevDocs/adr/0010), and Destroy must take exactly what the origin holds.
- The wizard's origin-quantity gate is now explicitly per-origin (`origins.every`), matching the backend rule.

Stage 2 (recorded in DevDocs/adr/0016) will replace the hard-coded definitions with user-configurable ones; the registry is the single seam that changes.

## Design decisions

Captured as ADRs under `DevDocs/adr/` (the repo's shared ADR home, one number sequence; the glossary lives in `CONTEXT.md`):

- `0006` frontend-declared operations, thin atomic backend (amended by 0015)
- `0007` origin amount adjusted independently of the created amounts
- `0008` derived-sample template is user-chosen
- `0009` every operation has a process name (the remember and naming key)
- `0010` amount-taken rejects over-removal (HTTP 400, not silent clamp)
- `0011` operation-functions registry (computed values, e.g. Passage number, disposal date)
- `0012` multi-origin operations (Pool)
- `0013` terminal operations (Destroy)
- `0014` configurable amount modes
- `0015` operation-aware backend validation (stage 1)
- `0016` user-configurable operations (stage 2 direction)

The endpoint remains a thin coordinator: the stateless validator interprets the shared definitions generically, the controller adds the live-quantity checks, and `InventoryOperationManager` applies the whole effect in one transaction via the `service.inventory.*Manager` AOP advice; creation reuses `SampleApiManager`. No new schema.

## Testing notes

- **Frontend**: Vitest unit and component tests across the `Operations/` module (config, request building, validation, remember bundle, wizard incl. the pool empty-origin gate, confirmation, picker, template step).
- **Backend**:
  - `InventoryOperationPostValidatorTest`: one golden request per operation, then each rule broken one at a time (46 tests), including cross-scale temperature cases (250 K passes Cryopreserve, fails Revive) and per-origin link checks.
  - `InventoryOperationConfigRegistryTest` + `InventoryOperationsConfigDriftTest`: config parsing, fail-fast, and frontend/backend copy equality.
  - `InventoryOperationsApiControllerTest`: the controller's live-state pass over mocked managers with the real validator chain (empty origin, over-removal, Destroy equality, short-circuiting).
  - `InventoryOperationsApiControllerMVCIT`: end-to-end via `POST /operations` (atomic create, links on the new sample and every created subsample, origin decrement, over-removal 400); verified green against a drop-recreated database.
- **Manual**: from an item-view subsample selection, open the operation menu and try:
  - Derive: remember a process, then reopen it to confirm the step-one fast path.
  - Pool: select two or more subsamples of the same measurement category, then try the Take all and Per subsample amount modes.
  - Destroy: confirm the origin is emptied and a disposal date is recorded, with no new sample created.


nhanlon2:
@greptileai

nhanlon2:
@greptileai

src/main/webapp/ui/src/Inventory/components/ContextMenu/ProcessAction.tsx
```diff
@@ -0,0 +1,64 @@
+import ScienceIcon from "@mui/icons-material/Science";
+import type MenuItem from "@mui/material/MenuItem";
+import { Observer } from "mobx-react-lite";
```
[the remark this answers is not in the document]
nhanlon2: Verified false positive: `pnpm tsc --noEmit` compiles clean. `React.ElementRef<typeof MenuItem>` is a type-query, which is a valid use of a type-only import (no value binding is needed), so no change. Leaving open for the author.

src/main/webapp/ui/src/components/Inputs/UnitSelect.tsx
```diff
@@ -39,7 +39,9 @@ function UnitSelect({ disabled, handleChange, value, categories }: UnitSelectArg
           inputProps={{
             "aria-label": t("inputs.unitSelect.quantityUnits"),
           }}
-          value={value}
+          // A non-positive id means "no unit chosen" (e.g. after amounts are cleared for a new
+          // process); render the dropdown empty rather than as an out-of-range value.
+          value={(value > 0 ? value : "") as number}
           size="small"
           sx={{
```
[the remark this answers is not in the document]
nhanlon2: Intentional and documented: an empty-string value is MUI's canonical "no selection" state (renders blank without an out-of-range warning). A 0 sentinel would need an extra hidden MenuItem for no behavioral gain, and the cast is localized. Leaving as-is.

src/main/java/com/researchspace/api/v1/controller/InventoryOperationsApiController.java
```diff
@@ -0,0 +1,64 @@
...
+    // where the user (hence read permission) is available. Only when the structural checks passed,
+    // so
+    // every origin has a valid id to load. Same 400/BindException contract as the other rules.
+    // The read runs in its own transaction, separate from the later performOperation mutation, so
+    // under concurrency this check is advisory: it can act on a slightly stale quantity. That is
+    // safe because registerApiSubSampleUsage subtracts and clamps at zero (an origin can only ever
+    // decrease, never go negative), so the worst case is an origin ending at zero rather than a
+    // 400.
+    if (!errors.hasErrors()) {
+      int index = 0;
```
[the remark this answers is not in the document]
nhanlon2: Known, documented design (see the comment above this block and adr/0002): the pre-load over-removal check is advisory under concurrency; the decrement clamps at zero so an origin can never go negative. True serialization would need pessimistic row-locking on the subsample, out of scope for this PR. Leaving open.

src/main/java/com/researchspace/api/v1/controller/InventoryOperationPostValidator.java
```diff
@@ -0,0 +1,101 @@
...
+      }
+      errors.popNestedPath();
+    }
+  }
+
+  private static boolean isNonNegativeQuantity(ApiQuantityInfo quantity) {
+    return quantity != null
+        && quantity.getNumericValue() != null
+        && quantity.getNumericValue().compareTo(BigDecimal.ZERO) >= 0;
+  }
```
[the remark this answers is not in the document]
nhanlon2: Fixed in c150a0126: the validator now rejects an amount-taken with a null unitId, so a missing unit fails with a 400 rather than NPEing in the manager's toQuantityInfo.

src/main/java/com/researchspace/service/inventory/InventoryOperationManager.java
```diff
@@ -0,0 +1,24 @@
...
+ * Java. The impl lives in {@code service.inventory} so the AOP transaction advice wraps it and the
+ * coordinated sub-manager calls join the same transaction (see adr/0001).
+ */
+public interface InventoryOperationManager {
+
+  /**
+   * @return the newly created sample (with its subsamples), as returned by the sample-creation
+   *     manager.
+   */
+  ApiSampleWithFullSubSamples performOperation(ApiInventoryOperationPost request, User user);
```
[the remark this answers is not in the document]
nhanlon2: Fixed in c150a0126: updated the interface Javadoc to state performOperation returns null for a terminal (noOutput) operation such as Destroy.

src/main/java/com/researchspace/api/v1/controller/InventoryOperationPostValidator.java
```diff
@@ -0,0 +1,101 @@
...
+          "The new sample must have a name");
+    }
+
+    if (CollectionUtils.isEmpty(request.getOrigins())) {
+      errors.rejectValue(
+          "origins",
+          "errors.inventory.operation.originsRequired",
+          "At least one origin subsample must be provided for the operation.");
+      return;
+    }
```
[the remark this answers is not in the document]
nhanlon2: Fixed in c150a0126: the validator now rejects duplicate origin ids, so the same subsample cannot be listed twice and decremented past the over-removal check.

src/main/webapp/ui/src/Inventory/components/Operations/TemplateStep.tsx
```diff
@@ -0,0 +1,155 @@
...
+            templateId: Number(template.id),
+            templateName: template.name,
+            quantityCategory: template.quantityCategory,
+          });
+        }
+      } finally {
+        setChecking(false);
+      }
+    })();
+  };
```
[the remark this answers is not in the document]
nhanlon2: Fixed in c150a0126: TemplateStep now guards the async template lookup with a latest-pick ref, so a superseded selection is discarded and the latest pick wins (regression test added).

src/main/webapp/ui/src/Inventory/components/Operations/operationsConfig.ts
```diff
@@ -0,0 +1,210 @@
...
+export type OperationInputConfig = v.InferOutput<typeof InputSchema>;
+export type OperationLinkSpec = v.InferOutput<typeof LinkSpecSchema>;
+export type OperationOriginFieldSpec = v.InferOutput<typeof OriginFieldSpecSchema>;
+export type OperationStep = v.InferOutput<typeof StepSchema>;
+export type ComputedArgSource = v.InferOutput<typeof ArgSourceSchema>;
+export type ComputedValueSpec = v.InferOutput<typeof ComputedSchema>;
+export type ConfirmSummaryField = v.InferOutput<typeof ConfirmSummaryFieldSchema>;
+export type InventoryOperation = v.InferOutput<typeof OperationSchema>;
+
+export const operations: Array<InventoryOperation> = v.parse(v.array(OperationSchema), rawConfig);
```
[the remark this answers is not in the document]
nhanlon2: Fixed in c150a0126: added assertEffectReferencesValid so every effect source key (nameFrom/contentFrom/...) must name a declared input or computed value, failing fast at module load.

nhanlon2:
@greptileai

src/main/java/com/researchspace/api/v1/controller/InventoryOperationsApiController.java
```diff
@@ -0,0 +1,64 @@
...
+      for (ApiInventoryOperationOriginUpdate origin : request.getOrigins()) {
+        errors.pushNestedPath(String.format("origins[%d]", index++));
+        ApiSubSample current = subSampleApiMgr.getApiSubSampleById(origin.getId(), user);
```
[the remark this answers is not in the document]
nhanlon2: Same concern as the thread above (Concurrent Over-Removal): a documented, accepted design. The pre-load over-removal check is advisory under concurrency; the actual decrement clamps at zero so an origin can never go negative, and true serialization would need pessimistic row-locking on the subsample, which is out of scope for this PR (see the comment in InventoryOperationsApiController and adr/0002). Leaving open for the author.

src/main/webapp/ui/src/Inventory/components/Operations/TemplateStep.tsx
```diff
@@ -0,0 +1,164 @@
...
+    const pickId = String(template.id);
+    latestPickRef.current = pickId;
+    setBlockError(null);
+    setChecking(true);
+    onChange({ ...value, templateId: null, templateName: template.name });
+    void (async () => {
+      try {
+        await template.fetchAdditionalInfo();
+        // A newer pick has superseded this one: drop this stale result without touching state.
+        if (latestPickRef.current !== pickId) return;
```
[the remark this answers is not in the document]
nhanlon2: Fixed in ae744d088: setMode now invalidates the pending-lookup ref and clears the checking spinner, so switching to "No template" / "From this sample" while a lookup is in flight discards its late result (regression test added).

src/main/webapp/ui/src/Inventory/components/Operations/operationsConfig.ts
```diff
@@ -0,0 +1,241 @@
...
+/**
+ * Fail fast (like the parse and the computed-values check above): every effect source key -
+ * nameFrom, the amount/temperature sources, and each text/origin field's contentFrom - must name
+ * either a declared input or a computed `into`, otherwise a config typo silently produces empty
+ * content or a malformed quantity at submit instead of failing when the config loads. See adr/0001.
+ */
+export function assertEffectReferencesValid(ops: Array<InventoryOperation>): void {
+  for (const op of ops) {
+    const available = new Set<string>(op.inputs.map((i) => i.key));
+    for (const computed of op.effect.computed ?? []) available.add(computed.into);
```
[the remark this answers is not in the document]
nhanlon2: Fixed in ae744d088: assertEffectReferencesValid now also validates computed `{ input: "key" }` args against the available-key set, so a typo'd input arg fails fast at module load (tests added).

nhanlon2:
@greptileai

nhanlon2:
@greptileai

nhanlon2:
@greptileai

nhanlon2:
@greptileai

nhanlon2:
@greptileai

nhanlon2:
@greptileai

adr/0001-frontend-declared-operations-thin-atomic-backend.md
```diff
@@ -0,0 +1,82 @@
...
+vocabulary validated for inventory record-to-record links, and
+`POST /api/inventory/v1/samples` which creates a Sample + N subsamples +
+extraFields (including links) + storage temperature in one call. What is missing
+is atomicity across the sample-creation call and the origin-quantity update: the
+bulk endpoint cannot mix CREATE and UPDATE, so combining them from the client is
+non-atomic.
+
+## Decision
+
+- Operations are declared in a single `operations_config.json` that lives with the
```
tchoi-rs, line 39: This is a major blocker for API developers because the shape of the operations are not defined in API specifications and would require reverse-engineering to replicate what the UI does.

Are we expecting a lot of new operations to arrive in the short to medium term, and do they truly need to be decoupled between the frontend and the backend? If the operations themselves don't change frequently, I think it's reasonable to hard-code them into the application, which would make the API definition more stable; instances wanting new operations can then upgrade their RSpace version to receive the new operations.

tchoi-rs (changes requested):

adr/0001-frontend-declared-operations-thin-atomic-backend.md
nhanlon2: This was what Tilo requested.

nhanlon2:
@greptileai

DevDocs/adr/0006-frontend-declared-operations-thin-atomic-backend.md
```diff
@@ -0,0 +1,82 @@
...
+vocabulary validated for inventory record-to-record links, and
+`POST /api/inventory/v1/samples` which creates a Sample + N subsamples +
+extraFields (including links) + storage temperature in one call. What is missing
+is atomicity across the sample-creation call and the origin-quantity update: the
+bulk endpoint cannot mix CREATE and UPDATE, so combining them from the client is
+non-atomic.
+
+## Decision
+
+- Operations are declared in a single `operations_config.json` that lives with the
```
nhanlon2: **I have added validation to the API endpoint. It parses a copy of operations_config.json to establish validation rules. This is a compromise so that in 'stage 2' the validation will already be somewhat flexible and ready to be user configurable.**

**Claude's summary:** Addressed in f0def2282 (stage 1 of the plan recorded in DevDocs/adr/0015): the OpenAPI spec now defines the contract rather than leaving it implicit - operationType is required with an enum of the seven keys, the per-operation shapes are documented with a conforming example for each, and the per-operation 400 rules are spelled out. The backend also enforces those definitions server-side (origin cardinality, newSample presence, amount semantics, temperature bounds, per-origin provenance links, plus the full samples-endpoint validation of newSample), so an API developer gets a clear 400 instead of silently diverging from what the UI does. The definitions stay data rather than hard-coded Java because stage 2 (DevDocs/adr/0016) makes them user-configurable, which is the direction that was requested; the backend keeps a drift-tested verbatim copy of the config so the two sides cannot disagree. Leaving this open for you to judge.

nhanlon2:
@greptileai

src/main/java/com/researchspace/api/v1/controller/InventoryOperationsApiController.java
```diff
@@ -0,0 +1,91 @@
...
+    // operation (e.g. Destroy) must take exactly what the origin holds. These need each origin's
+    // live quantity, which the stateless structural validator cannot load, so they run here where
+    // the user (hence read permission) is available. Only when the structural checks passed, so
+    // every origin has a valid id to load and the operation type is known. Same 400/BindException
+    // contract as the other rules.
+    // The read runs in its own transaction, separate from the later performOperation mutation, so
+    // under concurrency these checks are advisory: they can act on a slightly stale quantity. That
+    // is safe because registerApiSubSampleUsage subtracts and clamps at zero (an origin can only
+    // ever decrease, never go negative), so the worst case is an origin ending at zero rather than
+    // a 400.
```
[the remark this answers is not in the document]
nhanlon2: Deliberate, documented trade-off rather than an oversight: DevDocs/adr/0010 records the advisory-race contract (and DevDocs/adr/0015 inherits it for the empty-origin and must-empty checks). The pre-mutation checks give clean 400s in the non-concurrent case; under a genuine race the mutation path (registerApiSubSampleUsage) subtracts unit-aware and clamps at zero, so the worst case is an origin ending at zero, never negative stock or corruption. Enforcing the invariant atomically would need pessimistic locking (SELECT FOR UPDATE or version checks) inside the InventoryOperationManager transaction, which is a wider transactional design change than this PR should carry; happy to raise it as a follow-up ticket if the team wants the stricter contract.

src/main/java/com/researchspace/service/inventory/impl/InventoryOperationManagerImpl.java
```diff
@@ -0,0 +1,73 @@
...
+    // Mutate origins in ascending id order (not request order) so two concurrent multi-origin
+    // operations over overlapping origins acquire their row locks in one consistent order and
+    // cannot deadlock. The validator guarantees unique, non-null ids by this point.
+    List<ApiInventoryOperationOriginUpdate> originsById =
+        request.getOrigins().stream()
+            .sorted(Comparator.comparing(ApiInventoryOperationOriginUpdate::getId))
+            .toList();
+    for (ApiInventoryOperationOriginUpdate origin : originsById) {
+      subSampleApiMgr.registerApiSubSampleUsage(
+          origin.getId(), origin.getAmountTaken().toQuantityInfo(), user);
```
mKowalski256, line 59: here the subsample usage is registered, but somehow that doesn't result in new historical version being created.

mKowalski256 (changes requested):
I tested a bit on [AWS build](https://rsdev-1231-454a3c19-22.researchspace.com), and I can confirm the issue originally spotted by Tim i.e. that 'operations' endpoint doesn't seem to be bumping the revision of the processed subsample, even when its quantity is changed (reduced). That's different from regular 'edit' workflow, where changing the quantity does result in new revision being created.

To reproduce:
1. open a subsample, check its revision history & revision number (in 'info' panel)
2. edit the subsample in the UI, reduce its quantity, save
3. check subsample's quantity, revision history & revision number 
-- quantity is updated, revision number is bumped, and revision history records both revision
4. process the subsample with operation that reduces the volume (e.g. 'aliquot')
5. select the subsample again, check it's quantity, revision history & revision number
-- quantity is updated, but there is no new revision

In principle every operation that manipulates the original item should create a new revision of it.

src/main/java/com/researchspace/service/inventory/impl/InventoryOperationManagerImpl.java
nhanlon2: https://researchspace.atlassian.net/browse/RSDEV-1318

src/main/java/com/researchspace/dao/hibernate/SampleDaoHibernateImpl.java
```diff
@@ -216,12 +216,13 @@ public boolean entityNameExistsForUser(String name, User owner) {
...
     // The explicit editInfo.name path is required: unqualified embedded sub-properties (like
     // "name") resolve only on concrete-leaf persisters, not on the abstract hierarchy root.
+    // Deleted (soft-deleted) records do not count: a deleted sample's name is free to reuse.
     long count =
         sessionFactory
             .getCurrentSession()
             .createQuery(
                 "select count(se) from SampleEntity se where se.editInfo.name=:name and"
-                    + " se.owner=:owner",
+                    + " se.owner=:owner and se.deleted=false",
```
mKowalski256, line 225: This is a slight change to behaviour, which doesn't seem to be requested/documented (beyond the code comment), but seems fine to me.

mKowalski256 (commented):
Reviewed the backend code & retested everything on [AWS build #27](https://rsdev-1231-454a3c19-27.researchspace.com/).

Looks good to me, but I found one issue - using the 'destroy' operation is bumping item by two versions, while I'd expect it to be bumped by just one version; AI says that's because decrement action, and adding 'disposed date' action are registered separately. Or is that maybe known & covered by [PR#1048](https://github.com/rspace-os/rspace-web/pull/1048)?

src/main/java/com/researchspace/dao/hibernate/SampleDaoHibernateImpl.java
nhanlon2: Yes that is covered in https://github.com/rspace-os/rspace-web/pull/1048? - the operations all run withing a single transaction and the code will now not allow the version bump to happen twice within a transaction boundary.

mKowalski256 (approved):
Fine for me to merge the backend side.

src/main/java/com/researchspace/api/v1/controller/InventoryOperationsApiController.java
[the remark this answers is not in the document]
tchoi-rs: The inventory operation should assume a singular transaction context because that's what the contract (carrying out multiple operations in a single request) promise.

src/main/java/com/researchspace/api/v1/controller/InventoryOperationPostValidator.java
```diff
@@ -0,0 +1,462 @@
...
+
+  /**
+   * A valid amount-taken is a non-negative numeric value carrying a real unit. The unit is required
+   * because the manager converts it to a {@link com.researchspace.model.units.QuantityInfo}
+   * (unit-aware subtraction); a null or non-positive unit would fail there with a 500 rather than a
+   * clean 400. The frontend uses a non-positive unit id (UNSET_UNIT = 0) as an "unset" marker, so
+   * the unit id must be present and greater than zero. A zero numeric value is still allowed (a
+   * no-op decrement, e.g. Passage); a non-positive unit id is not.
+   */
+  private static boolean isValidAmountTaken(ApiQuantityInfo quantity) {
```
tchoi-rs, line 801: This does not take into accout that `QuantityInfo.java` has a precision limit of 3 digits. (I think the ownership of validity should live in `QuantityInfo`? But the positive check can stay here)

src/main/java/com/researchspace/api/v1/InventoryOperationsApi.java
```diff
@@ -0,0 +1,30 @@
...
+ * Runs a configured Inventory operation (see the frontend {@code operations_config.json}). A single
+ * generic, atomic endpoint: it creates one new sample (with its subsamples, custom fields and
+ * relation links) and sets the origin subsamples' quantities, all in one transaction. There is no
+ * per-operation endpoint or logic; a new operation is a new frontend config entry. See
+ * DevDocs/adr/0007.
+ */
+@RequestMapping("/api/inventory/v1/operations")
+public interface InventoryOperationsApi {
+
+  @PostMapping
```
tchoi-rs: AI agent found that this endpoint accepts YAML in `links` when it should not:

> Blocking: this JSON endpoint currently accepts YAML through the globally registered YAML converter. Live tests confirmed both YAML media types, alternate numeric forms, duplicate keys, and ignored trailing documents.
>
> Declare `consumes = MediaType.APPLICATION_JSON_VALUE` and add MVC tests proving YAML returns HTTP 415 without changing inventory state.

src/main/java/com/researchspace/api/v1/model/ApiInventoryOperationPost.java
```diff
@@ -0,0 +1,35 @@
...
+@JsonPropertyOrder({"operationType", "origins", "newSample"})
+public class ApiInventoryOperationPost {
+
+  @JsonProperty("operationType")
+  private String operationType;
+
+  @JsonProperty("origins")
+  private List<ApiInventoryOperationOriginUpdate> origins = new ArrayList<>();
+
+  @JsonProperty("newSample")
```
tchoi-rs, line 39: AI agent found a bypass for validation and operation costs:

> Blocking for the operation-specific image bypass: `newSample` does not cascade Bean Validation, and `ApiSampleWithFullSubSamples.subSamples` does not cascade into its elements. This lets operations bypass the image-size and note constraints that ordinary sample creation enforces.
> 
> Add `@Valid` through the request DTO graph. Also cap the explicit `subSamples` list at 100. The existing `newSampleSubSamplesCount` limit does not apply to this list. Please add an aggregate operation-cost limit or track the remaining nested collection limits as shared API work.

DevDocs/adr/0007-operation-wizard-decisions.md
```diff
@@ -0,0 +1,93 @@
...
+## The one thing a future reader must know first
+
+**User-configurable operations are the decided end state, deliberately not built
+yet.** The whole framework exists so that operations eventually become
+server-owned, user-editable data (storage, editing surface, and authorization are
+undecided stage-2 questions). Stage-1 code was written to survive that migration,
+which explains its most surprising artifact:
+
+**`operations_config.json` is deliberately duplicated.** The frontend owns the
+authoritative copy; the backend holds a verbatim copy at
```
tchoi-rs: The frontend must not own the authoritative copy. These are well-defined operations and in the current iteration, it is the backend that must validate the input strictly against a known schema.

tchoi-rs (changes requested):
I'm not seeing a lot of evidence that there is validation code that locks down exactly to what operations are permitted, what input each operation can accept (including whether the input set is exactly what the operation expects). There are some validation for some specific parts of the payloads but nothing that chains them together to validate the entire request before it's executed.

What I would like to see is Java code / JSON definitions that tightly compares the input against the specific schema and disallows deviations.

src/main/java/com/researchspace/api/v1/controller/InventoryOperationPostValidator.java
nhanlon2: This response is from Claude: Fixed in bc4bf1610. QuantityInfo.canStoreWithoutRounding now owns the 3dp rule; the validator rejects finer amountTaken with errors.inventory.operation.amountTakenTooPrecise (the non-negative check stayed in the validator).

src/main/java/com/researchspace/api/v1/InventoryOperationsApi.java
nhanlon2: This response is from Claude: Fixed in bc4bf1610. The endpoint now declares consumes=application/json, so YAML media types get HTTP 415. A standalone MockMvc test registers the production YAML converter, asserts 415 for application/x-yaml, application/yaml and text/yaml, and verifies the manager is never invoked.

src/main/java/com/researchspace/api/v1/model/ApiInventoryOperationPost.java
nhanlon2: This response is from Claude: Fixed in bc4bf1610. @Valid now cascades through the request DTO graph (newSample, origins, subSamples, notes), so the image-size and note-length constraints hold for operation payloads, and the explicit subSamples list is capped at 100. The aggregate operation-cost limit is left as shared API follow-up work.

DevDocs/adr/0007-operation-wizard-decisions.md
nhanlon2: This response is from Claude: Fixed in bc4bf1610. operations_config.json is backend-only now: the single copy lives in src/main/resources/inventory, is served verbatim from GET /api/inventory/v1/operations/config, and the wizard fetches and valibot-validates it. The frontend copy and the byte-equality drift test are deleted; ADR 0007 updated.

src/main/java/com/researchspace/api/v1/controller/InventoryOperationsApiController.java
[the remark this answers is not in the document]
nhanlon2: This response is from Claude: Fixed in bc4bf1610. The empty-origin, over-removal and must-empty checks now run inside InventoryOperationManagerImpl, in the same transaction as the mutation and before anything is written, so they hold against the state the mutation sees. Violations surface as the same field-scoped 400 BindException.

src/main/java/com/researchspace/api/v1/controller/InventoryOperationPostValidator.java
```diff
@@ -0,0 +1,786 @@
+package com.researchspace.api.v1.controller;
```
tchoi-rs, line 1: This file has become quite big. Can we split this into multiple, individual validators?

src/main/java/com/researchspace/api/v1/controller/InventoryOperationPostValidator.java
```diff
@@ -0,0 +1,786 @@
...
+   * clean 400. The frontend uses a non-positive unit id (UNSET_UNIT = 0) as an "unset" marker, so
+   * the unit id must be present and greater than zero. A zero numeric value is still allowed (a
+   * no-op decrement, e.g. Passage); a non-positive unit id is not.
+   */
+  private static boolean isValidAmountTaken(ApiQuantityInfo quantity) {
+    return quantity != null
+        && quantity.getNumericValue() != null
+        && quantity.getNumericValue().compareTo(BigDecimal.ZERO) >= 0
+        && quantity.getUnitId() != null
+        && quantity.getUnitId() > 0;
```
tchoi-rs, line 806: We should also check that `unitId` is a valid unit

src/main/webapp/ui/src/Inventory/components/Operations/TemplateStep.tsx
```diff
@@ -0,0 +1,179 @@
...
+        const fields = template.fields.map((f) => ({
+          name: f.name,
+          mandatory: f.mandatory,
+          hasDefault:
+            (f.selectedOptions?.length ?? 0) > 0 ||
+            (f.content !== null && f.content !== undefined && String(f.content).trim() !== ""),
+        }));
+        const { blocked, missingFields } = templateSelectionBlock(fields);
+        if (blocked) {
+          setBlockError(t("operations.template.mandatoryFieldsError", { fields: missingFields.join(", ") }));
```
tchoi-rs, line 107: Let the translation handle list joining

src/main/webapp/ui/src/Inventory/components/Operations/__tests__/OperationWizard.test.tsx
```diff
@@ -0,0 +1,677 @@
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
```
tchoi-rs, line 1: This file does not use MSW for network mocking (this might be because the skill for it was introduced after the fork from `origin/main`)

src/main/webapp/ui/src/Inventory/components/ContextMenu/__tests__/ContextDialog.test.tsx
```diff
@@ -0,0 +1,38 @@
+import { fireEvent, render } from "@testing-library/react";
```
tchoi-rs, line 1: This file does not use `userEvent` for interaction mocking (this might be because the skill for it was introduced after the fork from `origin/main`)

tchoi-rs (changes requested):

PR #972, opened 2026-07-24
Rsdev 173 Inventory Attachment Backlinks in Gallery

nhanlon2:
## Description

Jira: RSDEV-173

Extends the Gallery item info panel's "Related inventory items" section (added in RSDEV-1131 for inventory Links) to also list Inventory items that **attach** the Gallery file, not just those that link to it. Link rows keep their DataCite relation type in the Relation column; attachment rows are labelled "Attachment", so the two kinds are distinguishable within the one grid. Both record-level attachments and attachments made through a sample/template attachment field are covered, each reported against its owning inventory item (Sample, SubSample, Container, Instrument, or Template).

## Design decisions

- **Separate endpoint, merged client-side.** Attachments are served from a new session endpoint `GET /workspace/getAttachingInventoryItems/{globalId}` and merged with the existing links endpoint in the shared hook, keeping InventoryFile concerns out of `InventoryLinkManager`.
- **Non-disclosure gate reused (ADR-0002).** The endpoint resolves the target through `LinkTargetResolver.targetExistsAndIsReadable` (as the links endpoint does), so an unreadable, missing, malformed, or non-Gallery Global ID all return one identical not-found and never leak a file's existence or its inbound attachments. This is collision-safe for the shared `GL`/`FL`/`SD` id space.
- **Record-level and field-level attachments.** Field-level attachments are resolved back to their owning item through the owning attachment field (the attachment's own record reference is null in that case). Both new queries filter out soft-deleted attachment files, and the field query also excludes soft-deleted fields.
- **One row per connection**, no de-duplication, matching the links panel's behaviour.
- **Graceful frontend degradation.** The hook only calls the attachments endpoint for Gallery (`GL`) targets, and if that request fails it degrades to showing links only rather than blanking the whole section.

## Testing notes

- Backend: manager unit tests (permission filtering, record/field resolution, the non-disclosure gate), DB-backed `SpringTransactionalTest` cases (record-level, field-level, soft-deleted exclusion, and a `GL`-id-resolving-to-a-non-media-record rejection), and an MVCIT for the session endpoint.
- Frontend: Vitest coverage for the two-endpoint merge, attachment labelling, `GL`-gating, graceful degradation on an attachments-endpoint failure, stale-response handling, and malformed-row skipping.
- CI note: the `integration-tests` job is red on a pre-existing, order-dependent flake in the quarantined `ExportApiControllerMVCIT` (fails in test setup, `initUser`). It reproduces on the base commit with this change reverted, so it is unrelated to this PR; this PR's own MVCIT passes.

### DIFF
--- a/DevDocs/CONTEXT.md
+++ b/DevDocs/CONTEXT.md
@@ -29,3 +29,32 @@ useful: removed for deleted or inaccessible ELN targets (their routes produce
 error pages), kept for deleted Inventory targets (the trash viewer works) and
 for all readable targets.
 _Avoid_: view, go to
+
+## Related inventory items (Gallery info panel)
+
+**Referencing item**:
+An Inventory item that connects to a Gallery file the viewer is looking at, and
+so appears as a back-reference row in the Gallery info panel's "Related
+inventory items" section. The connection is one of two **relation kinds**:
+_Link_ or _Attachment_. The section shows both, one row per connection.
+_Avoid_: related item, back-link
+
+**Link (relation kind)**:
+An `InventoryLink` on the item pointing at the Gallery file by Global ID (RSDEV-1131).
+Carries a DataCite _relation type_ (IsPartOf, References, ...) shown in the
+Relation column. No file bytes are copied.
+_Avoid_: reference, pointer
+
+**Attachment (relation kind)**:
+An `InventoryFile` whose `mediaFile` is the Gallery file: an inventory item
+attached that Gallery file (its bytes were copied into the item). Reported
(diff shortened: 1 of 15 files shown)


nhanlon2:
@greptileai

nhanlon2:
@greptileai

tchoi-rs (approved):
Frontend portion looks fine!

rs-nicof:
PR backend code looks good! 
Tested successfully on AWS https://rsdev-173-30564cd8-3.researchspace.com/

rs-nicof (approved):

PR #973, opened 2026-07-24
PRT-1118: fix Gallery create-menu page freeze with DMP integrations enabled

nhanlon2:
## Summary

Fixes PRT-1118: with a DMP integration enabled (e.g. DMPTool), opening the Gallery "create" menu and then dismissing it (Esc or click-away) froze the whole page. Only a browser reload restored interactivity.

## Root cause

Confirmed from the frozen page: the stuck element is the create menu's own MUI Modal (`MuiPopover-root MuiMenu-root MuiModal-root`), left mounted after dismiss with `visibility: visible` and `pointer-events: auto`, covering the viewport. Its backdrop had faded to `opacity: 0` (invisible) but never unmounted.

When the menu closes, a re-render lands while it is mid-exit and cancels react-transition-group's `onExited` callback ([mui/material-ui#32286](https://github.com/mui/material-ui/issues/32286)). The Modal never reaches its `exited` state, so MUI never unmounts it or hides it, and the invisible full-screen backdrop keeps intercepting every click. The DMP integration is what triggers it: it adds async fetches that re-render the open menu. It is production-only (`-DgenerateReactDist`) because a dev build's StrictMode double-invoke masks the timing.

## Fix

`Sidebar.tsx`: make the create menu click-through (`pointer-events: none`) whenever it is closed, so a lingering closed Modal can no longer block the page. `pointer-events` is inherited and MUI sets it on neither backdrop nor paper, so the single root rule covers both. The open menu is unaffected.

We cannot make the upstream `onExited` fire reliably, so this neutralizes the symptom rather than the react-transition-group bug.

## Testing

- New browser-mode test `Sidebar.freeze.spec.tsx` guarding the deterministic half: the OPEN menu (and its DMP option) stays interactive, so the condition can never be inverted onto the open state.
- The freeze itself was reproduced and the fix verified manually under a production React build (`vitest --mode production` with a `requestAnimationFrame` re-render loop colliding with the menu exit): without the fix the stuck Modal intercepts clicks at screen centre; with it, the page stays clickable. The race is too timing-dependent to reproduce deterministically in CI, so it is documented rather than automated.
- Existing `Sidebar.test.tsx` (jsdom) passes, `pnpm tsc` and `pnpm lint` clean.

### DIFF
--- a/src/main/webapp/ui/src/eln/gallery/components/Sidebar.freeze.spec.tsx
+++ b/src/main/webapp/ui/src/eln/gallery/components/Sidebar.freeze.spec.tsx
@@ -1,0 +1,75 @@
+import { cleanup, render } from "@testing-library/react";
+import { HttpResponse, http } from "msw";
+import { afterEach, beforeEach, describe, expect, test } from "vitest";
+import { page } from "vitest/browser";
+import { worker } from "@/__tests__/browserSetup";
+import { galleryAppShellHandlers } from "@/__tests__/mocks/galleryMocks";
+import { DefaultSidebar } from "./Sidebar.story";
+
+/*
+ * PRT-1118. With a DMP integration enabled, dismissing the Gallery create menu
+ * froze the page: a re-render landing while the menu was mid-exit cancelled
+ * react-transition-group's `onExited` (mui/material-ui#32286), so the menu's
+ * Modal never unmounted and its invisible (opacity-0) backdrop kept intercepting
+ * every click. The fix (Sidebar.tsx) makes the closed menu click-through.
+ *
+ * The freeze only manifests under a production React build (`-DgenerateReactDist`,
+ * not `-DreactDevMode`, which StrictMode-masks it) and the underlying race is too
+ * timing-dependent to reproduce deterministically here, so there is no automated
+ * test for the frozen state. This spec guards the deterministic half in the normal
+ * browser suite: the OPEN menu must stay interactive (the pointer-events condition
+ * must never be inverted onto the open state).
+ */
+
+function integrationInfoHandlers() {
+  const info = (name: string, o: { available?: boolean; enabled?: boolean } = {}) => ({
+    data: {
+      name,
+      displayName: name,
+      available: o.available ?? false,
+      enabled: o.enabled ?? false,
+      oauthConnected: false,
+      options: {},
+    },
+    error: null,
+    success: true,
+    errorMsg: null,
+  });
+  return [
+    http.get("/integration/integrationInfo", ({ request }) => {
+      const name = new URL(request.url).searchParams.get("name") ?? "";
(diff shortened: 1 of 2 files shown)


nhanlon2:
@greptileai

nhanlon2:
@greptileai

nhanlon2:
@greptileai

tchoi-rs (changes requested):
Going by the comments on the actual issue, it looks like this issue was likely fixed in Material UI v9.1.0.

Could you try bumping the package version without the workaround and see if that fixes the issue?

nhanlon2:
> Going by the comments on the actual issue, it looks like this issue was likely fixed in Material UI v9.1.0.
> 
> Could you try bumping the package version without the workaround and see if that fixes the issue?

This is our current and its already on 9.2.0, which is the latest apparently. 

pnpm list "@mui/*"  
Legend: production dependency, optional only, dev only

rspace-web /Users/neilhanlon/projects/rspace-os-fork/rspace-web/.claude/worktrees/PRT-1118-create-button-unresponsive-with-DMPs (PRIVATE)
│
│   dependencies:
├── @mui/icons-material@9.2.0
├── @mui/material@9.2.0
├── @mui/styled-engine@9.1.1
├── @mui/system@9.2.0
├── @mui/x-chat@9.0.0-alpha.12
├── @mui/x-data-grid@9.8.0
├── @mui/x-date-pickers@9.8.0
└── @mui/x-tree-view@9.8.0

nhanlon2:
@tchoi-rs thanks for the pointer. I looked into the version-bump route and it does not resolve this one:

- We are already on the latest `@mui/material` (**9.2.0**). The published 9.x line ends at 9.2.0, and that is newer than the 9.1.0 mentioned, so there is no version to bump up to.
- [mui/material-ui#32286](https://github.com/mui/material-ui/issues/32286) is still **open** and labelled as blocked by the upstream react-transition-group bug ([reactjs/react-transition-group#817](https://github.com/reactjs/react-transition-group/issues/817)). It is fixed in no MUI release, so a version change in either direction would not include a fix.
- The freeze reproduces on 9.2.0 in a production build. From the frozen page the stuck element is the create menu's own Modal (`MuiPopover-root MuiMenu-root MuiModal-root`), left mounted after dismiss with `pointer-events: auto` and its backdrop faded to `opacity: 0`, so it silently intercepts every click. `onExited` never fires, so MUI never unmounts or hides it.

Given that, this fix keeps the create menu but makes it click-through (`pointer-events: none`) whenever it is closed, so a lingering Modal can no longer block the page. It is scoped to the closed state only; the open menu is unchanged.

Happy to revisit and drop the workaround if MUI or react-transition-group ship an actual fix for #32286.

tchoi-rs (approved):

PR #978, opened 2026-07-28
RSDEV-1220: Map Instrument template to b2inst

rs-nicof:
## Related PR ##
https://github.com/rspace-os/rspace-core-model/pull/83

## Description ##

Maps a concrete Instrument's custom fields into the B2INST (PIDINST) draft-record metadata, so registering an instrument PID sends real instrument metadata instead of just a name and owner.

A field participates when **both** its trimmed, case-insensitive name **and** its field type match a field of the default `Instrument (PIDINST 1.0)` template. Mapped properties: `Name`, `SchemaVersion`, `Description`, `Owner`, `Manufacturer`, `Model`, `InstrumentType`, `Date` (commissioned / decommissioned), `MeasuredVariable`, `LandingPage`, `AlternateIdentifier`.

Alongside the mapping, this PR also:

- **Adds the `Date` property to the B2INST metadata model.** Requires the matching rspace-core-model change (`B2instDate` + `List<B2instDate> date`, plus a new `PROVIDER_URL` identifier property).
- **Fixes publishing, which was completely broken.** Submitting a record for community review returned `415 Invalid 'Content-Type' header. Expected one of: application/json`. `RestTemplate` posted `HttpEntity.EMPTY`, and `HttpURLConnection` then defaults the content type to `application/x-www-form-urlencoded`, which Invenio rejects. Now sends an explicit empty JSON body. Verified end to end against `b2inst-test.gwdg.de`: the record reaches `submitted` and the review request opens against the configured community.
- **Stops the Inventory page crashing after publish.** B2INST reports the Invenio review status (`submitted`, `accepted`, `declined`, `cancelled`, `expired`), none of which are DataCite states. The state label used an exhaustive `match` with no fallback, so it threw during render and the error boundary blanked the whole Identifiers section. `StateInfo` had a second `throw` for the same reason.
- **Shows the identifier as a link in every state.** Previously the value was only rendered for `draft`; any other state rendered `publicUrl`, which is null for B2INST until a curator accepts, producing an empty `<a href="">`. The cell now keys off the URLs available, not the state.
- **Fills a blank `Landing page` on save.** An instrument shaped by the PIDINST template always ends up with a landing page: if the user leaves the field empty, saving fills it with the instrument's own `/globalId/` address, so the field and the registered metadata agree.
- **Disables Publish while a community review is open**, with a tooltip, since re-submitting would be rejected.

## Design decisions

- **Participation is by name + type conformance, never template lineage.** Any instrument carrying conforming fields is mapped, however it was created. A field matching by name but not by type is ignored — the type check is load-bearing: B2INST answers an opaque HTTP 500 when a `Date` value is unparseable, so a text field named "Commissioned" mapped blindly would break registration with no useful message.
- **`Date` wire format** is `{"Date": "...", "dateType": "Commissioned"|"DeCommissioned"}`. The EUDAT docs do not specify the inner keys; these were taken from production `b2inst.gwdg.de` records and confirmed against the test instance. Plain `yyyy-MM-dd` values are accepted and persist unchanged, so no normalisation. Note `DeCommissioned` deliberately differs from the field-name spelling, per the PIDINST vocabulary.
- **Three template fields are deliberately mapped to nothing:** *Measurement technique*, *Calibration* and *Last calibrated*. PIDINST has no property that fits a documented-by link, a calibration link, or a calibration date. Encoding them as `MeasuredVariable` narrative strings was implemented, then rejected as an abuse of a property specified for plain variable names. They stay on the template as **documentation-only fields**: ordinary fields users fill and read, which simply never leave RSpace. `MeasuredVariable` now carries only the *Measured quantity* content, verbatim. `DevDocs/adr/0005` records the rejected attempt and is marked superseded, because the underlying gap is unchanged.
- **`providerUrl` is a new field, deliberately not reused `publicUrl`.** `publicUrl` means "citable, publicly resolvable URL" and is rendered on the public identifier page as the DOI address. The B2INST record page (`/uploads/<rid>`, taken from the API's own `links.self_html` rather than string-built) redirects to a B2INST login even with our bearer token, so it is not public. Putting it in `publicUrl` would have published a login-gated link as the citable address the moment anyone maps `accepted` to `findable`.
- **The state label ends in a catch-all.** The server passes provider statuses through verbatim, so an unrecognised value now degrades to displaying the raw state rather than throwing out of render.

## Testing notes

Configure B2INST under System → Inventory settings: server URL, community id (`username`), token (`password`), enabled. The **community id is required for publishing**, not just registration.

Worth exercising by hand:

1. Create an instrument from `Instrument (PIDINST 1.0)`, fill the fields, register, and inspect the draft on B2INST.
2. Leave `Landing page` empty and save — the field should fill with `<serverUrl>/globalId/<globalId>`.
3. Publish, then reload the instrument page. Before this PR that page threw; it should now show `Submitted`, the identifier as a link, and Publish disabled.
4. An instrument with no fields at all should still register, carrying name, owner and the fallback landing page only.

Automated coverage: `RspaceToExternalProviderAdapterImplTest` (mapping, including that the three unmapped fields contribute nothing), `B2instConnectorImplTest` (now asserts `application/json` on both publish requests, which is what let the 415 through), `InventoryIdentifiersB2instApiControllerMVCIT` (end to end via a payload-capturing dummy connector, no live B2INST), `InstrumentsApiControllerMVCIT` (landing-page default on create and on save), and `Identifiers.test.tsx` (every B2INST state renders, link behaviour, Publish disabled).

> ⚠️ **Do not merge as-is:** `pom.xml` pins `rspace-core-model` to `RSDEV-1220-model-SNAPSHOT`. This needs the core-model change released and the version pinned to it first.


rs-nicof:
@greptileai

src/test/java/com/researchspace/api/v1/controller/InventoryIdentifiersB2instApiControllerMVCIT.java
```diff
@@ -0,0 +1,189 @@
...
+  public void setup() throws Exception {
+    ReflectionTestUtils.setField(identifierApiManager, "b2instConnector", b2instDummy);
+    super.setUp();
+    setB2instEnabled("true");
+  }
+
+  @After
+  public void teardown() throws Exception {
+    setB2instEnabled("false");
+  }
```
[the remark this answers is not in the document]
rs-nicof: Fixed in e8a532b. `setup()` now captures the real connector via `ReflectionTestUtils.getField` and `teardown()` puts it back, so the dummy cannot outlive this class in the cached context. Concrete case this avoids: `InventoryPublicApiControllerMVCIT` mutates the same `identifierApiManager` bean.

Verified by test-compile and Spotless only - this repo's automation is not permitted to execute `*MVCIT.java` tests, so the change has not been run.

rs-nicof:
@greptileai

src/main/webapp/ui/src/Inventory/components/Fields/Identifiers/Identifiers.tsx
```diff
@@ -297,7 +310,17 @@ export const IdentifiersList: ComponentType<IdentifiersListArgs> = observer(({ a
           values={{ link: identifierUrl || "" }}
         />
       );
-    throw new Error("Invalid state");
```
tchoi-rs, line 300: Are we supposed to allow arbitrary identifier states now?

tchoi-rs (changes requested):

src/main/webapp/ui/src/Inventory/components/Fields/Identifiers/Identifiers.tsx
rs-nicof: Good point. Long story short: different communities in b2inst can have different statuses description, the statuses are saved and checked by the backend anyways. The only thing we avoid here is the runtime error when the UI does not recognize a status :)

here the long story that Claude gave me:

> Yes, deliberately, and it isn't really new. The state is whatever the provider reported: the backend stores the providers' review status verbatim (InventoryIdentifierApiManagerImpl:600) and DataCite's state at 539, with no allowlist, and the frontend never validates it at runtime either — state: PublishingState on the API response is a type assertion, not a parse.
> 
> So an unrecognised state could always arrive. throw new Error("Invalid state") didn't prevent that, it just turned it into a render-time throw, and the error boundary then blanked the whole Identifiers section. That's the bug this PR is fixing: publishing to B2INST returns submitted, which is not a DataCite state, so the page died right after a successful publish.
> 
> Now an unmodelled state shows its raw value as the label with no explanatory paragraph, which is the graceful degradation. Covered by a state this frontend does not model degrades to the raw value instead of throwing in Identifiers.test.tsx.
> 
> I kept PublishingState as a closed union rather than widening it to string, so typos in comparisons like state === "submited" still fail to compile. Happy to revisit if you'd rather the type mirrored the runtime.

tchoi-rs: In that case, is it possible to derive, via any information available to the frontend at the moment, whether we're in pidinst mode or normal mode (since in normal mode, `identiferState` should be a known value)?

Also, it might be helpful to display to the user the exact state anyway if we don't have a translation for it.

rs-nicof: We cannot know from the UI what workflow we are, unfortunately.
We are displaying the state to the user but the only thing we are not doing is the identiferState check on the UI side.

tchoi-rs (approved):

rs-nicof:
new AWS built: https://rsdev-1220-map-template-to-b2inst-1038e470-26.researchspace.com

nhanlon2 (approved):

PR #981, opened 2026-07-28
Migrate all tests to JUnit Jupiter and remove JUnit 4 entirely

tchoi-rs:
-- AI writing below --

## Description ##

- Migrates the Java test suite from JUnit 4 to JUnit Jupiter 6 across 896 test files.
- Removes JUnit 4, the Vintage engine, and Jupiter migration support. A remaining JUnit 4 import will now fail compilation instead of producing a test that silently does not run.
- Replaces JUnit 4 runners, rules, lifecycle annotations, assertions, parameterized tests, categories, and custom system-property conditions with Jupiter equivalents.
- Renames test classes that did not match Surefire discovery patterns.
- Updates the test parent to `rspace-parent:3.1.0-junit6-r2-snapshot`.
- Documents the new test naming, Spring context, controller test, lifecycle, and environment-gating conventions in `AGENTS.md`.

This PR has no UI changes.

## Design decisions

- `@WithSpringContext` replaces direct use of Spring's JUnit runner for tests that need a Spring context but do not need the transactional test base classes.
- `@EnabledIfSystemProperty(matches = ".*")` replaces the old "property is defined" runner. Tests that previously required the value `true` still use `matches = "true"`.
- Database cleanup state is keyed by test class so a failed Spring context cannot reuse a `JdbcTemplate` registered by an earlier class.
- `DatabaseCleanerLifecycle` centralizes cleaner registration and final cleanup for tests outside the normal real-transaction base class.
- Test names now follow Surefire's `Test*`, `*Test`, `*Tests`, `*TestCase`, or `*IT` conventions.

## Files requiring human review

- `pom.xml`
  - Uses the pre-release `3.1.0-junit6-r2-snapshot` parent.
  - Removes Vintage and migration-support dependencies.
  - Confirm the parent version and managed dependency changes are ready before merge.

- `src/test/java/com/researchspace/testutils/DatabaseCleaner.java`
- `src/test/java/com/researchspace/testutils/DatabaseCleanerLifecycle.java`
- `src/test/java/com/researchspace/testutils/RealTransactionSpringTestBase.java`
- `src/test/java/com/researchspace/api/v1/controller/FilesAPIHandlerCachingTest.java`
  - Review the per-class database cleanup lifecycle, especially behavior after context-loading failures and classes that never reach `@BeforeEach`.

- `src/test/java/com/researchspace/integrations/omero/service/OmeroServiceNightlyTest.java`
- `src/test/java/com/researchspace/service/impl/BioPortalOntologiesServiceNightlyTest.java`
- `src/test/java/com/researchspace/testutils/TestRunnerController.java`
  - Review the conversion from the custom conditional runner to Jupiter system-property conditions.

- `src/test/java/com/researchspace/dao/hibernate/SandboxTest.java`
  - This class was previously undiscoverable as `Sandbox.java`. Its four database tests will now run.

- `src/test/java/com/researchspace/files/service/ExternalFileServiceTest.java`
  - The committed migration removes the old class-level `@Ignore`, so this test can attempt to load an Egnyte-configured context. Confirm that enabling it is intended.

- `src/test/java/com/researchspace/files/service/FileStoreConfiguredForExternalFS.java`
  - The committed changes delete this previously undiscoverable two-test Egnyte class. Confirm that the coverage is obsolete.

- `src/test/java/com/researchspace/webapp/controller/ProdProfileSignupControllerMVCIT.java`
- `src/test/java/com/researchspace/webapp/integrations/protocolsio/PIOComponentDiscoverabilityIT.java`
  - Review the changed Spring and controller test setup for these environment-gated tests.


src/test/java/com/researchspace/files/service/ExternalFileServiceTest.java
```diff
@@ -29,9 +29,10 @@
  * test exceptional cases;
  */
 @EgnyteTestConfig
-@Ignore
 public class ExternalFileServiceTest extends SpringTransactionalTest {
```
rs-nicof, line 32: The class-level `@Ignore` was removed here without a `@Disabled` replacement. It is the only one of 34 that did not convert, so it reads as accidental rather than deliberate.

It matters more than most: `@EgnyteTestConfig` sets `rs.filestore=EGNYTE`, which swaps the NOOP external file service for the real `ExternalFileServiceImpl`, and `egnyte.accessToken` has no value anywhere in the repo. The class is named `*Test`, so it now runs in the `test` phase on every full build and tries to reach a real Egnyte store, while also writing into the shared test file store (DB rows roll back, the written bytes do not).

Either restore `@Disabled` with a reason, or say in the description that enabling it is intentional and was verified green. Worth itemising the nine newly enabled tests either way.

rs-nicof (commented):
Reviewed the migration with several parallel passes: test-infrastructure layering, Spring transaction semantics, build/CI, and a systematic sweep for the usual automated-migration hazards.

The mechanical half came back clean, which is the main thing worth saying: no assertion argument-order flips, no losses across 159 `@Test(expected=...)` conversions, every `@EnabledIfSystemProperty` hoist preserves its original gating, all `@BeforeAll`/`@AfterAll` are static, and no JUnit 4 artefact survives anywhere in `src/`. `WithSpringContext` is also the right call, and its javadoc reasoning checks out against spring-test 6.2.x.

5 issues below. The `ExternalFileServiceTest` one is the only behavioural surprise, and the `pom.xml` one reaches outside this repo.

pom.xml
```diff
@@ -13,7 +13,7 @@
   <parent>
     <groupId>com.github.rspace-os</groupId>
     <artifactId>rspace-parent</artifactId>
-    <version>3.1.0</version>
+    <version>3.1.0-junit6-r2-snapshot</version>
```
rs-nicof: Two things here, and the second one leaves this repo.

The pin is a pre-release tag, so a real release wants cutting before this merges to `main`.

More importantly, that tag does not only bump versions: it also removes `junit:junit`, `junit-vintage-engine` and `junit-jupiter-migrationsupport` from the shared parent's `dependencyManagement`. Seven sibling repos declare those with no `<version>` and rely on the parent to supply it: `rspace-egnyte`, `rspace-core-model`, and the dataverse, dryad, figshare, zenodo and snapgene adapters. Each will fail with `'dependencies.dependency.version' is missing` the moment it bumps to this parent.

Cheapest fix is to keep the three `dependencyManagement` entries in the parent release, since they cost nothing where they are unused, and let each child drop its own declarations the way this PR does.

src/test/java/com/researchspace/integrations/omero/service/OmeroServiceNightlyTest.java
```diff
@@ -25,6 +25,7 @@
  * Uses the real public Omero repository hosted by Dundee University. Compares actual json responses
  * to the canned responses collected on 1st June 2023
  */
+@EnabledIfSystemProperty(named = "nightly", matches = ".*")
```
rs-nicof: `matches` widened from `"true"` to `".*"` when the gate moved to class level, which flips its meaning: `-Dnightly=false` now *enables* these tests instead of skipping them, and they call the live server at `idr.openmicroscopy.org`. `.*` is the right equivalent where it replaces `@RunIfSystemPropertyDefined`, which only checked that the property was defined, but these two were already Jupiter and required the literal value.

```suggestion
@EnabledIfSystemProperty(named = "nightly", matches = "true")
```

src/test/java/com/researchspace/service/impl/BioPortalOntologiesServiceNightlyTest.java
```diff
@@ -12,14 +12,14 @@
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.EnabledIfSystemProperty;
 
+@EnabledIfSystemProperty(named = "nightly", matches = ".*")
```
rs-nicof: Same widening as in `OmeroServiceNightlyTest`: `"true"` to `".*"` means `-Dnightly=false` now enables this class rather than skipping it.

```suggestion
@EnabledIfSystemProperty(named = "nightly", matches = "true")
```

src/test/java/com/researchspace/testutils/DatabaseCleaner.java
```diff
@@ -21,6 +21,13 @@ public void setDataSource(DataSource dataSource) {
 
   /** Manually deletes records and folders from tables. */
   public static void cleanUp() {
+    if (jdbcTemplate == null) {
```
rs-nicof, line 23: The guard is worth having, but it only covers part of the case the comment describes. `jdbcTemplate` is `public static` (line 15) and is set by `@Autowired setDataSource`, so once any test class in a surefire fork has loaded a context it stays non-null for the rest of that JVM.

So a *later* class whose context fails to load still reaches `@AfterAll`, finds a stale non-null template, and runs the full ~40-table delete against the shared database, with the real Spring startup error buried behind whatever the deletes do. Clearing the field in `@AfterAll`, or guarding on something per-class, closes that.

src/test/java/com/researchspace/files/service/ExternalFileServiceTest.java
tchoi-rs: I think we should now have an Egnyte mock setup? @nebay-abraha

nebay-abraha: @tchoi-rs  we do [here](https://github.com/rspace-os/rspace-web/tree/3e7ab07aa2ce4df3f19679e312b1d04fec6342ee/src/main/webapp/ui/src/modules/egnyte/__tests__)

tchoi-rs: Just checked again and the Egynte test doesn't actually need the real credentials.

src/test/java/com/researchspace/testutils/DatabaseCleaner.java
tchoi-rs: This is now fixed by storing the `jdbcTemplate` per test, so the cleanups of one test don't interefer with another.

rs-nicof (approved):

rs-nicof (approved):

PR #985, opened 2026-07-29
Add JSON i18n infrastructure

tchoi-rs:
-- AI writing below --

## Summary

Introduces the infrastructure required to use the frontend i18next JSON catalogues as the shared source of translations for Spring, Velocity, JSPs, and legacy JavaScript.

This PR:

- Replaces `ReloadableResourceBundleMessageSource` with `JsonMessageSource`.
- Migrates the existing `.properties` bundles into namespaced `server.*.json` catalogues.
- Adds locale-bound helpers for backend code and Velocity templates.
- Adds locale-aware list and date formatting.
- Exposes message-key helpers to JSP tag libraries.
- Builds a blocking legacy JavaScript catalogue through Vite.
- Adds catalogue validation and message-source tests.
- Documents the new extraction, typing, and linting workflow.
- Adds `deployment.experimental.locale` as an experimental system-wide BCP 47 locale override; blank or invalid values fall back to `en-US`.

This is the base of a three-PR stack. It is followed by the mechanical phrase migration and then the behavior-sensitive changes.

## Reviewer focus

Please spend most review time on:

1. **`JsonMessageSource` compatibility**
   - Namespace and flattened-key resolution.
   - ICU4J formatting and apostrophe behavior.
   - Missing-key behavior and locale fallback.
   - Validation-message interpolation, including `${validatedValue}` expressions.

2. **Catalogue loading and wiring**
   - Spring now reads the JSON catalogues instead of `.properties` bundles.
   - JSP, Velocity, and legacy JavaScript consumers must all see the same values.
   - Confirm the removed resource-bundle configuration has no remaining consumers.

3. **Legacy JavaScript bundle**
   - The catalogue must load as a classic blocking script before legacy callers use `RS.msg`.
   - Review Vite development and production behavior, asset deduplication, and script ordering.

4. **Locale-sensitive helpers**
   - `UserLocaleService`, `LocaleBoundMessages`, `ListFormatUtils`, and `LocaleAwareDateTool`.
   - Confirm their fallback behavior is suitable outside an HTTP request.

5. **Catalogue migration integrity**
   - Existing property keys and values should have equivalent JSON entries.
   - The generated `resources.d.ts` itself does not need line-by-line review.

The PyRAT component relocation is largely structural, but it is worth confirming that its TinyMCE bootstrap and dialog behavior remain unchanged.

## Testing

- TypeScript compilation
- Biome
- i18n catalogue lint
- 22 focused Java infrastructure and catalogue tests


src/main/java/com/researchspace/service/CommunicationStatusMessages.java
```diff
@@ -0,0 +1,19 @@
+package com.researchspace.service;
+
+import com.researchspace.model.comms.CommunicationStatus;
+
+public final class CommunicationStatusMessages {
+
+  private CommunicationStatusMessages() {}
+
+  public static String keyFor(CommunicationStatus status) {
```
nhanlon2, line 9: NotificationStatusMessages has a test class; this doesnt.

nhanlon2 (changes requested):
Agree with copilot's issue regarding dev mode and legacyI18n.ts emitting module export import statements that will break and throw SyntaxError: Unexpected token'.  Also could add tests for CommunicationStatusMessages

tchoi-rs:
I've redone the i18n JavaScript code so it's simpler and loads less catalogue code (which also makes the issue go away)

nhanlon2 (approved):

mKowalski256 (approved):
Looks good and works fine on test AWS build. 

I managed to build on localhost (Windows), but had to update the `vite.config.ts` by importing and applying normalizePath within that file (I've attached the git diff). I think it would be good to include that, Windows OS is not a supported ops environment, but we could keep it as supported dev environment.

[vite_diff.txt](https://github.com/user-attachments/files/30670628/vite_diff.txt)

PR #989, opened 2026-07-29
Rsdev 1250 gallery revision history

nhanlon2:
## Description ##

Adds a version history to Gallery items, and a shareable URL that shows an item as it was at one past version.

- **Version history dialog.** The Actions menu gains "View Version History", listing an item's versions newest first with each version's own filename, modification date, editor and size. Backed by a new `GET /gallery/ajax/versionHistory/{mediaFileId}`.
- **Pinned version view.** `/gallery/item/{id}/{version}` shows the item at that version, read-only, with the item selected. Each row of the dialog links to it, so what a user lands on is also a URL they can copy into a document or a paper.
- **Read-only means read-only.** Rename, Delete, Move, Duplicate, Export, Upload-new-version and Edit are all refused on a pinned view; Download and the version history itself still work. The item's grid tile carries the version with the same clock icon a version-pinned Inventory link uses, and the InfoPanel says which version is showing and links back to the latest.

Frontend: 2338 unit tests pass. Backend: 2483 pure unit tests pass. `tsc`, `lint`, `i18n:lint` and `spotless` clean.

## Per item type

Whether "View Version History" is active is decided entirely by `GalleryFile.canViewVersionHistory` plus the Actions menu's single-selection rule. Nothing keys off the gallery section, so it reduces to: is it a folder, is it stored locally, and is it a media file.

| Item type | Backing class / `type` | Menu item |
| --- | --- | --- |
| Images | `EcatImage` / `"Image"` | **Active** |
| Audio | `EcatAudio` / `"Audio"` | **Active** |
| Video | `EcatVideo` / `"Video"` | **Active** |
| Documents (incl. Office/WOPI/Collabora files) | `EcatDocumentFile` / `"Document"` | **Active** |
| Chemistry | `EcatChemistryFile` / `"Chemistry"` | **Active** |
| DMPs | `EcatDocumentFile` / `"DMPs"` | **Active** |
| Miscellaneous | `EcatDocumentFile` / `"Miscellaneous"` | **Active** |
| Exports (PdfDocuments) | `EcatDocumentFile` / `"PdfDocuments"` | **Active** |
| Extensionless local file | any of the above | **Active** (deliberate: unlike `canUploadNewVersion`, no extension is required) |
| Pinned historical version being viewed | `HistoricalGalleryFile` wrapping a local file | **Active** (delegates to the wrapped file) |
| Snippets | `Snippet`, a `Record` rather than an `EcatMediaFile` / `"Snippet"` | Disabled: "Snippets do not have a version history." |
| Ordinary folder in any section | `"Folder"` | Disabled: "Folders do not have a version history." |
| System folder (Gallery section roots, `SNIPPETS`, shared folders) | `"System Folder"` | Disabled, same reason |
| Filestore root (NetworkFiles) | `Filestore` | Disabled: "Filestores do not have a version history." |
| File on a filestore | `RemoteFile` | Disabled: "Files stored in filestores do not have a version history." |
| Folder on a filestore | `RemoteFile` with `isFolder` | Disabled, same reason |
| Nothing selected, or 2+ selected | any | Disabled: "Only one item's version history may be viewed at once." |

Snippets are the one row worth a note. A snippet is a `Record`, so no path ever gives it a new version, and the endpoint audits `EcatMediaFile`, which a snippet is not. `canUploadNewVersion` already excluded them only incidentally, because a snippet carries no extension and that predicate requires one; `canViewVersionHistory` deliberately drops the extension check so extensionless versioned files still show a history, so it has to refuse snippets explicitly.

## Design decisions

**The audit row is the source of truth for everything about a version except the item's identity and its inbound references.** This is the one thing to keep in mind when reading the diff; three separate defects during development came from delegating a field to the live item that actually varies per version.

- `name`, `description`, `size`, `modificationDate`, `thumbnailUrl` and `downloadHref` all differ per version. Uploading a new version can replace the file with a differently named one, so the live item's name says nothing about what version 1 was called. `name` and `description` come from the audit row via the new DTO; both live on the audited `EditInfo` embeddable, so Envers records them per revision.
- `globalId` deliberately stays unversioned (`GL42`, never `GL42v2`). The ELN linked-documents and Inventory referencing lookups read it, and neither records which version a reference was made against. The InfoPanel renders `GL42v1` for display only, and a notice above both reference lists says they belong to the item rather than the version.

**A revision is not a version.** Several audit revisions can share one version, because not every recorded change bumps the counter. The dialog shows one row per version using the newest revision of each, via a helper shared with the Inventory version history rather than reimplemented.

**Refusals live on the object, not in the callers.** A past version is modelled as a decorator (`HistoricalGalleryFile`) that returns `Result.Error` from the `can*` predicates, so the Actions menu disables the right things without a single new conditional. `canBeEdited` had to be added to the interface: edit availability was previously decided solely by which editor applies to the file's type, so a past version stayed editable.

**No thumbnail or preview endpoint is version-aware,** so a pinned view shows a stock type icon or no preview rather than live content under an older version's label. Images and PDFs are fine, resolving through the version-aware `downloadHref`. Aspose could be made correct later, since it already accepts a `revisionId`.

**Version history is not part of what publishing a document shares.** `GalleryController` is mapped at `/public/publicView/gallery` as well as `/gallery`, Shiro treats `/public/**` as `anon`, and viewing a published document logs a real session in as the anonymous guest, which holds read permission on media linked from that document. So the ordinary permission check would have passed for it and disclosed every past filename and description plus the full name of every user who had edited the item. The guest is refused outright. See `DevDocs/adr/0003`.

Every new route needs a matching rule in `urlrewrite.xml`, which forwards it to the SPA shell before Spring MVC sees it. The rules are anchored regexes, so a new URL shape 404s until one is added.

Longer rationale in `DevDocs/adr/0003-gallery-version-history-endpoint.md` (why this has its own endpoint rather than extending the ELN document one) and `DevDocs/adr/0004-gallery-pinned-version-links.md` (why a route rather than a versioned Global ID), plus `src/main/webapp/ui/src/eln/gallery/README.md`.

## Testing notes

Manual check needs an item with more than one version: upload a file to the Gallery, then upload a **different** file over it via Actions → "Upload new version", ideally with a different filename so the per-version fields are visibly distinct.

1. Actions → "View Version History". Each row should show that version's own filename, not the current one.
2. Click `v1`. The URL becomes `/gallery/item/<id>/1`; the tile shows `v1` with a clock icon; the image and filename shown are version 1's, not the live one's.
3. On that view, Rename / Delete / Move / Duplicate / Export / Upload-new-version / Edit are all disabled with a reason, and the name and description fields are not editable. Download gives you version 1's bytes.
4. "View the latest version" returns you to the editable item, with no stale version badge or notice left behind.
5. Paste `/gallery/item/<id>/1` into a fresh tab. It should load the pinned view directly rather than 404.
6. A version the item does not have, e.g. `/gallery/item/<id>/99`, reports an error rather than falling back to the live item. A version equal to the live one redirects to `/gallery/item/<id>`.

Worth exercising with an Office document if Collabora or Office Online is enabled, since that is where edit access was previously reachable from a pinned view.

Things reviewers should know:

- `urlrewrite.xml` rules cannot be covered by any MVCIT, because MockMvc bypasses the servlet filter chain. Covering the new rule needs a Playwright spec.

Known follow-ups, deliberately not in scope here: moving the audit-to-DTO assembly into a `*Manager` so the Envers snapshots are read inside a live session (safe today only because every field read is a scalar or the `EditInfo` embeddable); the audit query is unbounded and hydrates full entities; and the `can*` predicate messages are English literals, which wants one sweep across the file rather than translating only the new lines.


nhanlon2:
@greptileai

nhanlon2:
Both Copilot comments addressed in e51d7103b.

- **Externalised the refusal message.** Confirmed it is genuinely user-facing rather than internal: `ControllerExceptionHandler` adds the exception message to the model as `exceptionMessage` and both the ajax and non-ajax error views render it. Now `error.authorization.versionHistory.loginRequired` in `ApplicationResources.properties`, asserted in the unit test via an injected `MessageSourceUtils`.
- **Fixed the full-name memo for a deleted user.** Correct that `computeIfAbsent` will not store a null, so the memo stopped memoising exactly the case it exists for; replaced with a `containsKey` memo that caches the null too. One correction to the comment though: the DAO was not being re-queried, because `getFullNameByUsername` is `@Cacheable` and Spring caches null return values. So this was the memo not doing what its own comment claimed, rather than a measured cost.

@greptileai

nhanlon2:
Ordering comment addressed in a0e9000ee: revisions are now sorted by revision id in the controller, so the DTO's documented "newest revision id last" holds regardless of what the Envers query returns. Sorted there rather than adding `addOrder` to `getRevisionsForObject`, since that DAO method is shared and this is the only caller documenting an order. The pre-existing order assertion fed already-sorted input so could not have caught it; the new test feeds newest-first.

@greptileai

nhanlon2:
Latest Copilot pass raised two low-confidence suppressed comments on a0e9000ee. Both examined, neither actioned, reasons below so the decision is on the record.

**`GalleryVersionHistoryTest:70` test name vs assertions.** The comment says the name "listsEveryRevisionNewestVersionLast" contradicts assertions that verify "oldest revision/version first". Those are the same ordering: the test asserts revision 10/version 1 at index 0 and revision 20/version 2 at index 1, so the newest is last. Name is accurate, no change.

**`index.tsx:507` stale `fileName` during navigation.** Correct that the effect resets `resolution` but not `fileName`. It does not produce a visible stale title, though: the loading branch renders a plain `t("landingPage.loading")` string with no `WholePage`, so the `title` prop is never evaluated while loading, and on success `setFileName` and `setResolution` run in the same synchronous continuation after the await, so React batches them and the success render always sees the new name. The error and redirect branches render no title either. Leaving as is rather than adding a reset that changes nothing observable.

That completes the automated remediation loop: three cycles, two commits (e51d7103b, a0e9000ee), all Copilot review threads resolved.

Note for whoever picks this up: Greptile never responded across three `@greptileai` triggers and about 90 minutes of polling, so this PR has effectively only had Copilot's automated review. Worth checking the app is still installed rather than reading the silence as a clean pass.

src/main/webapp/ui/src/modules/common/i18n/locales/en-US/gallery.json
```diff
@@ -88,6 +88,23 @@
     "renamePrompt": "Please give a new name for {name}.",
     "saveAsNewImage": "Save as new image",
     "uploadNewVersion": "Upload New Version",
+    "versionHistory": {
+      "columns": {
+        "by": "By",
```
rs-fraser: It wasn't too clear to me whether this was Owner, which we usually display, or Modified By. Could we name it `Modified By` and the next line `Modification Date` which matches the document revision view naming.

src/main/webapp/ui/src/eln/gallery/components/ActionsMenu.tsx
```diff
@@ -808,6 +819,34 @@ function ActionsMenu({ refreshListing, section, folderId }: ActionsMenuArgs): Re
...
+              setVersionHistoryOpen(true);
+            }}
+            compact
+            disabled={versionHistoryAllowed.get().isError}
+            aria-haspopup="dialog"
+          />
+          {selection
+            .asSet()
+            .only.map((file) => (
+              <VersionHistoryDialog
```
rs-fraser, line 839: https://github.com/user-attachments/assets/f4861182-090d-45ef-beeb-d23c7a09eb70

When this is open and I try to tab with the keyboard I'd expect to switch between the link and the close button. However, the keyboard is still in focus on the Action Menu so tab closes the menu and dialog.

You can use `<EventBoundary>` like on line 883 to fix it.

rs-fraser (changes requested):
Looks good. Left 2 inline comments - a suggestion about wording and a minor bug with keyboard shortcuts.

While testing I also found and raised this issue https://researchspace.atlassian.net/browse/RSDEV-1269 - it's a pre-existing issue although made more prominent with the version links, so we may want to fix as part of the release.

tchoi-rs (commented):
Could you trim down the amount of comments in the PR?

Also, ChatGPT found a bug: The gallery version file isn't loaded as the first thing in the grid, which means when loading a version link, the actual file can be pushed off the grid by 20 other files, making the UI weird.

Prerequisite: Have one small local file named target.txt.

1. Open RSpace and log in as user1a.
2. Open Gallery → DOCUMENTS.
3. Set SORT → Modification Date → newest first.
4. Click CREATE → Upload Files.
5. Select target.txt and wait for it to appear.
6. Select target.txt.
7. Click ACTIONS → Upload New Version.
8. Select target.txt again and wait for version 2 to finish uploading.
9. With target.txt selected, click ACTIONS → View Version History.
10. Click v1 and confirm version 1 initially displays correctly.
11. Leave this tab open.
12. Open Gallery → DOCUMENTS in a second browser tab.
13. Select the original target.txt.
14. Click ACTIONS → Duplicate.
15. Wait for the duplicate to appear.
16. Repeat steps 13–15 another 19 times, producing 20 newer filler files.
17. Wait until the original and all 20 duplicates are present.
18. Return to the tab displaying version 1.
19. Reload the page.

Expected:
target.txt version 1 remains selected and its historical-version notice is displayed.

Actual:
- The URL still points to version 1.
- Only the 20 newer duplicates are initially displayed.
- The toolbar reports "No selection."
- The details panel reports "Nothing selected."
- The historical-version notice is absent.

20. Click LOAD MORE.

The original target then appears, becomes selected, and displays the version-1 notice.

tchoi-rs (changes requested):

src/main/webapp/ui/src/modules/common/i18n/locales/en-US/gallery.json
nhanlon2: Agree but that needs confirmation from @vaidap and @tilorspace because they want this UI to be compact

nhanlon2: fixed in d3780fa54

src/main/webapp/ui/src/eln/gallery/components/ActionsMenu.tsx
nhanlon2: fixed with 1ffcbc516 
Adding EventBoundary would cause a11y violation

nhanlon2:
@tchoi-rs trimmed comments in 388ccb050

nhanlon2:
@tchoi-rs what chatgpt has observed is how Gallery behaves for unversioned links  (/gallery/item/target.txt.id) as well - so the unversioned url to target.txt would do exactly the same - you can test this out to verify.


When urls link directly to gallery items (unversioned urls) the item can be pushed off the display by newer items and 'load more' has to be clicked before the item is actually shown. After that the items is also selected. Why would it be expected to be otherwise? 

I dont see how this is a bug, since it is the existing behaviour for direct urls to unversioned items, so the versioned links (/gallery/item/target.txt.id/version) are just behaving in the same way. How could an older gallery item (the target.txt) be expected to be displayed first, when the sorting is by modification date?

rs-fraser:
> @tchoi-rs what chatgpt has observed is how Gallery behaves for unversioned links (/gallery/item/target.txt.id) as well - so the unversioned url to target.txt would do exactly the same - you can test this out to verify.
> 
> When urls link directly to gallery items (unversioned urls) the item can be pushed off the display by newer items and 'load more' has to be clicked before the item is actually shown. After that the items is also selected. Why would it be expected to be otherwise?
> 
> I dont see how this is a bug, since it is the existing behaviour for direct urls to unversioned items, so the versioned links (/gallery/item/target.txt.id/version) are just behaving in the same way. How could an older gallery item (the target.txt) be expected to be displayed first, when the sorting is by modification date?

This is the same bug I mentioned and raised as https://researchspace.atlassian.net/browse/RSDEV-1269.

It is a pre-existing bug but becomes more prominent with this new feature. It is strange behaviour: currently, a link working and loading a gallery item is dependent on a) how many other items you have in that section of gallery and b) what your sort preferences are. A direct link to an item should always resolve regardless of where that item appears in search results.

nhanlon2:
> > @tchoi-rs what chatgpt has observed is how Gallery behaves for unversioned links (/gallery/item/target.txt.id) as well - so the unversioned url to target.txt would do exactly the same - you can test this out to verify.
> > When urls link directly to gallery items (unversioned urls) the item can be pushed off the display by newer items and 'load more' has to be clicked before the item is actually shown. After that the items is also selected. Why would it be expected to be otherwise?
> > I dont see how this is a bug, since it is the existing behaviour for direct urls to unversioned items, so the versioned links (/gallery/item/target.txt.id/version) are just behaving in the same way. How could an older gallery item (the target.txt) be expected to be displayed first, when the sorting is by modification date?
> 
> This is the same bug I mentioned and raised as https://researchspace.atlassian.net/browse/RSDEV-1269.
> 
> It is a pre-existing bug but becomes more prominent with this new feature. It is strange behaviour: currently, a link working and loading a gallery item is dependent on a) how many other items you have in that section of gallery and b) what your sort preferences are. A direct link to an item should always resolve regardless of where that item appears in search results.

All good but thats a new feature, to be agreed upon by product owner(s) and a ticket be written. None of this is meant to be handled during a code review.

rs-fraser:
> > > @tchoi-rs what chatgpt has observed is how Gallery behaves for unversioned links (/gallery/item/target.txt.id) as well - so the unversioned url to target.txt would do exactly the same - you can test this out to verify.
> > > When urls link directly to gallery items (unversioned urls) the item can be pushed off the display by newer items and 'load more' has to be clicked before the item is actually shown. After that the items is also selected. Why would it be expected to be otherwise?
> > > I dont see how this is a bug, since it is the existing behaviour for direct urls to unversioned items, so the versioned links (/gallery/item/target.txt.id/version) are just behaving in the same way. How could an older gallery item (the target.txt) be expected to be displayed first, when the sorting is by modification date?
> > 
> > 
> > This is the same bug I mentioned and raised as https://researchspace.atlassian.net/browse/RSDEV-1269.
> > It is a pre-existing bug but becomes more prominent with this new feature. It is strange behaviour: currently, a link working and loading a gallery item is dependent on a) how many other items you have in that section of gallery and b) what your sort preferences are. A direct link to an item should always resolve regardless of where that item appears in search results.
> 
> All good but thats a new feature, to be agreed upon by product owner(s) and a ticket be written. None of this is meant to be handled during a code review.

I agree, that's why I raised it as another ticket. I was responding to your part about "I don't see how this is a bug", and the reason I mentioned it initially is because it was found during code review and becomes more prominent with the new feature you're introducing

rs-fraser (approved):

PR #990, opened 2026-07-29
Upgrade JUnit Mockito strictness

tchoi-rs:
-- AI Writing below --

## Summary

Tightens Mockito strict-stubbing across the test suite. Shared setup now contains only stubs used by every test, while scenario-specific stubs live beside the tests that need them.

Strict stubbing also exposed several tests that passed without exercising the behavior described by their names. Those tests now assert the intended interactions directly.

This PR changes 83 files, all under `src/test/java`. It does not change production code or runtime behavior.

## What changed

- Removed `lenient()` calls and lenient class-level Mockito settings.
- Deleted unused stubs and moved conditional stubs into focused setup helpers.
- Added `MockitoExtension` to standalone tests that previously created unmanaged mocks.
- Replaced hand-written servlet and JSP mock state with Spring test doubles.
- Removed dead parameters from shared test fixtures.
- Fixed tests that previously passed because mocks returned default values.

## Files for human review

| File | Review focus |
| --- | --- |
| `src/test/java/com/researchspace/webapp/controller/IntegrationControllerTest.java` | The integration listing test now supplies the authenticated user. Previously the controller received `null`, the mock did not match, and the assertion passed on an empty result. |
| `src/test/java/com/researchspace/webapp/controller/WorkspacePermissionsDTOBuilderTest.java` | Media records and snippets now directly assert that move-permission evaluation short-circuits before `RecordManager.canMove`. This preserves the RSPAC-999 behavior check. |
| `src/test/java/com/researchspace/api/v1/controller/SysadminApiControllerTest.java` | The licence-unavailable disable test now starts with an enabled user and proves that disabling succeeds without consulting the licence service. |
| `src/test/java/com/researchspace/service/impl/IntegrationsHandlerTest.java` | A DSW URL-only edit now asserts that credential lookup, deletion, and saving are never attempted. |
| `src/test/java/com/researchspace/service/impl/RecordSharingManagerImplTest.java` | Permission parsing delegates to the real `PermissionUtils` mapping instead of duplicating it in a mock answer. |
| `src/test/java/com/axiope/webapp/taglib/AssetUrlTagTest.java` and `BundleTagTest.java` | These fixtures now use Spring servlet test doubles. Review request, application, writer, cache, development-mode, and HMR behavior. |
| `src/test/java/com/researchspace/webapp/controller/WorkspaceControllerTest.java` | Shared stubs were split by scenario, and the static custom-form menu cache is cleared between tests. |
| `src/test/java/com/researchspace/service/impl/MockFolderStructure.java`, `src/test/java/com/researchspace/testutils/FolderTestUtils.java`, and `src/test/java/com/researchspace/service/FolderOrganisationAndApiInboxFolderTest.java` | The unused external-folder parameter was removed. Callers now own any `FolderManager.getFolder` setup they require. |

The remaining changes are mostly mechanical removal or relocation of unused Mockito stubs.

## Verification

- 106 distinct focused tests passed across 11 affected test classes.
- The final 43-test DSW and sysadmin rerun passed after the strictness cleanup.
- `git diff --check` passes.
- Repository scan confirms no `lenient(...)` calls remain under `src/test/java`.

### DIFF
--- a/src/test/java/com/researchspace/api/v1/controller/SysadminApiControllerTest.java
+++ b/src/test/java/com/researchspace/api/v1/controller/SysadminApiControllerTest.java
@@ -12,7 +12,6 @@ import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyInt;
 import static org.mockito.Mockito.doThrow;
-import static org.mockito.Mockito.lenient;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
@@ -126,6 +125,7 @@ public class SysadminApiControllerTest extends JakartaValidatorTestJU5 {
     User toDelete = createAnyUserWithId();
     toDelete.setTempAccount(true);
     setCreationDate2yearsAgo(toDelete);
+    stubSuccessfulDeletion(sysadmin, toDelete);
     basicDeleteTempUserRequest(sysadmin, toDelete);
     assertDeletionManagerInvoked();
   }
@@ -161,6 +161,7 @@ public class SysadminApiControllerTest extends JakartaValidatorTestJU5 {
         .thenReturn(exportResult);
     mockServerUrl();
     // test delete
+    stubSuccessfulDeletion(sysadmin, toDelete);
     basicDeleteAnyUserRequest(sysadmin, toDelete);
     // then
     assertDeletionManagerInvoked();
@@ -223,12 +224,13 @@ public class SysadminApiControllerTest extends JakartaValidatorTestJU5 {
 
   private void setUpMocksForDeletingUser(User sysadmin2, User toDelete) {
     when(userMgr.get(toDelete.getId())).thenReturn(toDelete);
-    lenient()
-        .when(
-            userDelMgr.removeUser(
-                Mockito.eq(toDelete.getId()),
-                Mockito.any(UserDeletionPolicy.class),
-                Mockito.eq(sysadmin2)))
+  }
+
+  private void stubSuccessfulDeletion(User sysadmin2, User toDelete) {
+    when(userDelMgr.removeUser(
+            Mockito.eq(toDelete.getId()),
+            Mockito.any(UserDeletionPolicy.class),
+            Mockito.eq(sysadmin2)))
         .thenReturn(new ServiceOperationResult<User>(toDelete, Boolean.TRUE, "OK"));
   }
 
@@ -349,17 +351,18 @@ public class SysadminApiControllerTest extends JakartaValidatorTestJU5 {
     when(groupManager.removeGroupIfNoMemberLoggedInWithinOneYear(toDelete.getId(), sysadmin))
         .thenThrow(new IllegalStateException("internal: user 'alice' logged in 5 days ago"));
 
-    IllegalArgumentException ex =
+    String message =
         Assertions.assertThrows(
-            IllegalArgumentException.class,
-            () ->
-                controller.deleteGroupIfNoMemberLoggedInWithinOneYear(
-                    request, toDelete.getId(), sysadmin));
+                IllegalArgumentException.class,
+                () ->
+                    controller.deleteGroupIfNoMemberLoggedInWithinOneYear(
+                        request, toDelete.getId(), sysadmin))
+            .getMessage();
 
     // Generic message; no username leaked.
     Assertions.assertFalse(
-        ex.getMessage().contains("alice"), "response message must not leak internal usernames");
-    Assertions.assertTrue(ex.getMessage().contains("within the last year"));
+        message.contains("alice"), "response message must not leak internal usernames");
+    Assertions.assertTrue(message.contains("within the last year"));
     verify(auditService, never()).notify(Mockito.any(GenericEvent.class));
   }
 
@@ -507,8 +510,6 @@ public class SysadminApiControllerTest extends JakartaValidatorTestJU5 {
     userToEnable.setEnabled(false);
     userToEnable.addRole(Role.USER_ROLE);
     when(userMgr.get(userToEnable.getId())).thenReturn(userToEnable);
-    when(userMgr.save(userToEnable)).thenReturn(userToEnable);
-
     // when
     controller.enableUser(request, sysadmin, userToEnable.getId());
 
@@ -527,8 +528,6 @@ public class SysadminApiControllerTest extends JakartaValidatorTestJU5 {
     userToEnable.setEnabled(true);
     userToEnable.addRole(Role.USER_ROLE);
     when(userMgr.get(userToEnable.getId())).thenReturn(userToEnable);
-    lenient().when(userMgr.save(userToEnable)).thenReturn(userToEnable);
-
     // when
     controller.enableUser(request, sysadmin, userToEnable.getId());
 
@@ -545,7 +544,6 @@ public class SysadminApiControllerTest extends JakartaValidatorTestJU5 {
     userToEnable.setEnabled(false);
     userToEnable.addRole(Role.USER_ROLE);
     when(userMgr.get(userToEnable.getId())).thenReturn(userToEnable);
-    lenient().when(userMgr.save(userToEnable)).thenReturn(userToEnable);
     doThrow(new LicenseServerUnavailableException())
         .when(userEnablementUtils)
         .checkLicenseForUserInRole(anyInt(), any(Role.class));
@@ -590,8 +588,6 @@ public class SysadminApiControllerTest extends JakartaValidatorTestJU5 {
     disableToDisable.setEnabled(false);
     disableToDisable.addRole(Role.USER_ROLE);
     when(userMgr.get(disableToDisable.getId())).thenReturn(disableToDisable);
-    lenient().when(userMgr.save(disableToDisable)).thenReturn(disableToDisable);
-
     // when
     controller.disableUser(request, sysadmin, disableToDisable.getId());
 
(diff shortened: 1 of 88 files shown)


src/test/java/com/researchspace/webapp/controller/WorkspacePermissionsDTOBuilderTest.java
```diff
@@ -148,9 +144,6 @@ public void snippetsCantBeMovedOrExported() {
     final List<BaseRecord> records = toList(snip);
     parentFolder.addChild(snip, user);
     when(results.getResults()).thenReturn(records);
```
rs-nicof, line 147: Removing this stub costs the test the ability to fail. Production computes `move = !br.isMediaRecord() && !br.isSnippet() && recMgr.canMove(...)` (`WorkspacePermissionsDTOBuilder.java:89-92`), so for a snippet `&&` short-circuits before `canMove` is ever called, which is why the stub needed `lenient()`.

That is also what made it load-bearing. With `canMove` forced to `true`, deleting the `!br.isSnippet()` guard from production made `move` true and this test failed. Now `canMove` returns the unstubbed default `false`, so `move` stays `false` either way and `assertFalse(can(result, DOC_ID, PermissionType.SEND))` passes with the guard removed. The RSPAC-999 rule that snippets are never offered move or share is no longer tested.

Cleanest fix that stays strict-stubs clean is to assert the short-circuit directly:

```suggestion
    verify(recMger, never()).canMove(snip, parentFolder, user);
```

added alongside the existing assertions. Restoring the stub as `lenient()` with a comment explaining the short-circuit also works. The `EXPORT` half of the test is unaffected.

src/test/java/com/researchspace/webapp/controller/WorkspacePermissionsDTOBuilderTest.java
```diff
@@ -84,8 +82,6 @@ public void optionsForMediaRecords() throws IllegalAddChildOperation {
 
     parentFolder.addChild(media, user);
     Mockito.when(results.getResults()).thenReturn(records);
```
rs-nicof, line 84: Same mechanism as the snippet case below, for the `!br.isMediaRecord()` clause of the same expression. The removed `canMove(media, parentFolder, user) -> true` stub was what gave `assertFalse(can(result, DOC_ID, PermissionType.SEND))` its teeth; with it gone the default `false` satisfies the assertion whether or not the media-record guard still exists.

`verify(recMger, never()).canMove(media, parentFolder, user)` proves the short circuit and needs no lenient stub.

src/test/java/com/researchspace/service/impl/IntegrationsHandlerTest.java
```diff
@@ -644,11 +633,6 @@ public void testUpdateDSWAppConfigOptionUrl() {
     existingConnection.setAccessToken(origDswToken);
 
     when(appCfgMgr.findByAppConfigElementSetId(1l)).thenReturn(Optional.of(aces));
```
rs-nicof, line 627: With the `userConnectionManager.findByUserNameProviderName(...) -> Optional.of(existingConnection)` stub gone, this mock returns `Optional.empty()`, so `updateUserConnectionForMultipleOptionApp` falls through to `saveNewUserConnectionForMultipleOptionApp`, which returns early on the masked token. Neither a delete nor a save happens on either side of the guard, so both `times(0)` verifications below now pass for the wrong reason.

In other words the test is meant to prove that editing only the DSW server URL leaves the stored credential untouched, but it now passes because the test no longer pretends a credential exists at all. It would stay green if that guard broke.

Worth keeping the stub as `lenient()`; the comment further down already explains why it has to be present but unmatched. The 22 lines building `existingConnection` are otherwise dead setup.

src/test/java/com/researchspace/api/v1/controller/SysadminApiControllerTest.java
```diff
@@ -611,12 +607,6 @@ public void doNotRaiseExceptionDisablingUserAccountWhenNoLicense() throws Except
     userToEnable.setEnabled(false);
     userToEnable.addRole(Role.USER_ROLE);
     when(userMgr.get(userToEnable.getId())).thenReturn(userToEnable);
```
rs-nicof: This test no longer sets up the scenario its name describes. `performEnableUser` only consults `userEnablementUtils.checkLicenseForUserInRole` inside `if (enable)`, and the user here is created with `setEnabled(false)` and then disabled, so `enable != userToAmend.isEnabled()` is false and the method returns before any licence check. The setup is now equivalent to `disablingUserAccountThatWasAlreadyDisabled` just above.

So nothing in the suite proves a sysadmin can still switch off a compromised account while the licence server is unreachable.

The stub was already dead before this PR, which is why it needed `lenient()`. The change that makes the test match its name is `userToEnable.setEnabled(true)`, keeping the `doThrow`, so the disable actually changes state and the licence path is reached.

rs-nicof (commented):
Reviewed with three parallel passes: stub-removal fidelity, test-infrastructure layering, and a security-path sweep. Roughly 70 of the removed stubbings were traced back to their production call site.

The great majority of the 84 `lenient()` removals are correct, and several changes are real improvements: mocks replaced with real objects (`ApiSample`, `TextField`, Spring's `MockServletContext`), the mock-free tests split out of `ViteDevServerProxyServletMimeTypeTest` so the class could drop `@MockitoSettings(LENIENT)` wholesale, and `WorkspaceControllerTest` surfacing a genuine static-state leak that leniency had been papering over. `rg` confirms zero `lenient()`, `Strictness.LENIENT`, `@MockitoSettings`, `MockitoJUnitRunner.Silent` or `withSettings().lenient()` survive.

7 findings below. The theme worth attention is the first three: a `lenient()` stub on a short-circuited branch is sometimes the only thing that makes an assertion able to fail, so removing it leaves a test that passes whether or not the rule it is named after still exists. Mockito's own guidance treats `lenient()` as the sanctioned marker for exactly that case.

src/test/java/com/researchspace/service/impl/RecordSharingManagerImplTest.java
```diff
@@ -78,22 +78,22 @@ record = TestFactory.createAnyRecord(u);
...
-    lenient().when(permissnUtils.findBy(any(), any(), any(), any())).thenReturn(cbp);
-    lenient().when(permissnUtils.createFromString("WRITE")).thenReturn(PermissionType.WRITE);
-    lenient().when(permissnUtils.createFromString("READ")).thenReturn(PermissionType.READ);
-    lenient().when(baseRecordManager.get(docId01, u)).thenReturn(record);
-    lenient()
-        .when(groupshareRecordDao.getRecordGroupSharingsForRecordIds(List.of(docId01)))
-        .thenReturn(List.of(rgs));
+    when(groupshareRecordDao.get(anyLong())).thenReturn(rgs);
+    when(permissnUtils.findBy(any(), any(), any(), any())).thenReturn(cbp);
+    when(permissnUtils.createFromString(anyString()))
```
rs-nicof, line 90: This replaces two exact-argument stubs with a hand-rolled reimplementation of the collaborator, and the fake does not match the real one. `PermissionUtils.createFromString` (`src/main/java/com/researchspace/auth/PermissionUtils.java:183`) maps `read`/`view` to `READ` and `write`/`edit` to `WRITE`, and throws `IllegalArgumentException` for anything else.

`PermissionType.valueOf(arg.toUpperCase())` diverges in both directions: it would throw on the legal input `"edit"`, and it accepts values production refuses. A future caller change could then fail or pass for reasons unrelated to the code under test.

Two explicit stubs for `"READ"` and `"WRITE"` keep the pin, or delegate to a real `PermissionUtils` rather than restating its mapping.

src/test/java/com/researchspace/service/impl/MockFolderStructure.java
```diff
@@ -41,13 +38,8 @@ public UserFolderSetup create(User subject) {
   public UserFolderSetup create(
       User subject, FolderManager extfolderManagerMock, Folder extfolderMock) {
```
rs-nicof: `extfolderMock` is now dead. This hunk removed the only code that read it, but the signature still advertises it and `FolderTestUtils.createDefaultFolderStructure(User, FolderManager, Folder)` still passes one through. (`extfolderManagerMock` on the same line is still used at lines 41-42, so only the third parameter is affected.)

Nothing breaks today, because the one caller that passes a folder also installs its own `getFolder` stub. But this is a shared fixture, so the next test that passes a folder and expects it to be wired will get `null` back with nothing pointing at the cause.

Dropping the third parameter here and from the `FolderTestUtils` overload makes the compiler tell that caller to supply its own stub.

src/test/java/com/researchspace/service/inventory/impl/InventoryLinkManagerImplReferencingTest.java
```diff
@@ -77,19 +77,20 @@ void queriesByPrefixAndDbIdForInventoryTarget() {
   }
 
   private InventoryRecord parent(String globalIdString, String name, boolean deleted) {
```
rs-nicof: Same shape as `MockFolderStructure.create`: `parent(...)` no longer reads `globalIdString` or `name`, but callers still pass them and then repeat the same two literals into `stubSourceIdentity` (lines 109/111, 145/146, 162/163).

A reader at line 110 would reasonably conclude `parent("SA11", "secret sample", false)` created a record with that id and name; it did not. Reducing it to `parent(boolean deleted)` and letting `stubSourceIdentity` own identity removes the ambiguity.

src/test/java/com/researchspace/service/impl/IntegrationsHandlerTest.java
tchoi-rs: Moved the mock call to the right place

src/test/java/com/researchspace/webapp/controller/WorkspacePermissionsDTOBuilderTest.java
tchoi-rs: There's a `verify` call for this now just below

src/test/java/com/researchspace/webapp/controller/WorkspacePermissionsDTOBuilderTest.java
tchoi-rs: There's a `verify` call now

rs-nicof (approved):

PR #992, opened 2026-07-29
Validate uploaded file content against the claimed file type (PRT-1123)

rs-fraser:
## Description ##

This PR adds some hardening to image file uploads to address CWE-434 & CWE-20.

Previously, image uploads were stored and later served under a content type derived from the client-supplied filename, with no check that the bytes matched. Non-image content named `image.jpg` was accepted, stored, and returned as `image/jpeg`.

Now, uploads whose extension claims an image are checked against the actual content before being stored. Detection compares the detected type with the type the extension implies so `.jpg` extension must contain JPEG content, `.png` files must contain a PNG etc. 

The check sits in `MediaManagerImpl` at the two points every upload passes through, so it covers the public API, the gallery UI, and the importers.

Endpoints that serve stored file bytes now also send `X-Content-Type-Options: nosniff`, so the declared content type is honoured rather than guessed. It matters most where bytes are returned inline with no `Content-Disposition`, and it also helps for content uploaded before this change.

The header is applied to a named set of endpoints rather than globally: the file, gallery filestore and inventory file downloads; the workspace and generic download controllers; the network filestore and Box passthroughs; the WOPI file contents endpoint; and the endpoints returning image bytes inline, which are the image, gallery viewer, thumbnail, profile image and inventory image ones. 

Three endpoints that serve server-generated rather than stored content are deliberately excluded, namely the audit trail CSV, the calendar invitations, and the Snapgene PNG conversion.

## Design decisions

1. **Enforcement is limited to image extensions** (`jpg`, `jpeg`, `png`, `gif`, `bmp`, `tif`, `tiff`) as those all have reliable magic numbers for checking the content type. Text-based formats such as `txt`, `csv`, and the chemistry formats have no signature to verify, and audio and video container detection is inconsistent enough that enforcing it would risk rejecting valid files.

2. **Rejection throws `MediaContentMismatchException`**. It is a checked subclass of `IOException`, so existing file-handling callers can treat it as an I/O failure, while callers that need a precise validation message catch the subtype: the API advice maps it to a 422 with the resolved message, and the gallery controller returns its usual `AjaxReturnObject` error.

3. **The archive importer catches the rejection per gallery item and carries on.** The exception is checked rather than unchecked, so by Spring's default rule it does not mark the surrounding import transaction rollback-only, and nothing has been written when it is thrown. The importer reports the rejected file in the import report and continues with the rest, matching how it already handles image processing failures.

4. **The header is set per endpoint rather than by a filter over every response**. A global filter would be more durable and would cover endpoints added later, but it also applies to stylesheets and scripts, where a browser refuses a resource whose declared type is wrong instead of quietly correcting it. That is a plausible way to break the UI in a legacy webapp, and it wants checking against a running instance rather than shipping alongside a security fix. Worth doing as a follow-up.

5. **Detection reads only the leading bytes, so a valid image carrying appended trailing content still passes**. This is noted in the class javadoc. Closing it would mean re-encoding uploads from decoded pixels, which would alter the original bytes of the primary record, or per-format structural validation, which is awkward for TIFF in particular. Both are larger changes than this fix and better considered separately.


rs-fraser:
@greptileai

src/main/webapp/ui/src/modules/common/i18n/locales/en-US/server.core.json
```diff
@@ -22,6 +22,9 @@
     }
   },
   "errors": {
+    "upload": {
```
tchoi-rs, line 25: So that an attacker can't just brute-force guess their way into a vulnerable file type, I would not report the conclusion from the backend.

tchoi-rs (changes requested):

src/main/webapp/ui/src/modules/common/i18n/locales/en-US/server.core.json
rs-fraser: I've updated the message to remove the part about what was detected.

src/main/java/com/researchspace/service/impl/MediaManagerImpl.java
```diff
@@ -374,41 +405,35 @@ private EcatMediaFile doSaveMediaFile(
...
-          media =
-              mediaFactory.generateEcatChemistryFile(
-                  user, fp, extension, originalFileName, override);
-        } else {
-          // this will also include DMPs, which are always added to the top-level folder.
-          media =
-              mediaFactory.generateEcatDocument(
-                  user, fp, extension, mediaFolderType, originalFileName, override);
-        }
+                mediaFolderType, user, originalFileName, new FileInputStream(tempFile));
```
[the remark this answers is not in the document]
rs-fraser: Correct that the stream leaks, though it is pre-existing on main rather than introduced by this PR (the diff only re-indented these lines). Fixed in b4a19e0f8: both image paths now close the temp-file stream via try-with-resources once the file store has copied it. The same pattern existed in the update path, which is fixed too. Cleaning up the temp files themselves is left as separate housekeeping.

tchoi-rs (approved):

PR #980, opened 2026-07-28
RSDEV-1153: Fix PyRAT pagination lag

rs-fraser:
> [!NOTE]
> Stacked on #984 (the dialog extraction). Until that merges, this diff includes its changes too; review only the last three commits here (the fix, the regression tests, and the edge cases).

### Problem

In the PyRAT dialog, changing the page or the page size showed results one step behind the controls. The handlers fetched imperatively right after a `setState`, so each request was built from the previous render's `page` and `rowsPerPage` values. Changing the page size or sort while on a later page also fired a request for a page that no longer existed.

### Fix

- Drive the fetch off a memoised query string instead of fetching imperatively from the event handlers, so every pagination change produces exactly one request carrying the new values.
- Reset to the first page during render when the filters or sort change, so the reset and the new query land in a single request.

### Testing
AWS build has `user1a` configured with Pyrat to view the fix: https://rsdev-1153-pyrat-pagination-0b7cb50a-2.researchspace.com
(pangolin85 as `sysadmin1` is configured showing the previous bug behaviour): https://pangolin8085.researchspace.com

New browser-mode regression suite following the PubChem four-artifact pattern (`PyratDialog.spec.tsx`, `PyratDialog.story.tsx`, MSW mocks, page object):

- Changing the page size refetches with the new size and renders that many rows.
- Changing the page requests the matching offset.
- Page-size and sort changes made on a later page reset to the first page in a single request.
- Additional edge cases around the above.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #993, opened 2026-07-30
RSDEV-1256: Fix pre-existing issues in PyRAT, Omero, and Clustermarket dialogs

rs-fraser:
## Description ##

Fixes the eight issues listed in RSDEV-1256: seven pre-existing PyRAT dialog issues flagged during reviews of #980 / #984, plus the same render-time defect in the Omero and Clustermarket dialogs.

PyRAT dialog:

- The locations query sent a malformed `&=status=available` parameter (empty name), so buildings were never filtered by availability. Now sends `status=available`.
- Async filter callbacks updated state from captured values; both now use functional updates so an in-flight response cannot overwrite newer changes.
- Two string assignments to `element.style` replaced with `style.fontSize` / `style.fontWeight`.
- The `key` in `servers.map` moved from the nested `ListItem` to the returned fragment.
- Module-level `PYRAT_URL` / `PYRAT_ALIAS` / `VISIBLE_HEADER_CELLS`, mutated during render, replaced by component state passed explicitly to `createTinyMceTable`.
- Removed the unused `PyratListing` export.
- The inserted table's "Imported from ... on ..." line is now a single catalog key rendered through `TransRichText` rich markup, with the timestamp formatted per the active locale.

Omero and Clustermarket dialogs (same defect as the PyRAT globals): module-level header and selection variables were assigned during render and read back through exported getters by each plugin's `index.tsx`. They are replaced by component state and memos passed explicitly to the table builders, the TinyMCE insert listeners now register inside the components with cleanup, and each `index.tsx` reduces to the React bootstrap. One deliberate behaviour fix: Clustermarket's maintenance-notes column now reliably appears when "maintenance only" is ticked; the previous splice-based header mutation meant it never did.

## Testing notes

The mock-mode PyRAT e2e spec passes on chromium, firefox and webkit, and availability filtering was verified against a live PyRAT instance, since the mock ignores query parameters. The Omero and Clustermarket unit suites (49 tests) and the new PyRAT dialog browser tests pass. Intentional visible changes: locale-formatted timestamps in newly inserted PyRAT tables, and the Clustermarket maintenance-notes column behaviour above, which QA should check.


rs-fraser:
@greptileai

rs-fraser:
@greptileai please review the latest commit (Omero and Clustermarket dialog fixes).

src/main/webapp/ui/src/tinyMCE/clustermarket/Clustermarket.tsx
```diff
@@ -36,11 +38,102 @@ type HeaderCell = {
   numeric: boolean;
   label: string;
 };
-let VISIBLE_HEADER_CELLS: Array<HeaderCell> = [];
-let SELECTED_BOOKINGS: ReadonlyArray<BookingAndEquipmentDetails | EquipmentWithBookingDetails> = [];
-export const getSelectedBookings = (): ReadonlyArray<BookingAndEquipmentDetails | EquipmentWithBookingDetails> =>
-  SELECTED_BOOKINGS;
-export const getHeaders = (): typeof VISIBLE_HEADER_CELLS => VISIBLE_HEADER_CELLS;
+const makeBookingHeaderCells = (t: TFunction<"workspace">): Array<HeaderCell> => [
```
tchoi-rs, line 41: I would prefer to keep the functions unhoisted from the parent component becuase of the need of `t()` calls

src/main/webapp/ui/src/tinyMCE/clustermarket/Clustermarket.tsx
```diff
@@ -257,29 +236,53 @@ function Clustermarket({
...
+        mceAction: selectedBookings.length > 0 ? "enable" : "disable",
       },
       "*",
     );
-    return selected_bookings;
-  }, [selectedBookingIds]);
+  }, [selectedBookings]);
+
+  useEffect(() => {
+    const editor = window.parent.tinymce?.activeEditor;
```
tchoi-rs, line 265: I think the prevalent pattern is that the React dialogs are not aware of TinyMCE, and get callbacks from TinyMCE to use to manipulate editor content (i.e. creating the dialogs would belong in the TinyMCE plugin JS file).

tchoi-rs (changes requested):
No issues with tests this time, but caught some other issues.

src/main/webapp/ui/src/tinyMCE/clustermarket/Clustermarket.tsx
rs-fraser: fixed in in fc27ed727. Same change applied to the Omero dialog in 1c2021902.

src/main/webapp/ui/src/tinyMCE/clustermarket/Clustermarket.tsx
rs-fraser: fixed in dbc018dfc, reworked in f9efff0d6 so the table is built only when the plugin asks for it.

src/main/webapp/ui/src/tinyMCE/clustermarket/Clustermarket.tsx
```diff
@@ -249,38 +248,36 @@ function Clustermarket({
...
+        selectedBookings.length > 0
+          ? createTinyMceTable(selectedBookings, visibleHeaderCells, order, orderBy, clustermarket_web_url, t)
+          : null;
+      window.parent.postMessage(
+        { mceAction: "clustermarket-table", tableHtml: table?.outerHTML ?? null },
+        window.location.origin,
+      );
     };
-    editor.on("clustermarket-insert", insertSelectedBookings);
+    window.addEventListener("message", respondToInsertRequest);
```
tchoi-rs, line 278: This event name looks a bit too generic, but it sounds like TinyMCE set this event name?

tchoi-rs (approved):

src/main/webapp/ui/src/tinyMCE/clustermarket/Clustermarket.tsx
rs-fraser: Yeah it's not set by us, defined by the window message event.

PR #994, opened 2026-07-31
Rsdev 1250 gallery revision history

nhanlon2:
Supersedes #989, which was raised from a branch on the `ResearchSpace-ELN` fork. Same commits, rebased onto current `main`; this one is not cross-repo. Earlier review discussion is on #989.

## Description ##

Adds a version history to Gallery items, and a shareable URL that shows an item as it was at one past version.

- **Version history dialog.** The Actions menu gains "View Version History", listing an item's versions newest first with each version's own filename, modification date, editor and size. Backed by a new `GET /gallery/ajax/versionHistory/{mediaFileId}`.
- **Pinned version view.** `/gallery/item/{id}/{version}` shows the item at that version, read-only, with the item selected. Each row of the dialog links to it, so what a user lands on is also a URL they can copy into a document or a paper.
- **Read-only means read-only.** Rename, Delete, Move, Duplicate, Export, Upload-new-version and Edit are all refused on a pinned view; Download and the version history itself still work. The item's grid tile carries the version with the same clock icon a version-pinned Inventory link uses, and the InfoPanel says which version is showing and links back to the latest.

Frontend: 2338 unit tests pass. Backend: 2483 pure unit tests pass. `tsc`, `lint`, `i18n:lint` and `spotless` clean.

## Per item type

Whether "View Version History" is active is decided entirely by `GalleryFile.canViewVersionHistory` plus the Actions menu's single-selection rule. Nothing keys off the gallery section, so it reduces to: is it a folder, is it stored locally, and is it a media file.

| Item type | Backing class / `type` | Menu item |
| --- | --- | --- |
| Images | `EcatImage` / `"Image"` | **Active** |
| Audio | `EcatAudio` / `"Audio"` | **Active** |
| Video | `EcatVideo` / `"Video"` | **Active** |
| Documents (incl. Office/WOPI/Collabora files) | `EcatDocumentFile` / `"Document"` | **Active** |
| Chemistry | `EcatChemistryFile` / `"Chemistry"` | **Active** |
| DMPs | `EcatDocumentFile` / `"DMPs"` | **Active** |
| Miscellaneous | `EcatDocumentFile` / `"Miscellaneous"` | **Active** |
| Exports (PdfDocuments) | `EcatDocumentFile` / `"PdfDocuments"` | **Active** |
| Extensionless local file | any of the above | **Active** (deliberate: unlike `canUploadNewVersion`, no extension is required) |
| Pinned historical version being viewed | `HistoricalGalleryFile` wrapping a local file | **Active** (delegates to the wrapped file) |
| Snippets | `Snippet`, a `Record` rather than an `EcatMediaFile` / `"Snippet"` | Disabled: "Snippets do not have a version history." |
| Ordinary folder in any section | `"Folder"` | Disabled: "Folders do not have a version history." |
| System folder (Gallery section roots, `SNIPPETS`, shared folders) | `"System Folder"` | Disabled, same reason |
| Filestore root (NetworkFiles) | `Filestore` | Disabled: "Filestores do not have a version history." |
| File on a filestore | `RemoteFile` | Disabled: "Files stored in filestores do not have a version history." |
| Folder on a filestore | `RemoteFile` with `isFolder` | Disabled, same reason |
| Nothing selected, or 2+ selected | any | Disabled: "Only one item's version history may be viewed at once." |

Snippets are the one row worth a note. A snippet is a `Record`, so no path ever gives it a new version, and the endpoint audits `EcatMediaFile`, which a snippet is not. `canUploadNewVersion` already excluded them only incidentally, because a snippet carries no extension and that predicate requires one; `canViewVersionHistory` deliberately drops the extension check so extensionless versioned files still show a history, so it has to refuse snippets explicitly.

## Design decisions

**The audit row is the source of truth for everything about a version except the item's identity and its inbound references.** This is the one thing to keep in mind when reading the diff; three separate defects during development came from delegating a field to the live item that actually varies per version.

- `name`, `description`, `size`, `modificationDate`, `thumbnailUrl` and `downloadHref` all differ per version. Uploading a new version can replace the file with a differently named one, so the live item's name says nothing about what version 1 was called. `name` and `description` come from the audit row via the new DTO; both live on the audited `EditInfo` embeddable, so Envers records them per revision.
- `globalId` deliberately stays unversioned (`GL42`, never `GL42v2`). The ELN linked-documents and Inventory referencing lookups read it, and neither records which version a reference was made against. The InfoPanel renders `GL42v1` for display only, and a notice above both reference lists says they belong to the item rather than the version.

**A revision is not a version.** Several audit revisions can share one version, because not every recorded change bumps the counter. The dialog shows one row per version using the newest revision of each, via a helper shared with the Inventory version history rather than reimplemented.

**Refusals live on the object, not in the callers.** A past version is modelled as a decorator (`HistoricalGalleryFile`) that returns `Result.Error` from the `can*` predicates, so the Actions menu disables the right things without a single new conditional. `canBeEdited` had to be added to the interface: edit availability was previously decided solely by which editor applies to the file's type, so a past version stayed editable.

**No thumbnail or preview endpoint is version-aware,** so a pinned view shows a stock type icon or no preview rather than live content under an older version's label. Images and PDFs are fine, resolving through the version-aware `downloadHref`. Aspose could be made correct later, since it already accepts a `revisionId`.

**Version history is not part of what publishing a document shares.** `GalleryController` is mapped at `/public/publicView/gallery` as well as `/gallery`, Shiro treats `/public/**` as `anon`, and viewing a published document logs a real session in as the anonymous guest, which holds read permission on media linked from that document. So the ordinary permission check would have passed for it and disclosed every past filename and description plus the full name of every user who had edited the item. The guest is refused outright. See `DevDocs/adr/0003`.

Every new route needs a matching rule in `urlrewrite.xml`, which forwards it to the SPA shell before Spring MVC sees it. The rules are anchored regexes, so a new URL shape 404s until one is added.

Longer rationale in `DevDocs/adr/0003-gallery-version-history-endpoint.md` (why this has its own endpoint rather than extending the ELN document one) and `DevDocs/adr/0004-gallery-pinned-version-links.md` (why a route rather than a versioned Global ID), plus `src/main/webapp/ui/src/eln/gallery/README.md`.

## Testing notes

Manual check needs an item with more than one version: upload a file to the Gallery, then upload a **different** file over it via Actions → "Upload new version", ideally with a different filename so the per-version fields are visibly distinct.

1. Actions → "View Version History". Each row should show that version's own filename, not the current one.
2. Click `v1`. The URL becomes `/gallery/item/<id>/1`; the tile shows `v1` with a clock icon; the image and filename shown are version 1's, not the live one's.
3. On that view, Rename / Delete / Move / Duplicate / Export / Upload-new-version / Edit are all disabled with a reason, and the name and description fields are not editable. Download gives you version 1's bytes.
4. "View the latest version" returns you to the editable item, with no stale version badge or notice left behind.
5. Paste `/gallery/item/<id>/1` into a fresh tab. It should load the pinned view directly rather than 404.
6. A version the item does not have, e.g. `/gallery/item/<id>/99`, reports an error rather than falling back to the live item. A version equal to the live one redirects to `/gallery/item/<id>`.

Worth exercising with an Office document if Collabora or Office Online is enabled, since that is where edit access was previously reachable from a pinned view.

Things reviewers should know:

- `urlrewrite.xml` rules cannot be covered by any MVCIT, because MockMvc bypasses the servlet filter chain. Covering the new rule needs a Playwright spec.

Known follow-ups, deliberately not in scope here: moving the audit-to-DTO assembly into a `*Manager` so the Envers snapshots are read inside a live session (safe today only because every field read is a scalar or the `EditInfo` embeddable); the audit query is unbounded and hydrates full entities; and the `can*` predicate messages are English literals, which wants one sweep across the file rather than translating only the new lines.


tchoi-rs (approved):
Deferring the bug fixing to a later iteration

PR #995, opened 2026-07-31
RSDEV-1188: pin Inventory links to a Gallery item version

nhanlon2:
**Stacked on #994** (RSDEV-1250, gallery revision history). The base of this PR is `RSDEV-1250-gallery-revision-history`, not `main`, so the diff is only this change. Retarget it at `main` once #994 merges.

Unlocks version pinning for Gallery (`GL`) targets of Inventory Link fields.

Since RSDEV-1131 an Inventory item can link to Gallery, ELN and Inventory records, but a Gallery item had no revision history, so pinning a link to one of its versions was blocked. RSDEV-1250 gave Gallery items a revision history and a read-only pinned-version view, so the block can come off.

## What changed

Frontend only. The backend already resolved `GL` version pins: `LinkTargetSnapshotResolverImpl.resolveRevisionForVersion` has a `GL` branch calling `findRevisionNumberForMediaFileVersion`, and `InventoryLinkValidator` never gated pinning by target prefix.

- `iconForGlobalId.ts`: `supportsVersionPin("GL…")` is now `true`. That is the single gate the three call sites (`LinkField`, `UpdateField`, `LinkFieldValue`) read, so the pin affordance and the version dialog appear for Gallery targets.
- `iconForGlobalId.ts`: `openUrlForTarget` now honours the pin for `GL`, opening `/gallery/item/<id>/<version>` (the read-only view added in RSDEV-1250) by reusing `galleryItemHref` rather than repeating the route. Previously it dropped the version and always opened the live item.
- `VersionLockDialog.tsx`: `GL` added to the supported targets, plus one fetch branch on `GET /gallery/ajax/versionHistory/{id}`. That endpoint returns the inventory revisions shape inside an `AjaxReturnObject` envelope, so the branch is a fetch and a map.
- Same file: the inventory branch's inline revisions-to-versions collapsing was replaced by the shared `groupByVersion` helper, which both branches now use. Net fewer lines, and one grouping rule instead of two copies of it.
- `ElnRecordInfoDialog.tsx`: the info dialog forced the pin to `null` for every non-SD target, so its Open action dropped the version for a pinned Gallery link. It now keeps `effectiveVersionPin` (SD only) for the record-information fetch and the body, since that endpoint has no notion of a past Gallery revision, and uses a separate `openVersionPin` for the Open href.

Per ADR 0004 a pinned `GL` link deliberately opens the Gallery view rather than `/globalId/GL42v2`, which downloads bytes. The link card still displays `GL42v2`; that split is the accepted cost recorded in the ADR.

No new i18n keys, no schema change, no new dependency.

## Docs

ADRs 0003 and 0004 both recorded `GL` link pinning as deliberately left unimplemented. Those consequence bullets now record it as done here and note which route a pinned link opens.

## Testing

- `iconForGlobalId.test.ts`: the `supportsVersionPin` expectations flip for `GL`, plus a case for the pinned Gallery URL.
- `VersionLockDialog.test.tsx`: two `GL` cases, one that the gallery endpoint is called and several revisions of one version collapse to a single row, one that the confirmed pin is the version number and not the audit revision id.
- `ElnRecordInfoDialog.test.tsx`: a pinned `GL` target's Open href carries the version segment while the body is still fetched unpinned.
- Two existing tests encoded the old "GL cannot be pinned" rule and were flipped: `LinkField`'s clock-visibility case, and `UpdateField`'s greyed-clock case, which now uses `NB5` (the remaining unversionable kind) with a new GL-enabled case beside it.
- 213 unit tests pass across the link suites and the gallery version-history dialog suite; `pnpm tsc` and `pnpm lint` clean.

## Review history

An earlier copy of this branch was reviewed by Copilot on a fork PR. One finding was real and is fixed here (the `ElnRecordInfoDialog` Open href above). Two were declined: both claimed `/gallery/ajax/versionHistory/{id}` does not return `revisionsCount`, which the `GalleryVersionHistory` DTO (`record GalleryVersionHistory(List<Revision> revisions, int revisionsCount)`) shows it does. Copilot's follow-up pass agreed and reported no remaining issues.

### DIFF
--- a/src/main/webapp/ui/src/Inventory/components/Fields/Link/VersionLockDialog.tsx
+++ b/src/main/webapp/ui/src/Inventory/components/Fields/Link/VersionLockDialog.tsx
@@ -12,6 +12,7 @@ import VersionLockPicker, {
   type VersionLockSelection,
   type VersionRecord,
 } from "../../../../components/VersionLockPicker/VersionLockPicker";
+import { groupByVersion } from "../../../../util/versionHistory";
 
 export interface VersionLockDialogProps {
   open: boolean;
@@ -48,9 +49,12 @@ function parseGlobalId(globalId: string): ParsedTarget | null {
   };
 }
 
-/** Targets whose revision history this dialog can resolve: inventory items and SD documents. */
+/**
+ * Targets whose revision history this dialog can resolve: inventory items, SD documents and GL
+ * gallery files.
+ */
 function isSupportedTarget(parsed: ParsedTarget): boolean {
-  return parsed.inventoryPathSegment !== null || parsed.prefix === "SD";
+  return parsed.inventoryPathSegment !== null || parsed.prefix === "SD" || parsed.prefix === "GL";
 }
 
 interface ApiRevisionEntry {
@@ -67,6 +71,24 @@ interface ApiRevisionEntry {
 interface ApiRevisionList {
   revisions: ApiRevisionEntry[];
   revisionsCount: number;
+}
+
+/** The gallery endpoint's AjaxReturnObject envelope around the same revision shape. */
+interface AjaxRevisionList {
+  data: ApiRevisionList | null;
+}
+
+/**
+ * Collapses audit rows to one picker row per user-facing version (several revisions can share one
+ * version), pinning the version and carrying its newest revision. Shared by the inventory and
+ * gallery branches, whose responses have the same shape.
+ */
+function toVersionRecords(revisions: ApiRevisionEntry[]): VersionRecord[] {
+  return groupByVersion(revisions).map(({ version, revision }) => ({
+    version,
+    revisionId: revision.revisionId,
+    modificationDate: revision.record.lastModified ?? "",
+  }));
 }
 
 // ELN revisions endpoint shape (/workspace/revisionHistory/ajax/{id}/versions). Each entry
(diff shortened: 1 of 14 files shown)


src/main/webapp/ui/src/Inventory/components/Fields/Link/__tests__/VersionLockDialog.test.tsx
```diff
@@ -328,6 +360,37 @@ describe("VersionLockDialog (SD/ELN document target)", () => {
     expect(vi.mocked(axiosGet)).not.toHaveBeenCalled();
   });
 
+  it("fetches GL revisions from the gallery endpoint, collapsing revisions of one version", async () => {
+    // eslint-disable-next-line @typescript-eslint/unbound-method -- mock setup
+    const axiosGet = axios.get;
```
tchoi-rs: Use MSW instead of mocking axios

src/main/webapp/ui/src/Inventory/components/Fields/Link/VersionLockDialog.tsx
```diff
@@ -126,22 +148,15 @@ export default function VersionLockDialog(props: VersionLockDialogProps): React.
         }
         return [...byVersion.values()].sort((a, b) => b.version - a.version);
       }
+      if (parsed.prefix === "GL") {
+        // Gallery revisions come from the ELN gallery endpoint (RSDEV-1250), which wraps the same
+        // revision shape as the inventory API in an AjaxReturnObject.
+        const { data } = await axios.get<AjaxRevisionList>(`/gallery/ajax/versionHistory/${parsed.id}`);
```
tchoi-rs: Where is the error handled?

tchoi-rs (changes requested):

src/main/webapp/ui/src/Inventory/components/Fields/Link/__tests__/VersionLockDialog.test.tsx
nhanlon2: fixed in d630502c5

src/main/webapp/ui/src/Inventory/components/Fields/Link/VersionLockDialog.tsx
nhanlon2: fixed in d630502c5

src/main/webapp/ui/src/Inventory/components/Fields/Link/VersionLockDialog.tsx
```diff
@@ -103,52 +130,50 @@ export default function VersionLockDialog(props: VersionLockDialogProps): React.
...
-            version: entry.version,
-            revisionId: entry.revision,
-            modificationDate: entry.modificationDate ?? "",
-          });
-        }
-        return [...byVersion.values()].sort((a, b) => b.version - a.version);
+    if (parsed.prefix === "SD") {
+      // ELN document revisions come from the workspace endpoint, not the inventory
+      // API. Pin to the document version number (entry.version); carry the audit id.
+      const { data } = await axios.get<ElnRevisionHistoryResponse>(
```
tchoi-rs: Could you also add handling on this path as well while we're here

tchoi-rs (changes requested):

src/main/webapp/ui/src/Inventory/components/Fields/Link/VersionLockDialog.tsx
nhanlon2: fixed 5a452aec0

tchoi-rs (approved):

PR #997, opened 2026-07-31
RSDEV-1246: template link fields with a default link (+ RSDEV-1270 link-row fix, link UI rework)

nhanlon2:
## What

Template Link fields can now carry a **default link**. Every item created from that template gets it as the item's own ordinary, editable link. Follow-on from RSDEV-1131, where a template Link field could declare only a name and a set of allowed relationship types.

Two things ride along:

- **RSDEV-1270 fix.** The instrument manager now soft deletes the `InventoryLink` row when a link field is deleted. It did this in neither of the two places the sample manager does.
- **Link editor UI.** The shared editor is regrouped into Target / Relationship type / Version. Cosmetic, no functional change, and it applies to all three callers: item link fields, extra-field links, and the new template default.

Glossary terms: `DevDocs/CONTEXT.md`. Storage decision: `DevDocs/adr/0006-template-default-link-is-whole-or-nothing.md`.

## Why the backend is small

No schema change, no new API field, no new copy code. `InventoryEntityField.link_id` and `ApiInventoryEntityField.link` already exist on the single table and DTO shared by template fields and item fields (`changeLog-rsdev-1131.xml` changeSet 5). `link_id` was simply never populated on template rows, deliberately, in three places. Items are stamped by the existing `InventoryLinkField.shallowCopy()` chain, so bulk, API and UI creation behave identically down one code path.

Most of the diff on the two managers is therefore removal. Every link helper both of them had a copy of moved up to `InventoryApiManagerImpl`: `applyLinkFieldValue`, `applyLinkFieldValuesOnUpdate`, `applyLinkFieldValueOnCreate`, `softDeleteLinkOfDeletedLinkField`, `isRelationPermitted`, `rejectSelfLink`, `assertRelationAllowed`, `parseTargetOrNull`, joining `syncLinkFieldWhitelistsFromTemplate` (RSDEV-1200). Each manager keeps a one-line typed delegate. That step alone is **net −116 lines** across the three files; the whole feature leaves them at net +62.

## Decisions

1. **Whole or nothing.** A default is a relation type *and* a target, stored as an ordinary `InventoryLink` on the template field's existing `link_id`. Half a default is not supported: `InventoryLink`'s four identifying columns are all `NOT NULL`, and relaxing them pushes a null branch into every consumer forever. ADR-0006 records the two rejected alternatives.
2. **Stamped verbatim.** A default whose target has since been deleted, or which the creating user cannot read, is copied onto the item unchanged. The alternative makes one template produce different items for different users.
3. **Templates appear in "referenced by".** A template holding a default shows up in its target's referencing-items list. Wanted: it shows that new items will keep being created pointing at that record.

Two consequences fall out rather than being decided: defaults are **never retro-applied**, and **version pins are allowed and stay pinned**, because `InventoryLink.shallowCopy()` already carries `version_pin` and `target_revision_id`.

## Validation

Reusing the item write path brings the DataCite vocabulary check, GlobalID parse, allowed-target-kind, existence-and-readability, whitelist and self-link checks across to templates unchanged. Two things are new:

- **Whitelist conflict**, 422 `errors.inventory.field.link.defaultRelationTypeNotPermitted`. Narrowing a field's allowed relationship types past its own default is rejected instead of silently dropping the default. Needed because an unchanged default is a no-op on the per-link path, so a whitelist-only edit would otherwise slip through.
- **`linkProvided` on `ApiInventoryEntityField`.** An absent `link` key is now distinguishable from an explicit `"link": null`. Without it, absent meant clear, so a template PUT naming only `allowedRelationTypes`, which is exactly the request the error above asks the user to make, destroyed that field's default link and then passed the whitelist guard because the guard saw a null link. Item payloads keep RSDEV-1131 semantics: their field list arrives complete, so absent still clears. The flag is `@JsonIgnore` and excluded from equals/toString, and the outgoing constructor assigns the field rather than the setter so serialization cannot forge it.

## RSDEV-1270

Deleting an inventory field is a **soft** delete: an UPDATE, not a JPA remove, so the `cascade`/`orphanRemoval` on `InventoryLinkField.link` never fires and the link row must be soft deleted explicitly at the service layer.

- *Already broken on `main`.* Concrete instrument link fields do hold links, so a template field deletion propagating through `updateInstrumentToLatestTemplateVersion` orphaned that instrument's link row.
- *Would have been widened here.* Once instrument templates carry defaults, deleting a template field orphans the template's own row too.

Harm was latent: both referencing-items queries filter `f.deleted = false`, and user deletion hard deletes orphans by join. What broke was the invariant "a live `InventoryLink` belongs to a live field", plus an `InventoryLink_AUD` trail that never recorded the delete. Ordering matters: the snapshot is taken **before** the template sync, because `getActiveFields()` no longer returns the newly deleted fields afterwards.

## Frontend

The template editor reuses the item editor (`LinkFieldValue`) wholesale, so relation constraint, self-link rejection, target-existence checking, version pinning and the Apply-or-discard save guard all carry over for free. `LinkFieldValue` gains three optional props: `showFieldName` (false in the template, where the name is already entered in the Name field directly above), `targetHeading`, and `nestHeadings`.

Shared `LinkEditor` layout, applying to all three callers:

```
Target
  [Target Global ID]  [Browse Inventory]  [Browse ELN]
  (SA42 chip, with delete)
Relationship type ____
Version
  (Latest)  (clock)
```

- The Global ID field moved up onto the Browse buttons' row, sized to content (`11em`), with a placeholder instead of a floating label so it aligns with the buttons rather than sitting below them.
- The selected-target chip moved beneath that row. In the row it competed for width, and `MuiChip-label` is `overflow: hidden`, so a shrunk chip lost its own delete X. There is a regression test for this.
- Target and Version gained headings, rendered through `DynamicHeadingLevel` so they take the correct level at both of the depths this component mounts at.
- The template's default-link block is boxed and headed "Default Link (optional)", with its target group headed "Default link target" so it is not confused with an item's own link.

Each control keeps its own `aria-label`, so screen-reader naming is unchanged.

## Tests

- **27 new backend unit cases** across `SampleApiManagerImplLinkFieldTest`, `InstrumentEntityApiManagerImplLinkFieldTest`, `ApiInventoryEntityFieldLinkContentTest` and `JsonMessageSourceTest`.
- **22 new frontend cases**: a new `LinkEditor.test.tsx` (11), plus `DefaultValueField.test.tsx` (8) and `LinkFieldValue.test.tsx` (3).

Green: `mvn test -Dfast=true` 2699, `pnpm test` 420 files / 2458, plus `pnpm tsc`, `pnpm lint`, `pnpm run i18n:lint`, `mvn spotless:check`.

**Not run:** `SampleTemplateDefaultLinkMVCIT` (6) and `InstrumentTemplateDefaultLinkMVCIT` (4) are written and compile but have never been executed, following the repo convention for real-transaction MVC bases. They need a reviewer with a DB. The two RSDEV-1270 cases in the instrument one should **fail on `main`** at the `deleted=true` assertion, worth confirming, since the orphaned row is invisible through the API and the assertions therefore reach it via the DAO.

## Out of scope

Archive export/import does not carry `InventoryLink` at all, so a default link does not survive export. Pre-existing gap, unchanged here.

Known follow-up for a separate ticket: in `rspace-core-model`, `Instrument#updateToLatestTemplateVersion` calls `setFieldData(null)` where `Sample` calls `clearValue()`, and `InventoryLinkField` overrides only the latter. That asymmetry is why the instrument path needs `clearRetroStampedDefaultLinks` here to compensate. Fixing it upstream would delete those ~20 lines, but it needs a JitPack build and a version bump, so it does not belong in this PR.

### DIFF
--- a/src/main/java/com/researchspace/api/v1/model/ApiInventoryEntityField.java
+++ b/src/main/java/com/researchspace/api/v1/model/ApiInventoryEntityField.java
@@ -67,6 +67,24 @@ public class ApiInventoryEntityField extends ApiField {
   @JsonInclude(JsonInclude.Include.NON_NULL)
   private ApiInventoryLink link;
 
+  /**
+   * Whether the incoming payload mentioned {@code link} at all, so that an explicit {@code "link":
+   * null} ("clear it") can be told apart from a partial update that simply does not mention it. Set
+   * by the hand-written {@link #setLink} below, which replaces Lombok's generated one.
+   *
+   * <p>Without the distinction, absent meant clear: a template PUT naming only the field's
+   * allowed-relation-types (exactly the request the whitelist-conflict error asks the user to make)
+   * silently destroyed that field's default link, and left the whitelist guard seeing a null link
+   * so it passed (RSDEV-1246). Excluded from JSON, equals and toString: it describes the payload,
+   * not the field.
+   */
+  @JsonIgnore @EqualsAndHashCode.Exclude @ToString.Exclude private boolean linkProvided;
+
+  public void setLink(ApiInventoryLink link) {
+    this.link = link;
+    this.linkProvided = true;
+  }
+
   @JsonProperty("mandatory")
   private Boolean mandatory;
 
--- a/src/main/java/com/researchspace/api/v1/model/ApiInventoryEntityField.java
+++ b/src/main/java/com/researchspace/api/v1/model/ApiInventoryEntityField.java
@@ -68,15 +68,11 @@ public class ApiInventoryEntityField extends ApiField {
   private ApiInventoryLink link;
 
   /**
-   * Whether the incoming payload mentioned {@code link} at all, so that an explicit {@code "link":
-   * null} ("clear it") can be told apart from a partial update that simply does not mention it. Set
-   * by the hand-written {@link #setLink} below, which replaces Lombok's generated one.
-   *
-   * <p>Without the distinction, absent meant clear: a template PUT naming only the field's
-   * allowed-relation-types (exactly the request the whitelist-conflict error asks the user to make)
-   * silently destroyed that field's default link, and left the whitelist guard seeing a null link
-   * so it passed (RSDEV-1246). Excluded from JSON, equals and toString: it describes the payload,
-   * not the field.
+   * Whether the incoming payload mentioned {@code link} at all, telling an explicit {@code "link":
+   * null} apart from a partial update that never mentions it. Without the distinction, a template
+   * PUT naming only allowed-relation-types destroyed that field's default link (RSDEV-1246). Set by
+   * the hand-written {@link #setLink} below. Excluded from JSON, equals and toString: it describes
+   * the payload, not the field.
    */
   @JsonIgnore @EqualsAndHashCode.Exclude @ToString.Exclude private boolean linkProvided;
 
@@ -161,9 +157,8 @@ public class ApiInventoryEntityField extends ApiField {
       InventoryLinkField linkField = (InventoryLinkField) field;
       setAllowedRelationTypes(splitRelationTypes(linkField.getAllowedRelationTypes()));
       if (linkField.getLink() != null) {
-        // assign the field, not the setter: linkProvided means "the client sent a link key", and
-        // this is the outgoing direction. Going through setLink would make it true on every
-        // serialized field and rob the flag of its meaning.
+        // the field, not the setter: this is the outgoing direction, and setLink would set
+        // linkProvided on every serialized field, robbing the flag of its meaning
         this.link = new ApiInventoryLink(linkField.getLink());
       }
     }
--- a/src/main/java/com/researchspace/api/v1/model/ApiInventoryEntityField.java
+++ b/src/main/java/com/researchspace/api/v1/model/ApiInventoryEntityField.java
@@ -161,7 +161,10 @@ public class ApiInventoryEntityField extends ApiField {
       InventoryLinkField linkField = (InventoryLinkField) field;
       setAllowedRelationTypes(splitRelationTypes(linkField.getAllowedRelationTypes()));
       if (linkField.getLink() != null) {
-        setLink(new ApiInventoryLink(linkField.getLink()));
+        // assign the field, not the setter: linkProvided means "the client sent a link key", and
+        // this is the outgoing direction. Going through setLink would make it true on every
+        // serialized field and rob the flag of its meaning.
+        this.link = new ApiInventoryLink(linkField.getLink());
       }
     }
 
--- a/src/main/java/com/researchspace/service/inventory/impl/InstrumentEntityApiManagerImpl.java
+++ b/src/main/java/com/researchspace/service/inventory/impl/InstrumentEntityApiManagerImpl.java
@@ -264,8 +264,25 @@ public class InstrumentEntityApiManagerImpl extends InventoryApiManagerImpl<Inst
    * so the target is parsed/validated and the Envers revision captured. Mirrors the implementation
    * in {@link SampleApiManagerImpl}.
    */
+  /** Item semantics: see the four-argument overload. */
   boolean applyLinkFieldValue(
       InventoryLinkField field, ApiInventoryEntityField apiField, User user) {
+    return applyLinkFieldValue(field, apiField, user, false);
+  }
+
+  /**
+   * @param omittedLinkPreservesExisting how to read a payload carrying no {@code link} key at all.
+   *     True for a <b>template</b> field, whose PUT accepts a partial field list, so a
+   *     whitelist-only edit must not destroy the default link. False for an <b>item</b> field,
+   *     whose field list always arrives complete, so an absent link means the user cleared it
+   *     (RSDEV-1131). An explicit {@code "link": null} clears in both cases. Mirrors {@link
+   *     SampleApiManagerImpl}.
+   */
+  boolean applyLinkFieldValue(
+      InventoryLinkField field,
+      ApiInventoryEntityField apiField,
+      User user,
+      boolean omittedLinkPreservesExisting) {
     ApiInventoryLink apiLink = apiField.getLink();
     String target = apiLink == null ? null : apiLink.getTargetGlobalId();
     InventoryLink existing = field.getLink();
@@ -273,8 +290,8 @@ public class InstrumentEntityApiManagerImpl extends InventoryApiManagerImpl<Inst
       if (existing == null) {
         return false;
       }
-      if (!apiField.isLinkProvided()) {
-        // a partial update that never mentions the link: leave it alone rather than destroy it
+      if (omittedLinkPreservesExisting && !apiField.isLinkProvided()) {
+        // a partial template update that never mentions the link: leave it alone
         return false;
       }
       field.setLink(null);
@@ -330,7 +347,9 @@ public class InstrumentEntityApiManagerImpl extends InventoryApiManagerImpl<Inst
               .findFirst();
       if (dbFieldOpt.isPresent()) {
         rejectSelfLink(apiField.getLink(), dbInstrument);
-        changed |= applyLinkFieldValue((InventoryLinkField) dbFieldOpt.get(), apiField, user);
+        changed |=
+            applyLinkFieldValue(
+                (InventoryLinkField) dbFieldOpt.get(), apiField, user, dbInstrument.isTemplate());
       }
     }
     return changed;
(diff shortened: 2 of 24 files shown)


nhanlon2:
@greptileai

src/main/java/com/researchspace/service/inventory/impl/SampleApiManagerImpl.java
```diff
@@ -1073,15 +1052,32 @@ public ApiSampleTemplate createSampleTemplate(ApiSampleTemplatePost apiSample, U
...
   }
 
-  private void createFields(ApiSampleTemplatePost apiSample, SampleTemplate sample) {
+  void createFields(ApiSampleTemplatePost apiSample, SampleTemplate sample, User user) {
     InventoryFieldNameUniquenessValidator.assertNoDuplicateFieldNamesInRequest(
         apiSample.getFields(), null);
     for (ApiInventoryEntityField field : apiSample.getFields()) {
       InventoryEntityField toAdd = apiFieldToModelFieldFactory.apiInventoryFieldToModelField(field);
+      applyDefaultLinkOfNewTemplateField(toAdd, field, user);
       sample.addSampleField(toAdd);
```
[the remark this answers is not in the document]
nhanlon2: Checked this. The create path cannot produce a self-targeting default, but the check that stops it is not `rejectSelfLink`, it is the target-existence check one layer down.

`createFields` runs *before* `sampleTemplateDao.persistSampleTemplate`, so at that moment the template has no id **and no row**. `applyLinkFieldValue` therefore sees `existing == null` and calls `inventoryLinkManager.createLink`, which calls `assertTargetExistsAndReadable` before writing anything. The template's own future Global ID does not resolve to an existing readable record, so a crafted POST is rejected with 422 `targetNotFound`. Any id that *does* resolve belongs to some other record, which is not a self link.

Every path where a self link is genuinely reachable passes an already-saved record (`applyLinkFieldValuesOnUpdate`, and `createDeleteRequestedFieldsInDbSampleTemplate` adding a field to a saved template), and there `getId()` is non-null so `rejectSelfLink` does run.

So no post-persist check is being added for an unreachable case. I have extended the comment on the `getId() == null` guard to record that reasoning, so it no longer reads as an accidental bypass. Leaving this thread open for the human reviewer to disagree if they want the belt-and-braces check anyway.

src/main/java/com/researchspace/service/inventory/impl/InstrumentEntityApiManagerImpl.java
```diff
@@ -672,15 +670,30 @@ public ApiInstrumentTemplate createInstrumentTemplate(ApiInstrumentTemplatePost
...
-  private void addFieldsToNewInstrumentTemplate(
-      ApiInstrumentTemplatePost post, InstrumentTemplate template) {
+  void addFieldsToNewInstrumentTemplate(
+      ApiInstrumentTemplatePost post, InstrumentTemplate template, User user) {
     for (ApiInventoryEntityField apiField : post.getFields()) {
       InventoryEntityField toAdd =
           apiFieldToModelFieldFactory.apiInventoryFieldToModelField(apiField);
+      applyDefaultLinkOfNewTemplateField(toAdd, apiField, user);
       addNewFieldToInstrumentTemplate(template, toAdd);
     }
```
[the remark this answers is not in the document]
nhanlon2: Same reasoning as the sample-side thread. `addFieldsToNewInstrumentTemplate` runs before `instrumentTemplateDao.persistInstrumentTemplate`, so the template has neither an id nor a row; `applyLinkFieldValue` takes the `createLink` branch, and `createLink` calls `assertTargetExistsAndReadable` before any link row is written. The template's own future Global ID cannot resolve, so the request is rejected with 422 `targetNotFound` rather than creating a self-referencing default.

Comment on the `getId() == null` guard extended to record this. Left open for the human reviewer.

src/test/java/com/researchspace/api/v1/controller/SampleTemplateDefaultLinkMVCIT.java
```diff
@@ -0,0 +1,271 @@
...
+    ApiSampleTemplatePost templatePost = new ApiSampleTemplatePost();
+    templatePost.setName("template with a default link");
+    templatePost.setDefaultUnitId(RSUnitDef.GRAM.getId());
+    templatePost.setSampleSource(SampleSource.LAB_CREATED);
+    ApiInventoryEntityField linkField = new ApiInventoryEntityField();
+    linkField.setName("Related items");
+    linkField.setType(ApiFieldType.LINK);
+    linkField.setAllowedRelationTypes(List.of("References", "IsDerivedFrom"));
+    templatePost.setFields(List.of(linkField));
+
```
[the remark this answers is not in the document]
nhanlon2: Already done. The unused `ApiSampleTemplatePost` (and its import) were removed in be8fb1140 during an internal review pass, before this review landed. `SampleTemplateDefaultLinkMVCIT` no longer references the type at all; the request is built from the `templateJson` string as you suggest.

nhanlon2:
@greptileai

src/main/java/com/researchspace/api/v1/model/ApiInventoryEntityField.java
```diff
@@ -67,6 +67,24 @@ public class ApiInventoryEntityField extends ApiField {
...
+   * null} ("clear it") can be told apart from a partial update that simply does not mention it. Set
+   * by the hand-written {@link #setLink} below, which replaces Lombok's generated one.
+   *
+   * <p>Without the distinction, absent meant clear: a template PUT naming only the field's
+   * allowed-relation-types (exactly the request the whitelist-conflict error asks the user to make)
+   * silently destroyed that field's default link, and left the whitelist guard seeing a null link
+   * so it passed (RSDEV-1246). Excluded from JSON, equals and toString: it describes the payload,
+   * not the field.
+   */
+  @JsonIgnore @EqualsAndHashCode.Exclude @ToString.Exclude private boolean linkProvided;
```
mKowalski256, line 77: I debated this change a bit with my AI agent, as the current Inventory API convention is to use explicit properties like:
```
@JsonProperty(value = "newFieldRequest", access = Access.WRITE_ONLY)`
private boolean deleteFieldRequest;
```
so in pre-existing code/API the cleanup request is explicit, rather than derived from property absence/null value.

But I'm now convinced the approach from this PR is better and should be used for new functionalities.

mKowalski256 (approved):
I've reviewed the backend code, looks good to me. 

Also did some testing on [AWS build](https://rsdev-1246-template-links-with-targets-set-68f31761-4.researchspace.com/). The functionality works as expected, but I'm noting a small visual issue/inconsistency which is maybe worth fixing?

The issue: the new panel for editing the saved link has a 'cog' icon which 1) is not verticall-centered 2) I think is a different 'cog' icon than used in other places, e.g. in left-hand listing. 
 
See the screenshot:
Image

src/main/java/com/researchspace/api/v1/model/ApiInventoryEntityField.java
nhanlon2: The cog icon is used on all custom fields in inventory and is used to edit them. Tempates now define default links which need to be editable. I would need to change the link gui to be special in some way for deafult links in templates, in order to change the cog.

mKowalski256: Oh I see how the cog icon was reused now, seems like we have two different in the UI. Can stay as is then, cheers.

src/main/webapp/ui/src/Inventory/components/Fields/Link/LinkEditor.tsx
```diff
@@ -156,20 +156,64 @@ export default function LinkEditor({
...
+            // the accented theme strips from deletable chips
+            // (`&.MuiChip-deletable { padding: 0 }`). Without it the type
+            // icon — which gets no MuiChip-icon margin because
+            // RecordTypeIcon wraps it in a tooltip — sits flush against
+            // the left edge instead of the non-edit pill's 4px
+            // (spacing(0.5)). The selector is repeated to out-specify the
+            // theme's two-class `.MuiChip-deletable` rule. The cancel
+            // button just widens the chip.
+            size="small"
+            sx={{ "&.MuiChip-deletable": { pl: 0.5 } }}
```
tchoi-rs: Use MUI class name exports

tchoi-rs (changes requested):
No big issues as I can spot. Could you ask Claude to cleanup the comments please?

src/main/webapp/ui/src/Inventory/components/Fields/Link/LinkEditor.tsx
nhanlon2: Fixed this and one other instance in c5c3bb678 . Comments also shortened.

tchoi-rs (approved):

PR #999, opened 2026-07-31
Add RTL skill

tchoi-rs:
## Description ##
Add skills for writing React Test Library tests.

### DIFF
--- a/.agents/skills/react-testing-library/README.md
+++ b/.agents/skills/react-testing-library/README.md
@@ -1,0 +1,139 @@
+# React Testing Library Best Practices
+
+A comprehensive skill for writing maintainable, user-centric tests with React Testing Library.
+
+## Overview
+
+This skill contains 43 rules across 9 categories, prioritized by impact to guide test writing and code review. It covers query selection, async handling, user interactions, assertions, and common anti-patterns.
+
+### Structure
+
+```
+react-testing-library/
+├── SKILL.md              # Entry point with quick reference
+├── metadata.json         # Version, org, references
+├── README.md             # This file
+├── references/
+│   ├── _sections.md      # Category definitions
+│   ├── query-*.md        # Query selection rules (CRITICAL)
+│   ├── async-*.md        # Async handling rules (CRITICAL)
+│   ├── anti-*.md         # Anti-pattern rules (CRITICAL)
+│   ├── user-*.md         # User interaction rules (HIGH)
+│   ├── assert-*.md       # Assertion rules (HIGH)
+│   ├── setup-*.md        # Component setup rules (MEDIUM)
+│   ├── struct-*.md       # Test structure rules (MEDIUM)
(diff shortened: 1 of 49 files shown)


skills-lock.json
```diff
@@ -0,0 +1,11 @@
+{
```
rs-fraser, line 1: Should we mention this file in `AGENTS.md`?

.agents/skills/react-testing-library/README.md
```diff
@@ -0,0 +1,139 @@
...
+## Overview
+
+This skill contains 43 rules across 9 categories, prioritized by impact to guide test writing and code review. It covers query selection, async handling, user interactions, assertions, and common anti-patterns.
+
+### Structure
+
+```
+react-testing-library/
+├── SKILL.md              # Entry point with quick reference
+├── metadata.json         # Version, org, references
```
rs-fraser, line 14: We don't have this file

.agents/skills/react-testing-library/README.md
```diff
@@ -0,0 +1,139 @@
...
+│   ├── struct-*.md       # Test structure rules (MEDIUM)
+│   ├── debug-*.md        # Debugging rules (LOW-MEDIUM)
+│   └── a11y-*.md         # Accessibility rules (LOW)
+└── assets/
+    └── templates/
+        └── _template.md  # Template for new rules
+```
+
+## Getting Started
+
```
rs-fraser, line 33: Should this section be removed? Could be confused with our pnpm commands.

rs-fraser (changes requested):
Looks good. Left a couple of inline comments, and Claude tells there are some conflicts with existing instruction files. We should align those files with the new skill:

> The skill's trigger description is broad ("Triggers on test files, testing patterns..."), so agents will load it for every *.test.tsx task, yet it contradicts the repo's checked-in rules:
>- setup-custom-render.md and setup-wrapper-providers.md both tell agents to create a test-utils module that re-exports a wrapped render. The repo CLAUDE.md explicitly says to import render and within directly from @testing-library/react.
>- setup-mock-modules.md (and the vi.mock('./api') examples elsewhere) push module-level mocking of API code. The repo standard is MSW via the shared server export with strict unhandled-request errors. MSW is never mentioned anywhere in the skill.
>- The a11y section doesn't mention the repo's toBeAccessible / expectAccessible helpers, so it will steer agents toward weaker generic patterns.
> 
> Suggestion: add a short rspace-specific preamble to SKILL.md (repo conventions win; use MSW not vi.mock for network; import render directly, no custom test-utils; use toBeAccessible; *.spec.tsx work goes to the rspace-browser-tests skill instead).

skills-lock.json
tchoi-rs: This is only used for `pnpx skills install` so shouldn't be necessary

rs-fraser (approved):

PR #1000, opened 2026-08-03
Workspace e2e tests

nebay-abraha:
## Description ##
Adds a dedicated E2E test for workspace functionality.

## Testing notes
Ran e2e locally and [ci](https://github.com/rspace-os/rspace-web/actions/runs/30796230267) .


tchoi-rs (approved):
All looks good, I would suggest doing another scan for some of the regex locators to see whether they truly need to be in regex form

PR #1010, opened 2026-08-05
RSDEV-900 Exclude temporary autosave document copies from workspace search

rs-fraser:
## Description

Workspace search could return version 0 documents when a temporary document was created due to an autosave of doc ticking or session timeout.

The fix lives in rspace-core-model (pinned here, see [rspace-os/rspace-core-model#84](https://github.com/rspace-os/rspace-core-model/pull/84)): a routing binder on `StructuredDocument` so temporary copies never enter the Lucene index, which also stops the pointless reindex of the copy on every autosave tick.

## Design decisions

- The exclusion lives on the entity `@Indexed` annotation, where the indexing is defined.
- An earlier revision also added a defensive post-search filter for temp copies already in the index. Review pointed out it is redundant: this ships with the release whose Lucene 5 to 9 upgrade makes every pre-upgrade index unreadable, so any working index is rebuilt with the binder active and stale temp entries cannot survive. The filter was removed.

Related `rspace-core-model` PR: https://github.com/rspace-os/rspace-core-model/pull/84


mKowalski256 (approved):
Looks good & solves the issue for me on test AWS instance, but I wonder if we actually need the defensive code not displaying previously indexed docs, as with 1.125/2.25 release we are forcing everyone to clear the indexes anyway (`rs.indexOnstartup=true`, due to Spring6 changes). I didn't check but that probably means previously-indexed temp documents will no longer be in the index?

rs-fraser:
> Looks good & solves the issue for me on test AWS instance, but I wonder if we actually need the defensive code not displaying previously indexed docs, as with 1.125/2.25 release we are forcing everyone to clear the indexes anyway (`rs.indexOnstartup=true`, due to Spring6 changes). I didn't check but that probably means previously-indexed temp documents will no longer be in the index?

Yeah I was on the fence about keeping it since we need to re-index anyway (and the index will be built from scratch so won't contain those temp docs), so I've removed the check.

PR #1014, opened 2026-08-07
Making "Role" an actual object, per changes made to the DSW API.

mrowe-tan:
## Description ##
- Resolves https://researchspace.atlassian.net/browse/RSDEV-1294
- Treats Role as an actual object now, and not just a string, per changes to the DSW API


mKowalski256 (approved):
Codewise looks good, although I didn't check on the actual build.

PR #1017, opened 2026-08-07
Remove unsafe SuppressWarnings usages

tchoi-rs:
## Description ##
Remove unsafe SuppressWarnings usages. Many of them are no longer needed after Java 17 / Spring 6 upgrade.

### DIFF
--- a/src/main/java/com/axiope/webapp/taglib/AssetUrlTag.java
+++ b/src/main/java/com/axiope/webapp/taglib/AssetUrlTag.java
@@ -73,9 +73,10 @@ public class AssetUrlTag extends TagSupport {
   private Map<String, String> getProductionUrlCache() {
     ServletContext servletContext = pageContext.getServletContext();
     Object existing = servletContext.getAttribute(PRODUCTION_URL_CACHE_ATTR);
-    if (existing instanceof Map) {
-      @SuppressWarnings("unchecked")
-      Map<String, String> cached = (Map<String, String>) existing;
+    if (existing instanceof Map<?, ?> map) {
+      Map<String, String> cached = new ConcurrentHashMap<>();
+      map.forEach((key, value) -> cached.put(String.class.cast(key), String.class.cast(value)));
+      servletContext.setAttribute(PRODUCTION_URL_CACHE_ATTR, cached);
       return cached;
     }
     Map<String, String> created = new ConcurrentHashMap<>();
--- a/src/main/java/com/axiope/webapp/taglib/AssetUrlTag.java
+++ b/src/main/java/com/axiope/webapp/taglib/AssetUrlTag.java
@@ -73,20 +73,26 @@ public class AssetUrlTag extends TagSupport {
   private Map<String, String> getProductionUrlCache() {
     ServletContext servletContext = pageContext.getServletContext();
     Object existing = servletContext.getAttribute(PRODUCTION_URL_CACHE_ATTR);
-    if (existing instanceof Map<?, ?> map) {
-      Map<String, String> cached = new ConcurrentHashMap<>();
-      map.forEach((key, value) -> cached.put(String.class.cast(key), String.class.cast(value)));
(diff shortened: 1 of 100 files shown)


src/main/java/com/axiope/webapp/taglib/AssetUrlTag.java
```diff
@@ -73,9 +73,10 @@ && isLegacyAsset(value)
   private Map<String, String> getProductionUrlCache() {
     ServletContext servletContext = pageContext.getServletContext();
     Object existing = servletContext.getAttribute(PRODUCTION_URL_CACHE_ATTR);
-    if (existing instanceof Map) {
-      @SuppressWarnings("unchecked")
-      Map<String, String> cached = (Map<String, String>) existing;
+    if (existing instanceof Map<?, ?> map) {
+      Map<String, String> cached = new ConcurrentHashMap<>();
+      map.forEach((key, value) -> cached.put(String.class.cast(key), String.class.cast(value)));
+      servletContext.setAttribute(PRODUCTION_URL_CACHE_ATTR, cached);
```
mKowalski256: the changed code is recreating the 'cached' map on every call, rather than caching, I think?

src/main/java/com/researchspace/search/impl/FileIndexer.java
```diff
@@ -43,7 +43,6 @@ public class FileIndexer extends AttachmentSearchBase implements IFileIndexer {
 
   public static final String FIELD_CONTENTS = "contents";
 
-  @SuppressWarnings("unused")
   private static final String[] ENGLISH_STOP_WORDS = {
     "a", "an", "and", "are", "as", "at", "be", "but", "by", "for", "if", "in", "into", "is", "it",
     "no", "not", "of", "on", "or", "such", "that", "the", "their", "then", "there", "these", "they",
```
mKowalski256: this array seems genuinely unused actually - so could be deleted

src/main/java/com/researchspace/service/impl/UserManagerImpl.java
```diff
@@ -82,7 +82,6 @@ public UserManagerImpl(@Autowired UserDao userDao) {
   private @Autowired IVerificationPasswordValidator verificationPasswordValidator;
   private @Autowired MessageSourceUtils messages;
 
-  @SuppressWarnings("unused") // unused but required for spring ApplicationContext to load properly
```
mKowalski256, line 85: 'recordFactory' is not referenced directly, but if it is actually needed for spring to load correctly, then maybe the annotation and comment should stay?

mKowalski256 (commented):
Looks good in general, but I've commented on a few changes that I'm not sure about.

src/main/java/com/researchspace/service/impl/UserManagerImpl.java
tchoi-rs: Spring hasn't blown up in integration tests which is why I suspect this line is not needed.

mKowalski256: this line was explaining why the very next line, declaring unused autowired `recordFactory` field, is needed. Or are you saying you've tried integration test with both lines removed? If latter, then not just the suppression, but also the unused field could be removed.

tchoi-rs: Yes, the entire dep can go (updated in 50659d7)

src/main/java/com/researchspace/search/impl/FileIndexer.java
tchoi-rs: Fixed in 50659d7

src/main/java/com/axiope/webapp/taglib/AssetUrlTag.java
tchoi-rs: Fixed in 50659d7

mKowalski256 (approved):
Looks good!

PR #1019, opened 2026-08-10
RSDEV-1261 Fix duplicated instrument keeps the original instrument's Landing page

mrowe-tan:
## Description ##
- Resolves https://researchspace.atlassian.net/browse/RSDEV-1261
- Solution as per comments in that ticket.

Summary of changes by Claude:

  InstrumentEntityApiManagerImpl.java
  - duplicateInstrument: now calls clearSystemGeneratedLandingPage(dbInstrument,
  copy) before save, then calls fillBlankLandingPage(copy, user) after the
  first save (with a second save when it fills something, matching the 
  createInstrument pattern).
  - New private clearSystemGeneratedLandingPage(source, copy): uses
  GlobalIdUrls.globalIdUrl to build the source's expected default URL, then
  blanks the copy's Landing page field only if it matches -- preserving
  user-typed values.

  InstrumentEntityApiManagerImplLinkFieldTest.java
  - Three new tests covering the three cases from the ticket: system-generated
  URL gets replaced with the copy's own address, an instrument with no Landing
  page field is unaffected (one save only), and a user-typed URL survives
  unchanged. 
  - Added @Mock fields for instrumentDao, invPermissions, containerDao,
  properties, publisher, and userManager, injected in setUp() via
  ReflectionTestUtils.

### DIFF
--- a/src/main/java/com/researchspace/service/inventory/impl/InstrumentEntityApiManagerImpl.java
+++ b/src/main/java/com/researchspace/service/inventory/impl/InstrumentEntityApiManagerImpl.java
@@ -214,6 +214,26 @@ public class InstrumentEntityApiManagerImpl extends InventoryApiManagerImpl<Inst
     return update.applyChangesToDatabaseField(blankLandingPage.get(), user);
   }
 
+  /**
+   * Blanks the Landing page field in {@code copy} when it contains the system-generated default URL
+   * for {@code source} — i.e. the URL RSpace would have auto-filled for the source record. A value
+   * the user typed on the source is left untouched.
+   */
+  private void clearSystemGeneratedLandingPage(Instrument source, Instrument copy) {
+    GlobalIdUrls.globalIdUrl(properties, source.getGlobalIdentifier())
+        .ifPresent(
+            sourceUrl ->
+                copy.getActiveFields().stream()
+                    .filter(f -> f.getType() == FieldType.URI)
+                    .filter(
+                        f ->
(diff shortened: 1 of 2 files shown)


mKowalski256 (commented):
The change looks good, but is the new behaviour actually matching the JIRA specification for the scenario where user has manually provided the landing page on the source instrument?

The JIRA description, in last part, says:
```
The decision is taken: never copy the “Landing Page” when duplicating the instrument 
(becasue the landing page will be different anyways, also if the user originally set their 
own landing page for the original instrument)
```

Which I believe means that user-provided landing page should be always reset on the copy; with current implementation, the user-provided landing page is propagated to the copy (confirmed on [AWS build,](https://rsdev-1261-duplicated-instrument-landing-page-3dc7bc71-4.researchspace.com/), the PR even has a unit test ensuring that)

mrowe-tan:
That would be due to my own understanding of the comments - my understanding was that we wanted to keep and duplicate anything that the user had typed (even if it didn't necessarily make sense); but I could also be reading the intention in the comments incorrectly.  In any event, this solution is probably better for the system-generated landing page case, as this at least means that the user doesn't need to retype the landing page URL every time they do a duplicate of an instrument with a system-generated landing page, and we can review the behaviour with regards the custom-typed landing page later.

mKowalski256 (approved):
fair enough - let's merge the current one and confirm in jira discussion whether user-provided values should also reset on duplication, or be kept.

PR #1020, opened 2026-08-10
RSDEV-1226 Move barcode search out of the filter row into the main search bar

rs-fraser:
## Description ##

Moves Inventory barcode scanning out of the filter row and into the main search bar, as a 1-click scan (RSDEV-1226, supporting RPD-304 biobank workflows).

- The "Barcode" filter-row chip is removed
- The search bar gains a scan icon and, where scanning is available, the placeholder "Search or scan a barcode"
- Clicking the icon opens the camera in a popover; the first detected barcode automatically fills the search box and runs the search
- The QR-only fallback notice now reads "To scan other formats, please use Chrome or a Chromium-based browser."
- Scanning an RSpace item label (a permalink QR) navigates to that item as an in-app route change; `isInventoryPermalink` now requires a same-origin http(s) URL matched on the parsed pathname, so scanned `javascript:` URLs or off-site look-alikes cannot trigger navigation
- A scanned value goes through the normal search, so active filters are preserved (the old chip reset all filters first)

## Scope
Follow-up PRs will address:
1. Aligning other barcode scanners (IGSN and Sample field) with the updated design
2. IGSN scanner not working on non-chromium browsers

## Demo
Old flow:

https://github.com/user-attachments/assets/37384755-6fdc-4001-8235-13589a9178ad

New:

https://github.com/user-attachments/assets/e8522aa9-ca68-4bd3-97f7-5885920053c3


src/main/webapp/ui/src/Inventory/components/BarcodeScanner/BarcodeScannerSkeleton.tsx
```diff
@@ -73,76 +76,115 @@ export default function BarcodeScannerSkeleton({
     }
   }
 
+  /*
+   * The ref guards against the camera re-detecting the same code on a later
+   * interval tick before the popover has finished closing.
+   */
+  const submitted = useRef(false);
+  useEffect(() => {
```
tchoi-rs, line 84: This effect has no dependency array which means this will be run on every render

tchoi-rs (changes requested):

src/main/webapp/ui/src/Inventory/components/BarcodeScanner/BarcodeScannerSkeleton.tsx
rs-fraser: Fixed in https://github.com/rspace-os/rspace-web/pull/1020/commits/51f0dd03b7fb6c0a812d42e4b3c953522bdfa3d5

tchoi-rs (approved):

PR #1027, opened 2026-08-14
PRT-1133 Gallery Image Editing Broken

mrowe-tan:
## Description ##
- Resolves https://researchspace.atlassian.net/browse/PRT-1133

## Summary of changes from Claude: ##

Root cause analysis

  The dialog uses an HTML <canvas> element for both rotation and cropping. The
  Canvas API's toDataURL() and toBlob() methods only natively support image/png
  and image/jpeg output -- any other MIME type causes a silent fallback to PNG.
  This meant that editing any non-PNG/JPEG image (GIF, BMP) always produced a
  PNG blob on save, regardless of the original format.

  Two pre-existing bugs compounded the problem:

  - Bare MIME subtype bug: cropImage was calling canvas.toBlob(callback, format,
  1.0) where format was just the subtype (e.g. "jpeg"), not a full MIME type
  ("image/jpeg"). Unrecognised values fall back to PNG.
  - Stale closure bug (noted but left out of scope): the useEffect captured
  imageType before setImageType had re-rendered, so the initial editorData
  data-URI was built with the previous (initially empty) type string.

  ---
  Changes made

  src/main/webapp/ui/src/components/ImageEditingDialog.tsx
  
- MIME type prefix - Changed canvas.toBlob(…, format, …) and canvas.toDataURL("image/" + imageType, …) to use full MIME types ("image/" + format / template literal equivalents)
- GIF encoding - Added encodeAsGif(ctx, canvas) -- reads raw pixel data via getImageData, quantizes to a 256-colour palette using gifenc, encodes a single GIF frame, returns a correctly-typed Blob
- BMP encoding - Added encodeAsBmp(ctx, canvas) -- manually constructs a valid BMP binary: 14-byte file header + 40-byte encoding BITMAPINFOHEADER, pixel rows written bottom-up in BGR order with 4-byte row padding                                     - BMP MIME and variants - The BMP branch matches "bmp", "x-bmp", "x-ms-bmp", "octet-stream" to handle variation in how browsers report the MIME type of BMP files

  src/main/webapp/ui/src/types/gifenc.d.ts (new file)

  Added a minimal ambient type declaration for gifenc (which ships no .d.ts
  files), covering GIFEncoder, quantize, and applyPalette.

  Dependency added: gifenc (via pnpm add gifenc)

  ---
  Formats and their code path after the fix

- JPEG, PNG, WebP - canvas.toBlob("image/<format>") -- native browser support
- GIF - encodeAsGif -- gifenc library encodes canvas pixel data as single-frame GIF 
- BMP - encodeAsBmp -- fast-bmp library used to write BMP binary directly from canvas pixel data

## Summary of changes by Claude for subsequently-discovered Hibernate issues: ##

Bug 1: Gallery fails to load after 20 image edits

  Root cause. getPaginatedChildRecordsOfParentWithFilter in RecordDaoHibernate
  uses setMaxResults / setFirstResult to page results. When a user edits the
  same image 20 times, the 21st copy pushes the total count over the default
  page size. Any EcatImage whose originalImage was created before the current
  page window would then be a lazy proxy outside the current session, and any
  subsequent access to it (including serialisation for the UI response) would
  throw.

  Fix. A second HQL constant, RECORDS_IN_FOLDER_QUERY_WITH_IMAGE_FETCH, was
  added to RecordDaoHibernate. It extends the base query with a LEFT JOIN FETCH
  that eagerly initialises originalImage within the same session:

  private static final String RECORDS_IN_FOLDER_QUERY_WITH_IMAGE_FETCH =
      " from BaseRecord br inner join br.parents c"
          + " left join fetch treat(br as EcatImage).originalImage"
          + " where c.recordInFolderDeleted=false and br.deleted=false and 
  c.folder.id=:parentId ";

  TREAT is required to downcast BaseRecord to EcatImage inside the HQL join --
  without it Hibernate cannot resolve the subtype-specific association. A LEFT 
  JOIN FETCH on a @ManyToOne is safe to combine with pagination (unlike
  @OneToMany, it does not cause Hibernate to apply the row limit in memory).

  getPaginatedChildRecordsOfParentWithFilter was updated to use
  RECORDS_IN_FOLDER_QUERY_WITH_IMAGE_FETCH for the data query; the count query
  continues to use the original RECORDS_IN_FOLDER_QUERY.

  Secondary fix: ORDER BY ambiguity. Adding the fetch join introduced a second
  from-element that also exposes editInfo, causing Hibernate to reject
  unqualified ORDER BY paths with:

  Multiple from-elements expose unqualified attribute: editInfo

  makeOrderBy was updated to prefix all ORDER BY paths with the br. alias. A
  EDIT_INFO_ORDER_BY_FIELDS set was introduced to identify fields that need the
  additional editInfo. component. Unit assertions in RecordDaoHibernateTest were
  updated to match the new qualified paths.

  ---
  Bug 2: LazyInitializationException when uploading a new version

  Root cause. When a user uploads a new version of image B (a copy of image A),
  GalleryController.saveMediaFile calls mediaManager.saveMediaFile(...) and then
  immediately calls media.toRecordInfo(). toRecordInfo() on an EcatImage
  accesses originalImage to build the response. If image A falls outside the
  current gallery page, its proxy was never initialised, and the session used by
  saveMediaFile had already closed by the time toRecordInfo() executed.

  Fix. Three new methods were added across the DAO, service, and controller
  layers:

  - EcatImageDao.getWithOriginalImage(Long id) -- issues a single HQL query with
  left join fetch i.originalImage to return a fully initialised EcatImage.
  - MediaManager.getImageWithOriginalImage(Long imageId, User user) -- calls the
  DAO method and checks read permission.
  - In GalleryController.saveMediaFile, after saveMediaFile returns, the result
  is reloaded via getImageWithOriginalImage before toRecordInfo() is called:

  if (media instanceof EcatImage ecatImage) {
      media = mediaManager.getImageWithOriginalImage(ecatImage.getId(),
  subject);
  }
  return new AjaxReturnObject<>(media.toRecordInfo(), null);

  This reload is a targeted single-row fetch with the association
  pre-initialised, so it is cheap and does not affect the behaviour for
  non-image media types.


src/main/webapp/ui/src/components/ImageEditingDialog.tsx
```diff
@@ -108,7 +160,7 @@ function ImageEditingDialog({
         ctx.rotate(((direction === "clockwise" ? 90 : -90) * Math.PI) / 180);
         ctx.drawImage(image, -image.naturalWidth / 2, -image.naturalHeight / 2);
       }
-      return canvas.toDataURL(imageType, 1.0);
+      return canvas.toDataURL(`image/${imageType}`, 1.0);
     };
```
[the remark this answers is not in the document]
mrowe-tan: Having tested the code, this does not appear to be an issue.  Adding the "image/" prefix when building the editorData there and only there does not resolve the initial problem that this ticket is addressing - at least not for all image formats.

src/main/webapp/ui/src/components/ImageEditingDialog.tsx
```diff
@@ -28,6 +29,57 @@ const readAsBinaryString = (file: Blob): Promise<string> =>
...
+const encodeAsGif = (ctx: CanvasRenderingContext2D, canvas: HTMLCanvasElement): Blob => {
+  const { data, width, height } = ctx.getImageData(0, 0, canvas.width, canvas.height);
+  const palette = quantize(data, 256);
+  const index = applyPalette(data, palette);
+  const gif = GIFEncoder();
+  gif.writeFrame(index, width, height, { palette });
+  gif.finish();
+  return new Blob([gif.bytes()], { type: "image/gif" });
+};
+const encodeAsBmp = (ctx: CanvasRenderingContext2D, canvas: HTMLCanvasElement): Blob => {
```
tchoi-rs, line 42: I would not reinvent the wheel here and instead use https://github.com/image-js/fast-bmp

src/main/webapp/ui/src/components/ImageEditingDialog.tsx
tchoi-rs: This is also missing the `image/` prefix before `imageType`

package.json
```diff
@@ -113,6 +113,7 @@
     "dompurify": "3.4.12",
     "email-validator": "2.0.4",
     "es-toolkit": "1.50.0",
+    "gifenc": "^1.0.3",
```
tchoi-rs: Could you remove the caret here?

tchoi-rs (changes requested):
Overall looks fine, a few small issues:

rs-nicof:
Backend code looks good to me! ✅

rs-nicof (approved):

tchoi-rs (approved):

PR #1030, opened 2026-08-18
RSDEV-1307 Clear a manually entered Landing page on derived instruments

rs-nicof:
## Description ##
- Resolves https://researchspace.atlassian.net/browse/RSDEV-1307
- Follow-up to RSDEV-1261, which cleared the Landing page only when it matched the auto-generated value for the source instrument. A value the user had typed by hand was still carried over. A landing page names exactly one physical instrument, so it must be discarded on every derivation path regardless of where the value came from.

### Backend

- InstrumentEntityApiManagerImpl: clearSystemGeneratedLandingPage is replaced by an unconditional clearLandingPage, applied on all three derivation paths (duplicate instrument, duplicate template, create instrument from template). On a concrete instrument the blank is then refilled with the record's own address; on a template it stays blank, since stamping one instrument's address onto a reusable definition would hand it to every instrument later created from that template.
- The create-from-template path also discards a Landing page that the request echoes back unchanged from the template. Without this, any client that read a template and posted its fields verbatim re-established the template's value, leaving the guarantee dependent on client cooperation rather than on the service. A value the user supplies for the new record is still kept and still validated, so a malformed URI is rejected.
- New package-private PidinstFields helper holds the single copy of the "Landing page" name-and-type predicate, which previously existed as three byte-identical copies in the same package. Follows the GlobalIdUrls precedent. Only the Inventory UI's copy now remains, and it cannot be shared.
- The field-count mismatch error is externalized to server.inventory.json as errors.inventory.instrument.fieldCountMismatch.

### Frontend

- InstrumentModel.setTemplate blanks the inherited Landing page so the creation form shows what will actually be saved. Only the first matching field is blanked, mirroring the backend, which resolves and refills a single field.

### Tests

- Nine new fast backend tests and two new frontend tests, covering the echoed-value case, the user-typed boundary, template duplication (which previously had only DB-backed coverage), name-based rather than positional resolution, the absent-server-URL case, and malformed URI rejection.
- Removed one redundant MVCIT test and a redundant assertion.

CONTEXT.md records the Landing page as an identity-bound field, along with the paths deliberately left out of scope: template-version sync, creating a template from an instrument (opt-in per field), and the absence of a backfill for records derived before this change, since a blanket update cannot distinguish an inherited value from one a user legitimately typed.

### DIFF
--- a/src/main/webapp/ui/biome.jsonc
+++ b/src/main/webapp/ui/biome.jsonc
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://biomejs.dev/schemas/2.5.5/schema.json",
+  "$schema": "https://biomejs.dev/schemas/2.5.6/schema.json",
   "vcs": {
     "enabled": true,
     "clientKind": "git",
--- a/src/main/webapp/ui/src/stores/models/InstrumentModel.tsx
+++ b/src/main/webapp/ui/src/stores/models/InstrumentModel.tsx
@@ -69,6 +69,21 @@ export type InstrumentAttrs = {
   _links: Array<_LINK>;
 } & Record<string, unknown>;
 
+/**
+ * The name of the field holding an instrument's public landing page. Matched case-insensitively
+ * and ignoring surrounding whitespace, together with the field's URI type.
+ *
+ * This rule is duplicated, not shared: the backend's copy lives in `PidinstFields` (Java), which
+ * the service layer and the PIDINST mapping both resolve through. Nothing detects a divergence
+ * between that copy and this one, so a rename must be applied to both in lock-step.
+ */
+export const LANDING_PAGE_FIELD_NAME = "Landing page";
+
+const LANDING_PAGE_FIELD_NAME_LOWERCASE = LANDING_PAGE_FIELD_NAME.toLowerCase();
+
+const isLandingPageField = (field: Field): boolean =>
+  field.type === "uri" && field.name?.trim().toLowerCase() === LANDING_PAGE_FIELD_NAME_LOWERCASE;
+
 const FIELDS = new Set([...RESULT_FIELDS].concat(["fields", "template"]));
 const defaultVisibleFields = new Set([...FIELDS, ...defaultVisibleResultFields]);
 const defaultEditableFields = new Set<string>();
@@ -105,7 +120,9 @@ export default class InstrumentModel
 
   constructor(factory: Factory, params: InstrumentAttrs) {
     super(factory, params);
-    makeObservable(this, {
+    // the explicit key list is what lets a private member be annotated (MobX types the plain
+    // overload against the public shape only)
+    makeObservable<this, "blankLandingPageField">(this, {
       parentContainers: observable,
       immediateParentContainer: observable,
       createOptionsParametersState: observable,
@@ -115,6 +132,7 @@ export default class InstrumentModel
       templateVersion: observable,
       setTemplate: action,
       overrideFields: action,
+      blankLandingPageField: action,
       paramsForBackend: override,
       populateFromJson: override,
       updateFieldsState: override,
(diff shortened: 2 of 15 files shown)


rs-nicof:
@greptileai

src/main/webapp/ui/biome.jsonc
```diff
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://biomejs.dev/schemas/2.5.5/schema.json",
+  "$schema": "https://biomejs.dev/schemas/2.5.6/schema.json",
```
tchoi-rs, line 2: I'm not sure if it's intentional but it's more likely I forgot to bump this when upgrading dependencies (sorry) - should be fine to keep

src/main/webapp/ui/src/stores/models/InstrumentModel.tsx
```diff
@@ -69,6 +69,21 @@ export type InstrumentAttrs = {
...
+ *
+ * This rule is duplicated, not shared: the backend's copy lives in `PidinstFields` (Java), which
+ * the service layer and the PIDINST mapping both resolve through. Nothing detects a divergence
+ * between that copy and this one, so a rename must be applied to both in lock-step.
+ */
+export const LANDING_PAGE_FIELD_NAME = "Landing page";
+
+const LANDING_PAGE_FIELD_NAME_LOWERCASE = LANDING_PAGE_FIELD_NAME.toLowerCase();
+
+const isLandingPageField = (field: Field): boolean =>
```
tchoi-rs, line 84: This is a bit weird, would that mean users are supposed to create their own custom field called "Landing page"? Is the field name validation case-insensitive as well so that I can't add a duplicate field with random capitalisation?

tchoi-rs (approved):
Frontend portion looks fine!

src/main/webapp/ui/src/stores/models/InstrumentModel.tsx
rs-nicof: Good questions, and the second one turned up something real.

**On the field name**: users aren't expected to create it themselves. It ships with RSpace as part of the locked default PIDINST instrument template, which is seeded into every deployment, owned by the sysadmin, and flagged read-only. That lock is genuinely enforced, so nobody can rename or remove the field on the canonical template, not even the sysadmin who owns it. Users get the field by creating an instrument from that template.

Worth adding that this PR didn't introduce matching by field name. The PIDINST mapping already resolved all of its fields that way, and this change just pulled that rule into one shared place so the clearing, the refilling and the PIDINST mapping can't drift onto different fields.

Two caveats I'll own rather than defend. Users can duplicate the locked template, and the copy is editable, so if someone renames the field in their own copy the behaviour silently stops applying with no warning. Likewise, anyone who hand-builds a template with a URI field of that name gets the identity-bound treatment whether they intended it or not. There's a note in `CONTEXT.md` about the same fragility if the labels are ever translated.

**On case-insensitivity**, the answer splits between the two layers:

_The backend does compare case-insensitively_, after trimming whitespace, and it runs on every instrument and template path at both the validation and the service layer. So you cannot save two fields whose names differ only by capitalisation. That's actually what makes this PR safe: because only one matching field can ever exist in stored data, resolving "the first match" is unambiguous.
_The frontend does not_. Its own duplicate-name check compares the raw strings, with no lowercasing and no trimming, so the form will happily accept "landing page" alongside "Landing page" and you only discover the clash when you hit save and the server rejects it.

That frontend gap predates this change and applies to all Inventory record types rather than just instruments, so I'd rather not widen this PR to cover it. 

Happy to raise a separate ticket to bring the client-side check in line with the server's comparison if you agree.

(Claude helped me generating this reply 🤗 )

rs-nicof:
✅  Tested successfully in AWS instance: https://rsdev-1307-do-not-duplicate-landing-page-ddccf2aa-3.researchspace.com

PR #1033, opened 2026-08-19
PRT-1135: stop MUI portals being re-parented, which destroyed open di…

nhanlon2:
Fixes [PRT-1135](https://researchspace.atlassian.net/browse/PRT-1135).

## Description ##

Importing a DMP from the Gallery create menu could make the import dialog vanish on its own, leaving the create menu stranded open on screen. Reported against DMPTool and DSW, reproduced on the first import after a fresh page load.

**The cause is not in the Gallery.** `DialogBoundary` handed MUI's `Portal` a brand new `container` arrow on every render:

```tsx
<MuiMenu container={() => modalContainer.current} ... >
```

`Portal` keys a layout effect on that prop and re-resolves its mount node whenever the value changes, falling back to `document.body` when the boundary ref reads `null`:

```js
useEnhancedEffect(() => {
  setMountNode(getContainer(container) || document.body);
}, [container, disablePortal]);
```

So it ran on every render, and the ref genuinely does read `null` during the layout phase while a boundary is still mounting. Pointing a portal at a different node does not move it, React **rebuilds everything inside it**.

The DMP import dialog lives inside the create menu's portal, so it was destroyed along with the rest of that subtree. Destroyed rather than closed, so its `setOpen(false)` never ran, and that callback was the only thing clearing the create menu's anchor. Hence: dialog gone, menu still there, still interactive.

Memoising the getter resolves the mount node once and stops it moving. `Dialog`, `Menu` and `Drawer` all shared the defect, so all three use it.

## MUI 9.3.1, and removing the PRT-1118 workaround

PRT-1118 patched a cancelled exit transition ([mui/material-ui#32286](https://github.com/mui/material-ui/issues/32286)) by making the closed create menu click-through. A stuck Modal with an invisible backdrop otherwise swallowed every click and froze the page until reload.

MUI 9.3 fixes that upstream, so this bumps `@mui/material` and `@mui/icons-material` to 9.3.1 and `@mui/system` and `@mui/styled-engine` to 9.3.0, and deletes the workaround. The menu gets its close animation back.

**The bump and the deletion have to ship together.** Dropping the workaround on 9.2.0 would reinstate the freeze. Confirmed by hand against a production build: Escape and click-away no longer freeze the Gallery.

The browser test that guarded the workaround goes too. It asserted `pointer-events` was not `none` on the open menu. Nothing sets `pointer-events` on the menu any more, so it can no longer fail.

## On the review comment about using state

@tchoi-rs suggested `useState` with `ref={setModalContainer}` rather than a ref. That is the better shape and I tried it. It does not work here yet, and the reason is worth recording.

It breaks **61 tests across 6 files**, every one of them with the same signature:

> Unable to find role "dialog". There are no accessible roles. But there might be some inaccessible roles.

That is `aria-hidden`. Eleven components wrap their own dialog in a boundary:

```tsx
<DialogBoundary>
  <Dialog open>...</Dialog>
</DialogBoundary>
```

Boundary and dialog appear in the same commit, so the dialog lands on `document.body` first, `ModalManager` marks the rest of the body `aria-hidden`, and only then does the portal move it into the boundary, which is by then inside the hidden subtree. MUI never recomputes.

The ref version does the same move, but only when something happens to trigger another render, which those tests never do. State makes it happen every time, so the tests suddenly notice a problem that was already there.

Three attempts to fix it in this PR all broke the Gallery in a production build:

| Attempt | Result |
|---|---|
| Withhold the boundary's children until the div attaches | first click on Create does nothing |
| Withhold each modal for one commit (three-state context) | React error #185, infinite `Portal.setMountNode` loop |
| Build the container with `document.createElement` up front | Gallery renders blank |

So this PR keeps the memoised getter, which is verified working. The real fix is to stop those eleven components wrapping their own dialog and rely on the app-level boundary instead, which removes the race rather than racing it. That needs each mount point traced and each dialog checked by hand, and it belongs in its own ticket.

## Design decisions

**It memoises a getter, not a value.** `useCallback(() => modalContainer.current, [modalContainer])` keeps the identity stable while every call still reads `.current` fresh, so it cannot return a stale node. `useMemo(() => modalContainer.current, [])` would have captured the node and been genuinely unsafe.

**The dependency is the ref object, not the context value.** `DialogBoundary` recreates its context value each render, but the `useRef` inside is stable per instance, so nested boundaries still resolve and components outside any boundary get the stable module-level default.

**The change is app-wide by necessity.** Every `Dialog`, `Menu` and `Drawer` in RSpace went through the same unstable container, so any of them could silently lose state held inside.

## Testing notes

The new `DialogBoundary` test was confirmed to fail against the unfixed code (`expected 2 to be 1`, the portal's child remounting on a parent re-render), so it is not vacuous.

Manual, against a production build (`-DgenerateReactDist`; this does not reproduce under `reactDevMode`, which masks the timing):

1. Enable a DMP source under Apps.
2. Load the Gallery fresh.
3. Create, DMPTool, select a plan, Import.
4. The dialog stays open until Close, and Close dismisses both the dialog and the create menu.
5. Open the create menu and dismiss it with Escape and with a click-away. The page stays interactive.

Diagnosis used temporary unmount probes in a production build, narrowing the teardown boundary until a layout-phase probe caught `modalContainer.current` reading `null`. All probes removed.

**Not addressed here**, each worth its own ticket:

- The eleven components that wrap their own dialog in a boundary, described above. This is the real fix for the `aria-hidden` problem and would also make the memoised getter unnecessary.
- `DialogBoundary` writes `document.body.style.overflow` from three separate components with no cleanup, so nested modals fight over it and anything unmounting while open leaves the page unscrollable. MUI already locks scrolling itself.
- The DMP import dialogs are still rendered as children of the create menu. This PR removes the trigger we found, but the structural coupling remains.
- `ConfirmationDialog.tsx:152` throws `Cannot set properties of undefined (setting 'createConfirmationDialog')` on every Gallery load, because `window.RS` does not exist on that page.


nhanlon2:
@greptile

nhanlon2:
@greptileai

nhanlon2:
@greptileai

src/main/webapp/ui/src/components/DialogBoundary.tsx
tchoi-rs, line 82: It might be better to change this into a state and use `ref={setModalContainerState}` instead

tchoi-rs (changes requested):
You mentioned the bug https://github.com/mui/material-ui/issues/32286 is a culprit of the behaviour. It's supposedly fixed in https://github.com/mui/material-ui/pull/48881 which is now released as 9.3.1; does an update help?

src/main/webapp/ui/src/components/DialogBoundary.tsx
nhanlon2: Have bumped MUI. This fixes PRT-1118 by itself and so have deleted the code change for that and the test you wrote. However, I cant implement your suggestion on DialogBoundary. This reveals as accesibility bug that exists in production and that DialogBoundary change reveals it in broken tests. Claude says the following dialogs use DialogBoundary wrongly and cause screen readers to lose access. A full fix would require modifications to all of these files:

eln-dmp-integration/DSW/DSWImportDialog.tsx
eln-dmp-integration/Argos/DMPDialog.tsx
eln-dmp-integration/DMPAssistant/DMPDialog.tsx
eln-dmp-integration/DMPOnline/DMPDialog.tsx
eln-dmp-integration/DMPTool/DMPDialog.tsx
eln-inventory-integration/MaterialsListing/MaterialsDialog.tsx
components/ShareDialog.tsx
Toolbar/Workspace/RenameDialog.tsx
Toolbar/Workspace/CompareDialog.tsx
Toolbar/Workspace/TagDialog.tsx
tinyMCE/stoichiometry/StoichiometryDialogEntrypoint.tsx

(Some other fix needed for the following 5)
Boundary at the top of an app, which is the intended use, 5 files

components/Alerts/Alerts.tsx
eln/apps/App.tsx
eln/AppBar.tsx
eln/about/index.tsx
tinyMCE/gallery/GalleryEntrypoint.tsx

nhanlon2: Created this ticket to explain the full fix that seems to be required beyond just making the Gallery Create menu to work and fixing the issue in this PRT:
https://researchspace.atlassian.net/browse/RSDEV-1317

PR #1034, opened 2026-08-19
RSDEV-1306 Fix sidebar nav leaving stale record details in single-column layout

rs-fraser:
## Description ##
- In single-column layout (right panel hidden, or small viewports), clicking a record type in the Inventory sidebar left the previously selected record's details on screen instead of showing the listing.
- Cause: the sidebar's shared click handler unconditionally set the visible panel to "right" after every click. That was added for the Create menu but was also attached to the record-type tabs, overriding the navigation default of returning to the listing.
- Fix: remove the override from the shared handler. The create-new flow already shows the right panel itself (SearchStore's createNewHelper), so the Create menu is unaffected. Added a regression test asserting a record-type nav click does not touch the visible panel.


rs-fraser:
Superseded by a PR from the same-repo branch, following the new branch layout.

PR #1039, opened 2026-08-20
Fix various issues surrounding DialogBoundary

tchoi-rs:
## Description ##
Improves code that uses `<DialogBoundary>`, as well as related bugfixes. 

Fixes PRT-1135


nhanlon2 (commented):
**Claude's review** 

Critical
<Suspense> inside Menu interferes with MUI MenuList child detection — [DialogBoundary.tsx:154](vscode-webview://0f2b3lokhktqucm81sp56j8tgh76v1g53tvjg1f4lpaeq7l6sp7t/src/main/webapp/ui/src/components/DialogBoundary.tsx#L154)
MenuList walks React.Children to compute the active item and clone it with autoFocusItem/tabIndex; with a single Suspense child that detection misses (upstream limitation [mui#35786](https://github.com/mui/material-ui/issues/35786)). Two reviewers disagree on severity: one says keyboard-open first-item focus breaks for every menu using this primitive; the other checked all four current consumers (SidebarCreateMenu, gallery ActionsMenu, AppBar, HelpDocs) and found none use selected items, so nothing visibly regresses today (arrow-key DOM navigation is unaffected since Suspense adds no DOM node). Both agree the wrapper is also redundant: [Sidebar.tsx:456](vscode-webview://0f2b3lokhktqucm81sp56j8tgh76v1g53tvjg1f4lpaeq7l6sp7t/src/main/webapp/ui/src/eln/gallery/components/Sidebar.tsx#L456) already has the per-callsite Suspense that actually catches the i18next suspension, and a Menu-level catch would blank the whole menu, which is what the inner one exists to prevent. Recommended fix: drop the Suspense from the Menu wrapper only (the Dialog/Drawer ones at lines 138/179 are structurally safe), or at minimum document the constraint. Worth a manual keyboard check before deciding.

New devDependency @suspensive/react is unnecessary — [package.json:39](vscode-webview://0f2b3lokhktqucm81sp56j8tgh76v1g53tvjg1f4lpaeq7l6sp7t/package.json#L39), used only at [DialogBoundary.spec.tsx:9](vscode-webview://0f2b3lokhktqucm81sp56j8tgh76v1g53tvjg1f4lpaeq7l6sp7t/src/main/webapp/ui/src/components/DialogBoundary.spec.tsx#L9)
Flagged independently by three reviewers. The spec uses only fallback + children, which is exactly React's built-in Suspense (already imported in the same file). Violates the repo rule "do not add dependencies unless necessary". Replace with import { Suspense } from "react" and drop the dep and lockfile entry, unless there is an undocumented Vitest Browser Mode reason, in which case document it.

Important
Dead code: GalleryPage.openDSWImport has no callers — [GalleryPage.ts:112-115](vscode-webview://0f2b3lokhktqucm81sp56j8tgh76v1g53tvjg1f4lpaeq7l6sp7t/src/main/webapp/ui/src/__tests__/e2e/pageObjects/gallery/GalleryPage.ts#L112-L115). Its only consumer now calls openCreateMenu() + clickDSWImport() directly; rg confirms only the definition remains. Delete one of the two.

Unreachable Dialog: ArgosDMPDialog table entry — [DmpImportDialogs.tsx:40](vscode-webview://0f2b3lokhktqucm81sp56j8tgh76v1g53tvjg1f4lpaeq7l6sp7t/src/main/webapp/ui/src/eln-dmp-integration/DmpImportDialogs.tsx#L40) vs [:158-163](vscode-webview://0f2b3lokhktqucm81sp56j8tgh76v1g53tvjg1f4lpaeq7l6sp7t/src/main/webapp/ui/src/eln-dmp-integration/DmpImportDialogs.tsx#L158-L163). The argos early return means the table's entry is never read; two dispatch mechanisms for one source. Widening the table's Dialog type to accept an optional onImport collapses the special case.

Leftover redundant DialogBoundary wrappers in the story file — [ShareDialog.story.tsx](vscode-webview://0f2b3lokhktqucm81sp56j8tgh76v1g53tvjg1f4lpaeq7l6sp7t/src/main/webapp/ui/src/components/ShareDialog.story.tsx) (8 occurrences, lines 16-139), each nested directly inside <Alerts>, which already provides the boundary ([Alerts.tsx:49](vscode-webview://0f2b3lokhktqucm81sp56j8tgh76v1g53tvjg1f4lpaeq7l6sp7t/src/main/webapp/ui/src/components/Alerts/Alerts.tsx#L49)). This is the exact pattern the diff deleted from production, so the tests now exercise a wrapper depth production no longer has. (The two survivors in eln/apps/App.tsx and eln/about/index.tsx are correctly kept; they are not inside Alerts.)

Scroll lock applies to non-modal permanent Drawers — [DialogBoundary.tsx:166](vscode-webview://0f2b3lokhktqucm81sp56j8tgh76v1g53tvjg1f4lpaeq7l6sp7t/src/main/webapp/ui/src/components/DialogBoundary.tsx#L166). useBodyScrollLock(open) is not gated on variant === "temporary" (the container prop two lines below is). One reviewer reads this as the Gallery desktop sidebar (permanent, always open) pinning body { overflow: hidden } for the whole session; the other notes it faithfully preserves pre-existing behavior. Either way the gate is a one-line fix and this refactor was the natural moment.

window.RS ??= {} contradicts the ambient type — [ConfirmationDialog.tsx:158](vscode-webview://0f2b3lokhktqucm81sp56j8tgh76v1g53tvjg1f4lpaeq7l6sp7t/src/main/webapp/ui/src/components/ConfirmationDialog.tsx#L158) vs the non-optional interface Window { RS: RSGlobal } at [:23-25](vscode-webview://0f2b3lokhktqucm81sp56j8tgh76v1g53tvjg1f4lpaeq7l6sp7t/src/main/webapp/ui/src/components/ConfirmationDialog.tsx#L23-L25). The runtime fix itself is correct (all merged RSGlobal members are optional, and ??= cannot clobber the legacy JSP namespace), but declare RS?: RSGlobal so the guard is honest with the type system.

DmpImportMenuSection has zero direct test coverage — [DmpImportDialogs.tsx:92-104](vscode-webview://0f2b3lokhktqucm81sp56j8tgh76v1g53tvjg1f4lpaeq7l6sp7t/src/main/webapp/ui/src/eln-dmp-integration/DmpImportDialogs.tsx#L92-L104). Flagged by both coverage and quality reviewers. The integration-enablement filter, DSW multi-connection fan-out, and render-nothing branch decide whether users can import DMPs at all, yet the Sidebar stories substitute a hand-rolled MenuItem ([Sidebar.story.tsx:87,114-116](vscode-webview://0f2b3lokhktqucm81sp56j8tgh76v1g53tvjg1f4lpaeq7l6sp7t/src/main/webapp/ui/src/eln/gallery/components/Sidebar.story.tsx#L114-L116)), so no test renders the real component. An inverted filter or key mismatch would silently remove four integrations' menu entries with everything green.

DmpImportDialogs dispatch for dsw and argos branches untested — [DmpImportDialogs.tsx:144-163](vscode-webview://0f2b3lokhktqucm81sp56j8tgh76v1g53tvjg1f4lpaeq7l6sp7t/src/main/webapp/ui/src/eln-dmp-integration/DmpImportDialogs.tsx#L144-L163). Specs only exercise the dmptool fallthrough. In particular, dropping onImport={onImport} from the Argos branch (line 159) would break Gallery refresh after Argos import, the one surviving piece of the removed window.gallery contract, and no test would fail.

nhanlon2 (approved):

PR #1043, opened 2026-08-21
RSDEV-1318-bump-inventory-revision-eln-path

nhanlon2:
## Summary

[RSDEV-1318](https://researchspace.atlassian.net/browse/RSDEV-1318): registering a subsample usage (a stock decrement, e.g. from an ELN List of Materials with "update inventory quantity" enabled) now counts as a content edit. It bumps the subsample's user-facing version and appears as an entry in the subsample's revision history, exactly like a quantity edit made through `PUT /subSamples/{id}`.

Previously a usage decrement changed the stored quantity silently: the quantity changed but the version stayed the same and no revision was recorded, so the revision history did not explain where the stock went.

## Changes

- `SubSampleApiManagerImpl.registerApiSubSampleUsage` now calls `increaseVersion()` when it applies a non-zero decrement. A zero usage still returns early and remains a complete no-op (no quantity change, no version bump, no revision).
- Javadoc added to `SubSampleApiManager.registerApiSubSampleUsage` documenting the versioning behaviour.

## Tests

- `SubSampleApiManagerTest.saveUsagesOfSubSample`: asserts the version bumps on each non-zero decrement and that a zero usage is a no-op; `saveUpdateNewBasicSubSample` updated for the new expected version.
- New `ListOfMaterialsApiControllerMVCIT.stockDecrementViaLomBumpsSubSampleVersionAndAddsRevision`: end-to-end through the List of Materials API, asserts the quantity drops, the version bumps, and a new revision-history entry is returned by `/subSamples/{id}/revisions`.

Both classes pass locally (`SubSampleApiManagerTest` 17 run, 0 failures; the MVCIT class green across all three integration-test executions with `-Denvironment=drop-recreate-db`).

### DIFF
--- a/src/main/java/com/researchspace/service/inventory/SubSampleApiManager.java
+++ b/src/main/java/com/researchspace/service/inventory/SubSampleApiManager.java
@@ -70,8 +70,9 @@ public interface SubSampleApiManager extends InventoryApiManager<SubSample> {
 
   /**
    * Reduces the subsample's quantity by the given used amount (unit-aware, clamped at zero). A
-   * non-zero usage is a content edit: it bumps the subsample's user-facing version and is recorded
-   * in its revision history (RSDEV-1318). A zero usage is a complete no-op.
+   * usage that changes the stored quantity is a content edit: it bumps the subsample's user-facing
+   * version and is recorded in its revision history (RSDEV-1318). A usage that leaves the quantity
+   * unchanged (zero usage, or usage against already-empty stock) is a complete no-op.
    *
    * @return the updated subsample
    */
--- a/src/main/java/com/researchspace/service/inventory/SubSampleApiManager.java
+++ b/src/main/java/com/researchspace/service/inventory/SubSampleApiManager.java
@@ -69,10 +69,11 @@ public interface SubSampleApiManager extends InventoryApiManager<SubSample> {
   ApiSubSample addNewApiSubSampleToSample(ApiSubSample incomingSubSample, Long sampleId, User user);
 
   /**
-   * Reduces the subsample's quantity by the given used amount (unit-aware, clamped at zero). A
-   * usage that changes the stored quantity is a content edit: it bumps the subsample's user-facing
-   * version and is recorded in its revision history (RSDEV-1318). A usage that leaves the quantity
-   * unchanged (zero usage, or usage against already-empty stock) is a complete no-op.
+   * Reduces the subsample's quantity by the given used amount (unit-aware, clamped at zero).
+   * Requires edit permission on the subsample. A usage that changes the stored quantity is a
+   * content edit: it bumps the subsample's user-facing version and is recorded in its revision
+   * history (RSDEV-1318). A usage that leaves the quantity unchanged (null or zero usage, or usage
+   * against already-empty stock) makes no change to the stored subsample.
    *
    * @return the updated subsample
    */
--- a/src/main/java/com/researchspace/service/inventory/SubSampleApiManager.java
+++ b/src/main/java/com/researchspace/service/inventory/SubSampleApiManager.java
@@ -68,6 +68,13 @@ public interface SubSampleApiManager extends InventoryApiManager<SubSample> {
    */
   ApiSubSample addNewApiSubSampleToSample(ApiSubSample incomingSubSample, Long sampleId, User user);
 
+  /**
+   * Reduces the subsample's quantity by the given used amount (unit-aware, clamped at zero). A
+   * non-zero usage is a content edit: it bumps the subsample's user-facing version and is recorded
+   * in its revision history (RSDEV-1318). A zero usage is a complete no-op.
+   *
+   * @return the updated subsample
+   */
   ApiSubSample registerApiSubSampleUsage(Long subsampleId, QuantityInfo usedQuantity, User user);
 
   ApiSubSample addSubSampleNote(Long subSampleId, ApiSubSampleNote subSampleNote, User user);
--- a/src/main/java/com/researchspace/service/inventory/impl/SubSampleApiManagerImpl.java
+++ b/src/main/java/com/researchspace/service/inventory/impl/SubSampleApiManagerImpl.java
@@ -178,7 +178,7 @@ public class SubSampleApiManagerImpl extends InventoryApiManagerImpl<SubSample>
               user);
 
       if (contentChanged) {
-        // only content edits bump the user-facing version; moves, notes and usage don't
+        // only content edits bump the user-facing version; moves and notes don't
         dbSubSample.increaseVersion();
         registerSubSampleModification(user, dbSubSample);
       }
@@ -281,6 +281,8 @@ public class SubSampleApiManagerImpl extends InventoryApiManagerImpl<SubSample>
       }
 
       dbSubSample.setQuantity(newQuantity);
+      // RSDEV-1318: a stock decrement is a content edit, so it bumps the user-facing version
+      dbSubSample.increaseVersion();
       registerSubSampleModification(user, dbSubSample);
       dbSubSample = subSampleDao.save(dbSubSample);
 
--- a/src/main/java/com/researchspace/service/inventory/impl/SubSampleApiManagerImpl.java
+++ b/src/main/java/com/researchspace/service/inventory/impl/SubSampleApiManagerImpl.java
@@ -264,7 +264,7 @@ public class SubSampleApiManagerImpl extends InventoryApiManagerImpl<SubSample>
       Long subsampleId, QuantityInfo usedQuantity, User user) {
 
     SubSample dbSubSample = getIfExists(subsampleId);
-    if (usedQuantity.getNumericValue().equals(BigDecimal.ZERO)) {
+    if (usedQuantity.getNumericValue().signum() == 0) {
       return getPopulatedApiSubSampleFull(dbSubSample, user);
     }
 
@@ -280,11 +280,17 @@ public class SubSampleApiManagerImpl extends InventoryApiManagerImpl<SubSample>
         newQuantity.setNumericValue(BigDecimal.ZERO);
       }
 
-      dbSubSample.setQuantity(newQuantity);
-      // RSDEV-1318: a stock decrement is a content edit, so it bumps the user-facing version
-      dbSubSample.increaseVersion();
-      registerSubSampleModification(user, dbSubSample);
-      dbSubSample = subSampleDao.save(dbSubSample);
+      // RSDEV-1318: a stock decrement is a content edit, so it bumps the user-facing version;
+      // a usage that leaves the stored quantity unchanged (e.g. against empty stock) is a no-op
+      boolean quantityChanged =
+          orgQuantity.getNumericValue().compareTo(newQuantity.getNumericValue()) != 0
+              || !orgQuantity.getUnitId().equals(newQuantity.getUnitId());
+      if (quantityChanged) {
+        dbSubSample.setQuantity(newQuantity);
+        dbSubSample.increaseVersion();
+        registerSubSampleModification(user, dbSubSample);
+        dbSubSample = subSampleDao.save(dbSubSample);
+      }
 
     } finally {
       if (temporaryLock) {
--- a/src/main/java/com/researchspace/service/inventory/impl/SubSampleApiManagerImpl.java
+++ b/src/main/java/com/researchspace/service/inventory/impl/SubSampleApiManagerImpl.java
@@ -264,7 +264,8 @@ public class SubSampleApiManagerImpl extends InventoryApiManagerImpl<SubSample>
       Long subsampleId, QuantityInfo usedQuantity, User user) {
 
     SubSample dbSubSample = getIfExists(subsampleId);
-    if (usedQuantity.getNumericValue().signum() == 0) {
+    invPermissions.assertUserCanEditInventoryRecord(dbSubSample, user);
+    if (usedQuantity == null || usedQuantity.getNumericValue().signum() == 0) {
       return getPopulatedApiSubSampleFull(dbSubSample, user);
     }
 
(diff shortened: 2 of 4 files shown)


src/main/java/com/researchspace/service/inventory/impl/SubSampleApiManagerImpl.java
```diff
@@ -281,6 +281,8 @@ public ApiSubSample registerApiSubSampleUsage(
       }
 
       dbSubSample.setQuantity(newQuantity);
+      // RSDEV-1318: a stock decrement is a content edit, so it bumps the user-facing version
+      dbSubSample.increaseVersion();
       registerSubSampleModification(user, dbSubSample);
       dbSubSample = subSampleDao.save(dbSubSample);
```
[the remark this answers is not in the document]
nhanlon2: Fixed in 3e104efc1. The zero check now uses signum() (scale-insensitive), though in practice QuantityInfo normalizes the value, a test with BigDecimal("0.00") passed even before the change. The version bump and modification are now skipped entirely when the clamped quantity equals the stored one (e.g. usage against already-empty stock), covered by new unit-test assertions.

src/test/java/com/researchspace/service/inventory/SubSampleApiManagerTest.java
```diff
@@ -584,6 +593,8 @@ public void saveUsagesOfSubSample() {
 
     retrievedSubSample = subSampleApiMgr.getApiSubSampleById(subSampleId, testUser);
     assertEquals("3.444 g", retrievedSubSample.getQuantity().toQuantityInfo().toPlainString());
+    // each non-zero usage decrement is a content edit and bumps the version (RSDEV-1318)
+    assertEquals(2L, retrievedSubSample.getVersion());
```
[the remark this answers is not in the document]
nhanlon2: Fixed in 3e104efc1: the test now asserts the version after every decrement, including the clamp-to-zero case, plus a no-bump assertion for usage against already-empty stock and a scale-zero (0.00) no-op case.

src/test/java/com/researchspace/api/v1/controller/ListOfMaterialsApiControllerMVCIT.java
```diff
@@ -159,6 +160,71 @@ public void createEditDeleteListOfMaterialsForDocField() throws Exception {
...
+    // create a list of materials, then register a 1 g usage that reduces the inventory stock
+    String newListJson = "{ \"name\": \"my list\", \"elnFieldId\": " + myField.getId() + " } ";
+    MvcResult result =
+        this.mockMvc
+            .perform(
+                createBuilderForPostWithJSONBody(apiKey, "/listOfMaterials", anyUser, newListJson))
+            .andExpect(status().isCreated())
+            .andReturn();
+    ApiListOfMaterials createdList =
+        mvcUtils.getFromJsonResponseBody(result, ApiListOfMaterials.class);
```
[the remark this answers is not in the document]
nhanlon2: Fixed in 3e104efc1: added assertNull(result.getResolvedException()) after the POST.

src/test/java/com/researchspace/api/v1/controller/ListOfMaterialsApiControllerMVCIT.java
```diff
@@ -159,6 +160,71 @@ public void createEditDeleteListOfMaterialsForDocField() throws Exception {
...
+    assertNull(result.getResolvedException());
+
+    // the subsample's stock is reduced AND its user-facing version is bumped
+    result =
+        this.mockMvc
+            .perform(
+                createBuilderForGet(API_VERSION.ONE, apiKey, "/subSamples/" + subSampleId, anyUser))
+            .andExpect(status().isOk())
+            .andReturn();
+    ApiSubSample reloaded = mvcUtils.getFromJsonResponseBody(result, ApiSubSample.class);
```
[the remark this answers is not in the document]
nhanlon2: Fixed in 3e104efc1: added assertNull(result.getResolvedException()) after both GETs.

src/main/java/com/researchspace/service/inventory/impl/SubSampleApiManagerImpl.java
```diff
@@ -280,9 +280,17 @@ public ApiSubSample registerApiSubSampleUsage(
...
       }
 
-      dbSubSample.setQuantity(newQuantity);
-      registerSubSampleModification(user, dbSubSample);
-      dbSubSample = subSampleDao.save(dbSubSample);
+      // RSDEV-1318: a stock decrement is a content edit, so it bumps the user-facing version;
+      // a usage that leaves the stored quantity unchanged (e.g. against empty stock) is a no-op
+      boolean quantityChanged =
+          orgQuantity.getNumericValue().compareTo(newQuantity.getNumericValue()) != 0
+              || !orgQuantity.getUnitId().equals(newQuantity.getUnitId());
```
rs-nicof, line 289: **The unit half of this check fires when nothing was actually used.**

`qUtils.sum` returns its result in the larger of the two units, and the clamp above only resets the number, not the unit. So a subsample already at `0 g`, used with `45 mg`, comes back as `0 mg`: same amount, different unit, which counts as changed here.

Result: a version bump and a revision-history entry for a usage that consumed nothing, plus the stored amount silently relabelling itself from `0 g` to `0 mg`. That contradicts the "already-empty stock is a complete no-op" promise in the new Javadoc.

Simplest fix is to clamp in the original unit on line 280, so the zero case compares like for like:

```java
newQuantity = QuantityInfo.of(BigDecimal.ZERO, orgQuantity.getUnitId());
```

Do keep the unit comparison itself, it is load-bearing: `1 g` minus `0.999 g` normalises to `1 mg`, which compares numerically equal to the stored `1`.

src/main/java/com/researchspace/service/inventory/impl/SubSampleApiManagerImpl.java
```diff
@@ -264,7 +264,7 @@ public ApiSubSample registerApiSubSampleUsage(
       Long subsampleId, QuantityInfo usedQuantity, User user) {
 
     SubSample dbSubSample = getIfExists(subsampleId);
-    if (usedQuantity.getNumericValue().equals(BigDecimal.ZERO)) {
+    if (usedQuantity.getNumericValue().signum() == 0) {
```
rs-nicof: **`usedQuantity` can be null here, giving a 500 instead of a 400.**

`ApiMaterialUsage.usedQuantity` has no `@NotNull`, and `toMaterialUsage` leaves it null, so a `POST /listOfMaterials` with `"updateInventoryQuantity": true` and no `usedQuantity` reaches this line as null.

Pre-existing, but you are already rewriting this line, so it is a one-token fix:

```java
if (usedQuantity == null || usedQuantity.getNumericValue().signum() == 0) {
```

src/main/java/com/researchspace/service/inventory/impl/SubSampleApiManagerImpl.java
```diff
@@ -280,9 +280,17 @@ public ApiSubSample registerApiSubSampleUsage(
...
-      registerSubSampleModification(user, dbSubSample);
-      dbSubSample = subSampleDao.save(dbSubSample);
+      // RSDEV-1318: a stock decrement is a content edit, so it bumps the user-facing version;
+      // a usage that leaves the stored quantity unchanged (e.g. against empty stock) is a no-op
+      boolean quantityChanged =
+          orgQuantity.getNumericValue().compareTo(newQuantity.getNumericValue()) != 0
+              || !orgQuantity.getUnitId().equals(newQuantity.getUnitId());
+      if (quantityChanged) {
+        dbSubSample.setQuantity(newQuantity);
+        dbSubSample.increaseVersion();
```
rs-nicof, line 292: **Worth checking: the version can now outrun the revision history.**

Envers writes one revision per entity per transaction, but this line can bump twice within one. If a single request decrements the same subsample twice (the stoichiometry path at `StoichiometryInventoryLinkManagerImpl:152` runs once per link), the version advances by 2 while only one revision is written.

The skipped version then has no revision at all, so `findNewestRevisionForVersion` returns null: `GET /subSamples/{id}/version/{skipped}` finds nothing, and an ELN link pinned to that version quietly falls back to live data instead of the snapshot it was pinned to.

src/test/java/com/researchspace/api/v1/controller/ListOfMaterialsApiControllerMVCIT.java
```diff
@@ -159,6 +160,74 @@ public void createEditDeleteListOfMaterialsForDocField() throws Exception {
...
+            .perform(
+                createBuilderForGet(
+                    API_VERSION.ONE, apiKey, "/subSamples/" + subSampleId + "/revisions", anyUser))
+            .andExpect(status().isOk())
+            .andReturn();
+    assertNull(result.getResolvedException());
+    ApiInventoryRecordRevisionList history =
+        mvcUtils.getFromJsonResponseBody(result, ApiInventoryRecordRevisionList.class);
+    assertEquals(
+        2, history.getRevisions().size(), "a stock decrement must add a revision-history entry");
```
rs-nicof, line 228: **This assertion passes without the fix, so it is not guarding the change.**

The old code already called `setQuantity` and saved, so Envers was already writing a second revision. The decrement was invisible only because the UI groups revisions that share a version (`groupByVersion` in `versionHistory.ts`).

The version check on line 214 is what actually pins this. Asserting the two revisions carry versions 1 and 2 would make this assertion meaningful too.

src/main/java/com/researchspace/service/inventory/impl/SubSampleApiManagerImpl.java
```diff
@@ -264,7 +264,7 @@ public ApiSubSample registerApiSubSampleUsage(
       Long subsampleId, QuantityInfo usedQuantity, User user) {
 
     SubSample dbSubSample = getIfExists(subsampleId);
```
rs-nicof, line 266: **Pre-existing, but this PR raises the stakes: no edit-permission check on this write path.**

There is no `assertUserCanEditInventoryRecord` here, and the two callers disagree. The stoichiometry path asserts edit rights, but the list-of-materials path only checks read/limited-read (`ListOfMaterialsApiManagerImpl:146` and `:176`), and on the update path the decrement runs before that check.

So someone who can only *see* another user's subsample appears able to decrement it, and after this PR that also bumps the owner's version and leaves a permanent entry in their revision history. One guard here would cover both callers.

rs-nicof (commented):
Automated parallel review (5 reviewers: layering, security, persistence, i18n, generic quality). Reporting Critical/High only; Copilot's earlier points look correctly addressed in `3e104efc1`, so they are not repeated.

The unit-comparison issue below was independently flagged by four of the five reviewers and is the one I would fix before merge.

src/main/java/com/researchspace/service/inventory/impl/SubSampleApiManagerImpl.java
nhanlon2: This response is from Claude: Fixed in 9771ea094. The clamp now rebuilds the zero in the stored unit (new QuantityInfo(ZERO, orgQuantity.getUnitId())), and the unit comparison stays. Unit test covers 45 mg against 0 g: still 0 g, no bump.

src/main/java/com/researchspace/service/inventory/impl/SubSampleApiManagerImpl.java
nhanlon2: This response is from Claude: Fixed in 9771ea094: null usedQuantity takes the no-op path, pinned by a unit test.

src/test/java/com/researchspace/api/v1/controller/ListOfMaterialsApiControllerMVCIT.java
nhanlon2: This response is from Claude: Fixed in 9771ea094: the test now fetches both revisions and asserts they carry versions 1 and 2, plus quantities 5 g and 4 g.

src/main/java/com/researchspace/service/inventory/impl/SubSampleApiManagerImpl.java
nhanlon2: This response is from Claude: Fixed in 9771ea094: registerApiSubSampleUsage asserts edit permission up front, covering both callers. New unit test: a limited-read user's decrement throws and leaves the subsample untouched.

src/main/java/com/researchspace/service/inventory/impl/SubSampleApiManagerImpl.java
nhanlon2: This response is from Claude: Confirmed real, but it only occurs when one transaction decrements the same subsample twice, and the clean fix (per-transaction bump dedup) is bigger than this PR. Deferred; suggest a follow-up ticket.

src/main/java/com/researchspace/service/inventory/SubSampleApiManager.java
```diff
@@ -68,6 +68,14 @@ List<ApiSubSample> createNewSubSamplesForSample(
    */
   ApiSubSample addNewApiSubSampleToSample(ApiSubSample incomingSubSample, Long sampleId, User user);
 
+  /**
+   * Reduces the subsample's quantity by the given used amount (unit-aware, clamped at zero). A
+   * usage that changes the stored quantity is a content edit: it bumps the subsample's user-facing
+   * version and is recorded in its revision history (RSDEV-1318). A usage that leaves the quantity
+   * unchanged (zero usage, or usage against already-empty stock) is a complete no-op.
```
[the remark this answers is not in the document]
nhanlon2: Fixed in 81f3d72dc: Javadoc now states the edit-permission requirement and promises only that the stored subsample is unchanged, rather than claiming a complete no-op.

rs-nicof (approved):

PR #1041, opened 2026-08-21
RSDEV-1147: Merge core-model, core-util, audit and test-util back into rspace-web

rs-fraser:
## Description ##

Merges the four core library repositories back into rspace-web: rspace-core-util (at 2.0.0), rspace-audit (1.0.0), rspace-core-model (3.2.1 initially, caught up through its 3.3.0 and 3.3.1 releases as they landed), and rspace-test-util (3.0.0).

**Reviewing this PR**: ~99% of the diff is verbatim file adds from the four merges. The commits worth human and bot attention are the delta commits:

- `ade1eeccf` Initialise Fieldmark DTO collections at declaration (real NPE bugs surfaced by an imported hygiene test)
- `b1c1088bd` Apply Spotless to the imported sources (formatting only, 761 files)
- `8a716e7d7` Update docs for the absorbed core repositories
- `6213f3bcb` Guard the imported audit test config behind a profile
- `dedd17466` Widen the harness timeout on the invalid-URL retriever test
- `f66bdc2b4` Skip the hibernate model tests when no database is listening
- `1f99b4458` Run the hibernate model tests on the spring CI shards (adds the hibtest scratch DB to the workflow's MariaDB setup)
- `c570c7725` Let hibernate model tests create their own hibtest database
- `8a9e15cf5` Review response: drop the ArchUnit fence this PR originally added, the imported changelogs, and the stale Maven module diagram

plus the pom changes inside each merge commit (dependency removals, exclusions, and the zxing / commons-compress / indriya additions that were previously resolved only through the absorbed jars).

**Landing**: squash merge. We decided not to carry the incoming repositories' git history into rspace-web, so the imported commits are flattened into one commit on main; the original history stays readable in the four repositories, which are archived rather than deleted. Worth writing the squash commit message by hand, since the default concatenates every commit on the branch. #1045 and #1059 are stacked on this branch, so they need rebasing onto main once this lands.

## Scope
This PR copied the code as-is from the libraries. #1045 removes the dead code copied across now that it's visibly unused due to being in the same project. #1059 addresses some review comments from the copied code.

## Design decisions
- Flatten into the existing single module rather than a multi-module build. An ArchUnit fence re-creating the repo split's compile-time layering was added and then removed at review request, as the rule cost more upkeep than it was judged to be worth.
[...]

PR #1044, opened 2026-08-21
RSDEV-1147: Harden the imported measurement-input validators

rs-fraser:
## Description ##

Stacked on #1041. Fixes the four real defects Copilot's review of the core-repository merge found in the imported validation helpers, all in `com.researchspace.model.units`:

- `AmountConstraintValidator`, `ValidA260InputValidator`, and `ValidCellDoublingInputValidator` threw `NullPointerException` when class-level validation ran against null values or null fields; they now treat null as valid and leave requiredness to the field-level `@NotNull` constraints, per the usual Bean Validation contract.
- `HalfLifeInput`'s time-unit patterns had a trailing `|` that made the empty string validate, and `@Min(0)` accepted 0 while the message promised "> 0"; the patterns are fixed and the constraint is now an exclusive `DecimalMin`.

Severity context: none of these classes currently has a live caller in the tree, so the bugs were unreachable in production; this is hardening of latent helpers rather than a behaviour change.

Copilot's remaining three comments on #1041 (raw `Set` generics in `PermissionStringConverter`, `ZipEntryProcessor` javadoc and unused import, JUnit 4 `@Rule` in `AllowanceTrackerSourceImplTest`) are cosmetic and deliberately not included here.

## Testing notes

New `InputValidatorNullSafetyTest` covers the null paths and the surviving rejection cases; `HalfLifeInputTest` gains cases for the empty-string unit and zero times.

### DIFF
--- a/src/main/java/com/researchspace/model/units/AmountConstraintValidator.java
+++ b/src/main/java/com/researchspace/model/units/AmountConstraintValidator.java
@@ -11,6 +11,9 @@ public class AmountConstraintValidator implements ConstraintValidator<ValidAmoun
 
   @Override
   public boolean isValid(Quantifiable value, ConstraintValidatorContext context) {
+    if (value == null) {
+      return true;
+    }
     return AmountValidator.validate(value);
   }
 }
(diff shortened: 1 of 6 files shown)


rs-fraser:
Superseded: #1045 deletes every class this PR hardened (they were dead code), so the stack collapses to #1041 + #1045. The four validator findings from Copilot's review of #1041 are resolved by deletion.

PR #1046, opened 2026-08-21
RSDEV-779-omero-bug

nhanlon2:
## Problem

The OMERO Apps card never showed "connected" and had no Disconnect button, and
the connection died on logout. Three causes:

1. Credentials lived in a `@SessionScope` `HashMap` bean, not the `UserConnection`
   table, so they evaporated with the session.
2. `IntegrationsHandlerImpl.postProcessInfo` had no `OMERO` case, so
   `/integration/integrationInfo` never reported the connection.
3. `Omero.tsx` rendered the credentials form unconditionally — no disconnect branch existed.

The stale comment in `OmeroController` ("we save omero credentials ... in the
access token field of UserConnection **table**") shows it originally used the DB.
Broken since the initial squash commit.

## Fix

OMERO now behaves like every other credential-based app: the username/password
go into `UserConnection.accessToken` as a `username_,_password` string, encrypted
at rest by the DAO, reported to the Apps page as `MASKED_TOKEN` only.

- `connect` persists through one `UserConnectionManager.replaceConnection` call
  (new): delete-then-save in **one transaction**, so a failed save cannot leave
  the user with no connection at all.
- `disconnect` deletes the row; the card gets a Disconnect button.
- `OmeroController`'s eight copies of the same "look up the connection or throw"
  block collapsed into one helper, which also removed seven hard-coded English
  error strings.
- The `userNameToUserConnection` session bean is deleted.
- `useClustermarket.tsx` generalised into a shared `useDisconnect.ts`.

No Liquibase change: `UserConnection` already exists and the session map held
nothing durable. Users reconnect once — which they had to do every session anyway.

See `DevDocs/adr/0006-persist-omero-credentials-in-userconnection.md`.

## Found by review, fixed here

Persisting the credentials raised the stakes on some pre-existing problems:

- **`POST /apps/omero/connect` logged the OMERO password in cleartext.**
  `LoggingInterceptor` is mapped to `/**` and logs every request param;
  `/apps/**` is not excluded. Added
  `@IgnoreInLoggingInterceptor(ignoreRequestParams = {"omeropassword"})`, with a
  reflection test so it cannot silently regress.
- **`JSONClient.login` returns `null` instead of throwing** when OMERO rejects
  the credentials, and `connect` ignored the result — so a refused password was
  stored and then replayed against the user's real OMERO account on every later
  call. Now checked.
- **A failed disconnect flipped the UI to "disconnected"** while the row
  survived. The shared hook swallowed the error and resolved; it now resolves
  `false` and both cards gate on it.
- **The typed password stayed in component state** after connecting, ready to be
  resubmitted. Cleared on success.
- A blank password stored `"user_,_"`, which splits back into one element and
  breaks every later OMERO request. `connect` now rejects blank halves and the
  `_,_` delimiter, and reports a misconfigured `omero.servername` instead of
  "Could not log in to OMERO: null".
- `OmeroUser` NPE'd on a form submission that omitted a field; passwords are now
  `@ToString.Exclude`.
- The TinyMCE OMERO dialog still said "Your session with Omero has expired",
  which is now false. Reworded to "not connected", matching the renamed
  `apps.omero.errors.notConnected` server key.

## Tests

`mvn clean test -Dfast=true`: **2690 pass**. Vitest: **2446 pass**. `tsc` and
`biome` clean.

New: `OmeroAuthControllerTest`, `OmeroControllerTest`, `replaceConnection`
ordering in `UserConnectionManagerTest`, two OMERO cases in
`IntegrationsHandlerTest`, six cases in `Omero.test.tsx`, `Clustermarket.test.tsx`
(the shared hook's other call site), and `decodeOmero` coverage.

No MVCIT added — every gap was reachable from a pure unit test.

One existing test was passing for the wrong reason: `getOmeroConnectionStatus`
left `appCfgMgr` unstubbed, so the masking assertion never exercised the real
option-map ordering. Now stubbed.

## Deliberately not done

- Moving `OmeroAccessTokenReader` / `OmeroUser` / `JSONClient` down out of
  `webapp.integrations` — `OmeroClientImpl` (client layer) importing from a
  controller is a real pre-existing layering inversion, but the fix is a package
  move with a wide blast radius. Worth its own ticket.
- Migrating the other seven `useX` disconnect hooks onto `useDisconnect.ts`.
  They are byte-for-byte equivalent; follow-up ticket.
- An OMERO session key instead of the password. Smaller at-rest liability, but
  every call path currently authenticates with username and password. Recorded as
  a rejected option in the ADR.
- Re-showing the stored username when connected. No other app re-displays
  credentials, and the password must never go back to the browser.


src/main/webapp/ui/src/eln/apps/integrations/Clustermarket.tsx
```diff
@@ -44,12 +44,12 @@ export const CLUSTERMARKET_CONNECTION_CHANNEL = "rspace.apps.clustermarket.conne
  * success alerts, etc.).
  *
  * The process of disconnecing is via a standard API call made by
```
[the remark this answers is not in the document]
nhanlon2: Fixed in 7476acbfa. Note the same typo exists in five sibling cards (Dryad, Figshare, NextCloud, OwnCloud, ProtocolsIO) at the same line, inherited from a copy-pasted block comment. Left those alone to keep this PR scoped to the OMERO fix.

src/main/java/com/researchspace/webapp/integrations/omero/OmeroAuthController.java
```diff
@@ -54,18 +51,51 @@ public static OmeroUser createOmeroUserCredentialsFromDelimitedString(String tar
...
+   * The credentials are stored as a delimited string, so neither half may contain the delimiter and
+   * neither may be blank: a blank password would store "user_,_", which splits back into a single
+   * element and breaks every later OMERO request.
+   */
+  private void validateCredentials(OmeroUser loginData) {
+    if (StringUtils.isBlank(loginData.getOmerousername())
+        || StringUtils.isBlank(loginData.getOmeropassword())) {
+      throw new IllegalArgumentException(getText("apps.omero.errors.blankCredentials"));
+    }
+    if (loginData.getOmerousername().contains(OmeroAccessTokenReader.credentialsDelimiter)
```
rs-nicof: The per-half `contains` check misses a delimiter formed across the join boundary: username `user_,` plus password `pass` stores `user_,_,_pass`, which splits back as `user` / `,_pass`. Connect succeeds (it validates the raw values against OMERO before joining), but every later OMERO call replays wrong credentials with no error explaining why. Cheapest guard: after building the delimited string, reject it unless it splits back into exactly the two original values.

src/test/java/com/researchspace/service/UserConnectionManagerTest.java
```diff
@@ -85,4 +87,19 @@ public void testDeleteByUserNameProviderName() {
     assertEquals(result, userConnMgr.deleteByUserAndProvider(USERNAME, PROVIDER_NAME));
     verify(connectionDao).deleteByUserAndProvider(USERNAME, PROVIDER_NAME);
   }
+
+  @Test
+  public void replaceConnectionDeletesAnyExistingConnectionBeforeSaving() {
```
rs-nicof, line 92: This InOrder check can't protect the two things `replaceConnection` actually relies on: the AOP transaction wrapper and the bulk delete executing before Hibernate's flush. If a refactor swapped the bulk delete for `session.delete(entity)`, flush ordering would invert, same-username reconnects would fail on the unique index, and this test would stay green. One case in the existing `UserConnectionIT` (save, replace under the same and a different OMERO username, assert one decryptable row) would pin it for real.

DevDocs/adr/0006-persist-omero-credentials-in-userconnection.md
```diff
@@ -0,0 +1,81 @@
...
+* `apps.omero.errors.authenticationExpired` became
+  `apps.omero.errors.notConnected`: with durable storage a missing row means
+  "never connected or disconnected", not "expired".
+* The `userNameToUserConnection` session bean is deleted; the two OMERO
+  controllers were its only users.
+* `OmeroAuthController.newJsonClient()` exists purely as an overridable seam so
+  `connect` can be unit-tested without a live OMERO server.
+* Unlike a revocable OAuth token, the stored value is a reusable third-party
+  account password. It is encrypted at rest, but recoverable from a database
+  dump plus the single `apitoken.encryption.key` deployment property, and there
```
rs-nicof: Worth adding here (or as a startup warning): `apitoken.encryption.key` has a committed default in `defaultDeployment.properties`, so on any deployment that never overrode it this encryption is only nominal. Since this PR is the first to store a reusable third-party password (rather than a revocable token) under that key, the 'set your own key' requirement deserves an explicit callout.

src/main/webapp/ui/src/eln/apps/integrations/__tests__/Omero.test.tsx
```diff
@@ -76,4 +80,154 @@ describe("Omero", () => {
...
+        }}
+        update={() => {}}
+      />,
+    );
+
+    await user.click(screen.getByRole("button"));
+    await user.click(screen.getByRole("button", { name: "apps:actions.disconnect" }));
+
+    // the UserConnection row still exists, so the card must not claim to be disconnected
+    expect(await screen.findByRole("button", { name: "apps:actions.disconnect" })).toBeVisible();
```
rs-nicof: Nothing waits for the failed DELETE to be processed. The Disconnect button is present from the start, so this assertion passes immediately even if the `if (await disconnect())` guard is removed, which is the exact regression this test exists to catch. Waiting for the error alert first (an `AlertContext.Provider` with a `vi.fn()` `addAlert` plus `waitFor`, as the test below does) would make it a real safety net.

src/main/webapp/ui/src/eln/apps/integrations/__tests__/Clustermarket.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+          credentials: { ACCESS_TOKEN: Optional.present("MASKED") },
+        }}
+        update={() => {}}
+      />,
+    );
+
+    await user.click(screen.getByRole("button"));
+    await user.click(screen.getByRole("button", { name: "apps:actions.disconnect" }));
+
+    expect(await screen.findByRole("button", { name: "apps:actions.disconnect" })).toBeVisible();
```
rs-nicof: Same race as the Omero copy of this test: the assertion can pass before the 500 is handled, so it can't catch removal of the `if (await disconnect())` guard. Wait for the error alert before asserting the Disconnect button is still shown.

rs-nicof (commented):
Ran a six-lens parallel review (layering, i18n, security, persistence, frontend, general quality) against this branch. Five findings worth a look: one credential-encoding edge case, one at-rest-encryption caveat for the ADR, and three test-strength gaps. Everything else came back clean, including the transaction, cache-eviction and masking analysis.

DevDocs/adr/0006-persist-omero-credentials-in-userconnection.md
```diff
@@ -0,0 +1,81 @@
+# OMERO credentials live in UserConnection, not the session
```
rs-nicof: minor: ADR number 0006 is already taken by `0006-registered-landing-page-and-public-link-suffix.md` (0007 is free), so citations like 'see ADR 0006' become ambiguous. Worth a one-line rename before merge.

src/main/java/com/researchspace/webapp/integrations/omero/OmeroAuthController.java
```diff
@@ -54,18 +51,51 @@ public static OmeroUser createOmeroUserCredentialsFromDelimitedString(String tar
...
+        || loginData.getOmeropassword().contains(OmeroAccessTokenReader.credentialsDelimiter)) {
+      throw new IllegalArgumentException(
+          getText(
+              "apps.omero.errors.credentialsDelimiter",
+              new Object[] {OmeroAccessTokenReader.credentialsDelimiter}));
+    }
+  }
 
   @PostMapping("/connect")
+  @IgnoreInLoggingInterceptor(ignoreRequestParams = {"omeropassword"})
```
rs-nicof: minor: `OmeroUser` also has a writable `webClientPassword` property, which Spring will bind from a request param and the interceptor will log, since only `omeropassword` is excluded. The UI never sends it, but adding it to `ignoreRequestParams` is a one-token hardening.

src/main/java/com/researchspace/webapp/integrations/omero/OmeroAuthController.java
nhanlon2: This response is from Claude: Fixed in b59710e8a. Round-trip guard added: connect rejects credentials unless the delimited string splits back into the two original values. Covered by connectRejectsCredentialsFormingTheDelimiterAcrossTheJoinBoundary.

src/test/java/com/researchspace/service/UserConnectionManagerTest.java
nhanlon2: This response is from Claude: Added replaceConnectionLeavesExactlyOneDecryptableRow to UserConnectionIT in b59710e8a (save, replace under same and different provider user id, assert one decryptable row). Compiled but not executed locally; it needs the drop-recreate DB harness, so CI will run it.

DevDocs/adr/0006-persist-omero-credentials-in-userconnection.md
nhanlon2: This response is from Claude: The ADR was removed from this PR at the repo owner's request (394afd862), so there is no place for the callout here. The default-key concern stands; a startup warning would be a separate change.

src/main/webapp/ui/src/eln/apps/integrations/__tests__/Omero.test.tsx
nhanlon2: This response is from Claude: Fixed in b59710e8a. The test now waits for the error alert (AlertContext + waitFor) before asserting the card is still connected.

src/main/webapp/ui/src/eln/apps/integrations/__tests__/Clustermarket.test.tsx
nhanlon2: This response is from Claude: Fixed in b59710e8a, same pattern as the Omero test: wait for the error alert before asserting.

DevDocs/adr/0006-persist-omero-credentials-in-userconnection.md
nhanlon2: This response is from Claude: Moot. The ADR file was removed from this PR at the repo owner's request (394afd862).

src/main/java/com/researchspace/webapp/integrations/omero/OmeroAuthController.java
nhanlon2: This response is from Claude: Fixed in b59710e8a. webClientPassword added to ignoreRequestParams, with an assertion in connectExcludesTheOmeroPasswordFromTheRequestLog.

rs-nicof (approved):

src/main/webapp/ui/src/eln/apps/integrations/Omero.tsx
```diff
@@ -24,13 +25,17 @@ export interface OmeroConnectedMessage extends Record<string, unknown> {
...
+ * backend stores them in the UserConnection table, so the connection outlives
+ * the session and this card shows a Disconnect button once connected.
  */
 function Omero({ integrationState, update }: OmeroArgs): React.ReactNode {
   const { t } = useTranslation("apps");
   const { addAlert } = useContext(AlertContext);
+  const { disconnect } = useDisconnectEndpoint("/apps/omero", t("integrations.omero.name"));
   const [username, setUsername] = useState("");
   const [password, setPassword] = useState("");
+  const [connected, setConnected] = useState(integrationState.credentials.ACCESS_TOKEN.isPresent());
```
tchoi-rs, line 38: nit: It seems wrong that we pass the access token back to the user after it has been saved, but I won't chase it down further

tchoi-rs (approved):

PR #1049, opened 2026-08-21
RSDEV-1288 CSV Import for Instruments

mrowe-tan:
## Description ##
- Resolves https://researchspace.atlassian.net/browse/RSDEV-1288

## Summary of changes by Claude ##

Store layer

  ImportStore.ts — Added "INSTRUMENTS" to the ImportRecordType union;
  ?recordType=INSTRUMENTS in the URL now initialises the store correctly.

  ImportModel.ts — Added all instrument-specific state (instrumentsFile,
  instrumentsMappings, instrumentCreateNewTemplate, instrumentTemplateName,
  instrumentTemplate) with corresponding actions and computed properties
  (isInstrumentsImport, instrumentsSubmittable, validInstrumentTemplateName,
  importInstrumentMatchesExistingTemplate). Wired instruments into every
  existing switch-style method: setFile, resetMappingsByRecordType,
  resetAllLoadedFiles, fileByRecordType, mappingsByRecordType,
  someFileSubmitted, and importSubmittable.

  ---
  UI — new files

  Import/Fields/InstrumentTemplateDetails.tsx — Template-selection panel for
  instruments. Offers a radio choice between creating a new template (name
  field) and choosing an existing one (picker, summary info, mismatch alert).
  Uses a stable useCallback-wrapped handleSetTemplate to avoid a spurious
  re-selection bug.

  InstrumentTemplate/SummaryInfo.tsx — Displays the selected instrument
  template's name, thumbnail, and global ID above the picker, parallel to the
  existing sample-template SummaryInfo.

  ---
  UI — modified files

  RecordsImport.tsx — Added INSTRUMENTS to the tab list and conditionally
  renders <InstrumentTemplateDetails />.

  ImportRouter.tsx — Permits INSTRUMENTS as a valid record type for the wizard.
  
  ColumnFieldMapping.tsx — Added an instrument-template mismatch alert driven by
  importInstrumentMatchesExistingTemplate.

  CreateNew.tsx — Consolidated the per-type CSV menu entries into a single "From
  CSV" option covering all four types; extended handleImport to accept
  "INSTRUMENTS".

  TemplateName.tsx — Decoupled from the sample-template store; now accepts
  value/onChange/error as props so it can be reused by both the sample and
  instrument template flows.

  TemplateDetails.tsx — Stabilised the setTemplate prop with useCallback to fix
  a bug where selecting a template from the picker immediately re-triggered
  selection via a stale callback chain.

  ---
  Assets and i18n

  Icons/Template.tsx — Fixed the SVG viewBox from "0 0 16 16" to "-1 -1 16 16"
  to centre the template icon path, which only occupies the top-left 14×14 units
  of the canvas.

  inventory.json — Added fromCsv/fromCsvDescription keys for the consolidated
  menu entry; added selectInstrumentTemplate under import.templateDetails.


src/main/webapp/ui/src/stores/models/ImportModel.ts
```diff
@@ -826,6 +968,37 @@ export default class Import {
...
+  transformInstrumentTemplateInfoForSubmission(): { fields: Array<unknown>; name: string } {
+    if (!this.instrumentTemplateInfo) throw new Error("InstrumentTemplateInfo is null");
+    const templateFieldWithMappings: Array<{
+      field: { name: string; type: string; [key: string]: unknown };
+      mapping: ColumnFieldMap;
+    }> = zipWith(
+      this.instrumentTemplateInfo.fields,
+      this.instrumentTemplateInfo.fields
+        .map(({ name }) => this.instrumentsMappings.find((f) => f.fieldName === name) ?? null)
+        .filter(isNotNil),
```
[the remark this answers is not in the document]
mrowe-tan: This does not appear to be an actual issue - it is not possible to edit field names during import in the UI.

src/main/webapp/ui/src/Inventory/Import/Fields/InstrumentTemplateDetails.tsx
```diff
@@ -0,0 +1,79 @@
...
+        <FormControl component="fieldset" fullWidth>
+          <FormGroup sx={{ maxWidth: 660 }}>
+            <TextField
+              variant="standard"
+              label={t("import.fields.templateName")}
+              fullWidth
+              disabled={!createNewTemplate}
+              value={templateName}
+              onChange={({ target }) => importData?.setInstrumentTemplateName(target.value)}
+            />
```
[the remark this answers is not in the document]
mrowe-tan: This discrepancy with Samples has been resolved by updating the component that Samples uses to be shareable and usable by both Samples and Instruments.

src/main/webapp/ui/src/Inventory/Import/Fields/ColumnFieldMapping.tsx
```diff
@@ -79,6 +79,7 @@ function ColumnFieldMapping({ onTypeSelect }: MappingArgs): React.ReactNode {
   const hasRows = rowCount > 0;
 
   const matchExistingTemplate = importData.importMatchesExistingTemplate;
+  const matchExistingInstrumentTemplate = importData.importInstrumentMatchesExistingTemplate;
```
nhanlon2, line 82: This comment is from Claude. This file learned about instruments here, but the mapping table below did not. Line 203 still passes `existingTemplate={!importData.createNewTemplate}` for every record type, which is the **sample** flag.

On the Instruments tab the custom-field name/type editors are therefore shown or hidden based on what the user picked on the Samples tab: pick an existing sample template and the instrument editors vanish while creating a new instrument template; pick an existing instrument template and they stay editable when they should not.

Suggest `existingTemplate={importData.isInstrumentsImport ? !importData.instrumentCreateNewTemplate : !importData.createNewTemplate}`.

src/main/webapp/ui/src/stores/models/ImportModel.ts
```diff
@@ -826,6 +968,37 @@ export default class Import {
...
 
+  transformInstrumentTemplateInfoForSubmission(): { fields: Array<unknown>; name: string } {
+    if (!this.instrumentTemplateInfo) throw new Error("InstrumentTemplateInfo is null");
+    const templateFieldWithMappings: Array<{
+      field: { name: string; type: string; [key: string]: unknown };
+      mapping: ColumnFieldMap;
+    }> = zipWith(
+      this.instrumentTemplateInfo.fields,
+      this.instrumentTemplateInfo.fields
+        .map(({ name }) => this.instrumentsMappings.find((f) => f.fieldName === name) ?? null)
```
nhanlon2, line 978: This comment is from Claude. These two arrays can be different lengths. The second is the field list mapped to mappings and then `.filter(isNotNil)`, so a single field whose `fieldName` has no match shortens it. `zipWith` pads to the longer array, so every field after the miss is paired with the wrong mapping and the last pair gets `undefined`, which then throws on the `({ mapping: { selected } })` destructure at line 984.

A user can trigger this just by renaming a custom field in the mapping table. They get a generic import failure instead of a validation message.

Building the pairs in one pass avoids it:

```ts
this.instrumentTemplateInfo.fields
  .map((field) => ({ field, mapping: this.instrumentsMappings.find((m) => m.fieldName === field.name) }))
  .filter(({ mapping }) => isNotNil(mapping))
```

Same shape exists at line 937 for samples, so one shared fix covers both.

src/main/webapp/ui/src/stores/models/ImportModel.ts
```diff
@@ -269,6 +278,14 @@ export class ColumnFieldMap {
         "custom",
       ]),
       SUBSAMPLES: new Set(["expiry_date", "source", "import_identifier", "custom"]),
+      INSTRUMENTS: new Set([
```
nhanlon2, line 281: This comment is from Claude. This set does not exclude `custom`, so instrument columns can now become custom template fields. But `validFieldName` at line 261 still returns early with `this.recordType !== "SAMPLES"`, so it is unconditionally true for instruments.

That means the blank/reserved-name check (`Name`, `Description`, `Tags`, `Source`, `Expiry Date`) and the per-record-type uniqueness check this PR otherwise fixed never run for instruments. A user can blank a field name or call it "Description" and see no error; `validateMappings` then blocks submission, so they get a disabled Import button with nothing on screen explaining why.

Either widen the guard to include `INSTRUMENTS`, or drive it off whether `custom` is available for the record type.

src/main/webapp/ui/src/stores/models/ImportModel.ts
```diff
@@ -451,12 +519,14 @@ export default class Import {
...
     this.samplesMappings = [];
     this.subSamplesMappings = [];
+    this.instrumentsMappings = [];
   }
 
   resetAllLoadedFiles() {
     this.samplesFile = null;
     this.containersFile = null;
     this.subSamplesFile = null;
+    this.instrumentsFile = null;
```
nhanlon2, line 529: This comment is from Claude. `instrumentsFile` is cleared here, but none of the instrument template state is. `instrumentTemplateName`, `instrumentTemplateInfo`, `instrumentTemplate` and `instrumentCreateNewTemplate` all survive, and `clearFile()` at line 535 resets the sample template name only.

So after one instrument import, or after removing the chosen CSV, the name box still reads "New template from <old file name>" and the previously chosen template stays selected. The next import quietly reuses those stale settings unless the user spots it.

src/main/webapp/ui/src/stores/models/ImportModel.ts
```diff
@@ -872,36 +1045,58 @@ export default class Import {
...
-          subSampleSettings: this.subSamplesSubmittable
+          subSampleSettings: subSamplesSubmittable
             ? {
                 fieldMappings: this.makeMappingsObject(this.subSamplesMappings),
               }
             : null,
+          instrumentSettings: instrumentsSubmittable
+            ? {
+                // biome-ignore lint/style/noNonNullAssertion: instrumentTemplate is non-null when !instrumentCreateNewTemplate and instrumentsSubmittable is true
+                templateId: instrumentCreateNewTemplate ? null : instrumentTemplate!.id,
```
nhanlon2: This comment is from Claude. This branch, importing instruments against an **existing** template, has no test coverage. `instruments.test.tsx` never calls `importFiles()`, and the e2e at `import.e2e.ts:113` only exercises the create-new-template path, so the non-null assertion here is justified only by a comment.

A fast unit test would cover it: stub `ApiService.post`, set `instrumentCreateNewTemplate = false` with a mock template, call `importFiles()`, and assert the serialised `instrumentSettings` is `{ templateId: <id>, templateInfo: null, fieldMappings: ... }`. No integration infrastructure needed.

src/main/webapp/ui/src/stores/models/ImportModel.ts
```diff
@@ -1028,14 +1230,88 @@ export default class Import {
...
+        reason: i18n.t("inventory:import.columnMapping.templateRequiredValueMissing", {
+          columns: formatList(columnsWithMissingData, i18n.resolvedLanguage ?? i18n.language),
+        }),
+      };
+    }
+
+    return { matches: true };
+  }
+
+  get importInstrumentMatchesExistingTemplate(): { matches: false; reason: string } | { matches: true } | null {
```
nhanlon2, line 1305: This comment is from Claude. Three copies to flag together, since they are the same pattern:

- `importInstrumentMatchesExistingTemplate` (1299-1371) is a near line-for-line copy of `importMatchesExistingTemplate` (1227-1297); only `instrumentTemplate`/`instrumentsMappings` differ.
- The `isInstrumentsImport` parse branch (818-860) copies the samples branch (773-817), differing only in `quantityUnitId: null` and one optional chain.
- `transformInstrumentTemplateInfoForSubmission` (971) copies `transformTemplateInfoForSubmission` (923-969).

Each could take the target template and mappings as parameters instead. As it stands, any future fix to the matching rules or to CSV column interpretation has to be made twice, and samples and instruments will drift apart.

`TemplateName.tsx` in this same PR shows the alternative: it was refactored to serve both flows rather than copied.

src/main/webapp/ui/src/Inventory/InstrumentTemplate/SummaryInfo.tsx
```diff
@@ -0,0 +1,68 @@
...
+        aria-hidden
+      />
+    );
+    return <OneItemList avatar={avatar} text={template.name} action={<GlobalId record={template} />} />;
+  }
+  if (loading) {
+    return <OneItemList avatar={<FontAwesomeIcon icon={faSpinner} spin size="lg" />} text={t("common:loading")} />;
+  }
+  if (paddingless) {
+    return <NoValue label={t("template.summary.noTemplate")} />;
```
nhanlon2: This comment is from Claude. This instrument component resolves `template.summary.noTemplate`, which is the **sample** template key (same at line 63). The catalogue already has an instrument namespace, e.g. `inventory:instrumentTemplate.field.noTemplate`.

The two screens are now welded to one string, so anyone rewording it for sample templates silently changes the instrument import screen too.

src/main/webapp/ui/src/__tests__/e2e/specs/inventory/import.e2e.ts
```diff
@@ -108,4 +109,31 @@ test.describe(`Inventory CSV Import`, { tag: [tags.INVENTORY, tags.MOBILE] }, ()
...
+    pageInventory,
+  }) => {
+    await pageInventory.open();
+    await pageInventory.isLoaded();
+
+    const importPage = await pageInventory.openCsvImport("INSTRUMENTS");
+
+    await expect(importPage.main.getByRole("radio", { name: "Create new template." })).toBeVisible();
+    await expect(importPage.main.getByRole("radio", { name: "Choose existing template." })).toBeVisible();
+    await expect(importPage.importButton).toBeDisabled();
```
nhanlon2, line 124: This comment is from Claude. No file has been uploaded at this point, and the Import button is disabled for any record type with no file, so this assertion passes whether or not the instrument logic works.

To make it prove something: upload the instruments CSV, choose "Choose existing template", and assert disabled in that state.

nhanlon2 (commented):
This comment is from Claude. Automated parallel review of the frontend, i18n, security and quality dimensions, generated with [Claude Code](https://claude.com/claude-code). 8 findings posted inline, each re-checked against the source before posting; suggestions and lower-severity items were reviewed locally and left off the PR.

Nothing blocking on the security side: no new XSS sink, no new endpoint, and the record-type query parameter is correctly allow-listed. The most consequential item is the `zipWith` pairing in `ImportModel.ts`, which a user can crash by renaming a custom field. There is also a recurring theme of instrument code being copied from the sample code rather than shared, which is where three of the comments land.

src/main/webapp/ui/src/Inventory/Import/Fields/ColumnFieldMapping.tsx
mrowe-tan: I've made the suggested change.

src/main/webapp/ui/src/stores/models/ImportModel.ts
mrowe-tan: This is a valid comment, but does not appear to actually be a scenario that can be reached in the UI, so I'll leave this for now.

src/main/webapp/ui/src/stores/models/ImportModel.ts
nhanlon2: From @nhanlon2  I have confirmed ths bug is real in a manual test of the GUI

src/main/webapp/ui/src/stores/models/ImportModel.ts
mrowe-tan: The `custom` field is not excluded for Instruments, as this is a mapping that we want to allow during import, the same as we do for Samples, which also does not exclude the `custom` field.  i.e.  This is an intended feature, not an oversight.

From my own claude analysis, it further clarifies with the following:

"Fields.custom is the value assigned to each mapping in instrumentsMappings
  when the CSV is parsed (line 836: field: Fields.custom). Those rows represent
  custom template fields extracted from the CSV — the instrument template's
  user-defined fields. Excluding "custom" from the dropdown would mean the user
  could not keep a column mapped to a custom template field, which defeats the
  purpose of those rows entirely.

  The reason "custom" is excluded for CONTAINERS and SUBSAMPLES is different:
  neither of those record types supports user-defined template fields, so
  offering custom in their column mapping dropdowns would be meaningless. For
  SAMPLES and INSTRUMENTS, custom template fields are a core part of the data
  model, so "custom" must stay available."

src/main/webapp/ui/src/__tests__/e2e/components/inventory/InventoryCreateMenu.ts
```diff
@@ -37,8 +35,8 @@ export class InventoryCreateMenu {
     await this.menuItem(item).click();
   }
 
-  async clickCsvImport(item: InventoryCsvImportItem): Promise<void> {
-    await this.menuItem(item).click();
+  async clickCsvImport(): Promise<void> {
+    await this.root.getByRole("menuitem", { name: /From CSV/ }).click();
```
tchoi-rs: Unnecessary regex

src/main/webapp/ui/src/__tests__/e2e/specs/inventory/import.e2e.ts
tchoi-rs: Yes, this test looks incomplete

src/main/webapp/ui/src/components/Inputs/SelectedFileInfo.tsx
```diff
@@ -40,8 +40,8 @@ function StatusChip({
     [() => true, t("values.none")],
   ])();
   const iconStyle = {
-    width: 18,
-    height: 18,
+    width: "18px !important",
```
tchoi-rs, line 43: Have you looked into how to get rid of these `!important`s? It's fine if they are too hard to remove

src/main/webapp/ui/src/assets/graphics/RecordTypeGraphics/Icons/Template.tsx
```diff
@@ -6,7 +6,7 @@ import { mergeSx } from "../../../../modules/common/utils/styles";
 function TemplateIcon({ sx }: { sx?: SxProps<Theme> }): React.ReactNode {
   return (
     <SvgIcon
-      viewBox="0 0 16 16"
+      viewBox="-1 -1 16 16"
```
tchoi-rs: Is this intentional? I think you should get a new SVG export instead

src/main/webapp/ui/src/Inventory/Import/Fields/TemplateName.tsx
```diff
@@ -20,15 +19,14 @@ function TemplateName({ disabled }: TemplateNameArgs): ReactNode {
...
       error={error}
-      id="templateNameField" // for a11y
-      value={importStore.importData?.templateName ?? ""}
+      id="templateNameField"
+      value={value}
       helperText={error ? t("import.fields.templateNameValidation") : ""}
       onChange={({ target }) => {
-        if (target instanceof HTMLInputElement && importStore.importData)
-          importStore.importData?.setTemplateName(target.value);
+        if (target instanceof HTMLInputElement) onChange(target.value);
```
tchoi-rs: Not your change but the `instanceof` is redundant and can go

src/main/webapp/ui/src/stores/models/ImportModel.ts
```diff
@@ -557,19 +646,29 @@ export default class Import {
...
+    );
   }
 
   get fileByRecordTypeLoaded(): boolean {
     return this.isContainersImport
       ? Boolean(this.containersFile)
       : this.isSamplesImport
         ? Boolean(this.samplesFile)
-        : Boolean(this.subSamplesFile);
+        : this.isInstrumentsImport
```
tchoi-rs, line 671: I think it's time these ternaries get rewritten into if/switches

src/main/webapp/ui/src/stores/models/ImportModel.ts
```diff
@@ -872,36 +1045,58 @@ export default class Import {
...
             : null,
-          subSampleSettings: this.subSamplesSubmittable
+          subSampleSettings: subSamplesSubmittable
             ? {
                 fieldMappings: this.makeMappingsObject(this.subSamplesMappings),
               }
             : null,
+          instrumentSettings: instrumentsSubmittable
+            ? {
+                // biome-ignore lint/style/noNonNullAssertion: instrumentTemplate is non-null when !instrumentCreateNewTemplate and instrumentsSubmittable is true
```
tchoi-rs: If possible, I would tighten the type declaration to make this biome-ignore unnecessary

tchoi-rs (changes requested):

src/main/webapp/ui/src/stores/models/ImportModel.ts
mrowe-tan: Whilst I get what Claude's saying here, I don't think I want to genericise these ones.  `isInstrumentsImport` and `transformTemplateInfoForSubmission` may differ only by additional fields, but that's enough in my mind to want to keep them separate, in case there's any other behaviour changes we want in the future.  As for `importInstrumentMatchesExistingTemplate`, whilst that could be combined with the other one, it would need a lot of the internals of the `importPage` object passed through as parameters, which makes things messier, and breaks the nice functional encapsulation that they have at present.  If we ever need yet another object to be imported by CSV I'd consider making a wrapper function for some of these, maybe, but whilst we only have Samples and Instruments at the moment I think it's okay to leave it as is.

src/main/webapp/ui/src/components/Inputs/SelectedFileInfo.tsx
mrowe-tan: Yes, these were being overridden with other stylings when the page rendered, hence having to resort to this.

src/main/webapp/ui/src/stores/models/ImportModel.ts
mrowe-tan: No argument here.

src/main/webapp/ui/src/assets/graphics/RecordTypeGraphics/Icons/Template.tsx
mrowe-tan: I've modified the path and restored the viewBox settings.

src/main/webapp/ui/src/__tests__/e2e/specs/inventory/import.e2e.ts
```diff
@@ -135,5 +136,11 @@ test.describe(`Inventory CSV Import`, { tag: [tags.INVENTORY, tags.MOBILE] }, ()
     await expect(importPage.main.getByRole("radio", { name: "Create new template." })).toBeVisible();
     await expect(importPage.main.getByRole("radio", { name: "Choose existing template." })).toBeVisible();
     await expect(importPage.importButton).toBeDisabled();
+
+    await importPage.uploadCsv(INSTRUMENTS_CSV);
+    await expect(importPage.importButton).toBeEnabled();
+
+    await importPage.chooseExistingTemplate();
+    await expect(importPage.importButton).toBeDisabled();
```
tchoi-rs, line 130: Could the e2e test also check for the data that's been imported?

tchoi-rs (changes requested):

src/main/webapp/ui/src/__tests__/e2e/specs/inventory/import.e2e.ts
mrowe-tan: I've merged the two Instruments import tests into one - seems to make more sense that way.

tchoi-rs (approved):

tchoi-rs:
Frontend portion looks fine!

nhanlon2 (approved):

PR #1054, opened 2026-08-24
RSDEV-1253: Map `Calibration` and `Measurement Technique` for PIDINST (DataCite and B2inst)

rs-nicof:
related PR: https://github.com/rspace-os/datacite-java-client/pull/6
AWS instance: https://rsdev-1253-map-clibration-and-measurement-f50c365a-11.researchspace.com

## Description ##

- Resolves https://researchspace.atlassian.net/browse/RSDEV-1253

An instrument's **Measurement technique** and **Calibration** fields hold links to other RSpace records: the document describing how a measurement is taken, the certificate behind a calibration. Until now those links stayed inside RSpace and were left out of the instrument's registered metadata, so the published identifier said nothing about the material behind it.

Registering an instrument PID now sends both links to the provider as related identifiers, for B2INST and DataCite alike. Each entry carries the linked record's RSpace address, labelled "Measurement Technique" or "Calibration".

A field that is empty, or whose link has since been cleared, is simply left out. Nothing changes in the UI, and no API or database changes were needed.

## Design decisions

- **The relation we send is always `IsDescribedBy`.** The PIDINST standard's list of permitted relations contains neither of the two the fields actually store (`IsDocumentedBy`, `IsCalibratedBy`), and `IsDescribedBy` is the one usable option in it. The relation stored on the link inside RSpace is untouched.
- **The address we send is the linked record's ordinary RSpace page**, which asks for a sign-in. Unlike a landing page, a related identifier makes no promise of being publicly resolvable, and the linked records generally have no public page to point at.
- **B2INST is told once, at registration**, as it has no way to update a record's metadata afterwards. **DataCite is told on publish and again on retract**, since both resend the full metadata and the entries would otherwise be dropped.
- If no usable web address can be built, for example on a deployment with no server URL configured, the entry is left out with a warning rather than registered wrong. Same rule as the landing page: a missing value can be added later, a published wrong one cannot.
- **Last calibrated** is still not mapped, as the standard has no property that fits it.
- The reasoning and the rejected alternatives are recorded in `DevDocs/adr/0007-pidinst-related-identifiers.md`.

## Testing notes

- ⚠️ **Before merge:** `datacite-java-client` needs a release (suggested `1.1.0`) and the pin in `pom.xml` swapped off `RSDEV-1253-datacite-SNAPSHOT`.
- Unit tests cover the mapping rules; a new Spring test covers both providers end to end against real instruments and links, including that publish and retract each send the entries.

### DIFF
--- a/DevDocs/adr/0005-measured-variable-narratives.md
+++ b/DevDocs/adr/0005-measured-variable-narratives.md
@@ -7,15 +7,14 @@ status: superseded
 > **Superseded (2026-07-27), during RSDEV-1220, before any of this shipped.**
 > The narratives were dropped. `MeasuredVariable` now carries only the *Measured
 > quantity* field's content verbatim, which is what the property was designed
-> for. *Measurement technique*, *Calibration* and *Last calibrated* are no longer
-> mapped to anything and stay on the template as documentation-only fields (see
-> `CONTEXT.md`). The reasoning below is kept because the underlying problem is
-> unchanged — PIDINST still has no home for those three facts — so anyone
-> proposing to map them again should start here and read the Considered options.
->
-> RSDEV-1253 later did map the two link fields, resolving the vocabulary clash by
-> fixing the sent relation to `IsDescribedBy`; see ADR 0007. *Last calibrated* alone
-> remains unmapped.
+> for. Nothing narrative reaches `MeasuredVariable` from the other fields.
+> *Measurement technique* and *Calibration* were documentation-only from then
+> until RSDEV-1253 mapped them to `RelatedIdentifier` entries, resolving the
+> vocabulary clash by fixing the sent relation to `IsDescribedBy` (see ADR
+> 0007). Only *Last calibrated* remains unmapped, PIDINST's Date vocabulary
+> still having no home for it. The reasoning below is kept because it explains
+> why the narrative route stays closed; anyone proposing to map these fields
+> another way should start here and read the Considered options.
 
 ## Context
 
--- a/DevDocs/adr/0005-measured-variable-narratives.md
+++ b/DevDocs/adr/0005-measured-variable-narratives.md
@@ -12,6 +12,10 @@ status: superseded
 > `CONTEXT.md`). The reasoning below is kept because the underlying problem is
 > unchanged — PIDINST still has no home for those three facts — so anyone
 > proposing to map them again should start here and read the Considered options.
+>
+> RSDEV-1253 later did map the two link fields, resolving the vocabulary clash by
+> fixing the sent relation to `IsDescribedBy`; see ADR 0007. *Last calibrated* alone
+> remains unmapped.
 
 ## Context
 
--- a/pom.xml
+++ b/pom.xml
@@ -1779,7 +1779,7 @@
     <dependency>
       <groupId>com.github.rspace-os</groupId>
       <artifactId>datacite-java-client</artifactId>
-      <version>RSDEV-1253-datacite-SNAPSHOT</version>
+      <version>1.1.0</version>
     </dependency>
     <dependency>
       <groupId>com.github.rspace-os</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -1779,7 +1779,7 @@
     <dependency>
       <groupId>com.github.rspace-os</groupId>
       <artifactId>datacite-java-client</artifactId>
-      <version>1.0.0</version>
+      <version>RSDEV-1253-datacite-SNAPSHOT</version>
     </dependency>
     <dependency>
       <groupId>com.github.rspace-os</groupId>
--- a/src/main/java/com/researchspace/service/inventory/RspaceToExternalProviderAdapter.java
+++ b/src/main/java/com/researchspace/service/inventory/RspaceToExternalProviderAdapter.java
@@ -7,9 +7,10 @@ import com.researchspace.model.inventory.InventoryRecord;
 
 /**
  * Translates an RSpace inventory record into the domain wrapper of the requested external PID
- * provider. This is the seam the future generic registration endpoints (RSDEV-1209) will build on;
- * in this story only the B2INST path is routed through it, while DataCite delegates to the existing
- * {@link ApiInventoryDOI#convertToDataCiteDoi()}.
+ * provider. This is the seam the future generic registration endpoints (RSDEV-1209) will build on.
+ * The B2INST path is built here in full; the DataCite path delegates the base conversion to {@link
+ * ApiInventoryDOI#convertToDataCiteDoi()} and adds the PIDINST properties that have no home on the
+ * RSpace DOI representation (ADR 0007).
  *
  * <p>Provider guard rails: {@code PIDINST_*} providers accept only Instrument records ({@code
  * IN*}); {@code IGSN_*} providers accept only {@code IC*}/{@code SA*}/{@code SS*}.
@@ -25,10 +26,11 @@ public interface RspaceToExternalProviderAdapter {
    * wrong (see ADR 0006). Owner always has exactly one entry (field content, falling back to the
    * record owner).
    *
-   * <p>Not every template field is mapped. "Measurement technique", "Calibration" and "Last
-   * calibrated" are documentation-only and feed nothing, because PIDINST has no property that fits
-   * them; see CONTEXT.md ("Documentation-only field") and the superseded
-   * DevDocs/adr/0005-measured-variable-narratives.md for the attempt that was rejected.
+   * <p>The "Measurement technique" and "Calibration" link fields are mapped to RelatedIdentifier
+   * entries (fixed IsDescribedBy relation, the target's globalId page as a URL; RSDEV-1253, ADR
+   * 0007). "Last calibrated" remains documentation-only, because PIDINST has no property that fits
+   * it; see CONTEXT.md ("Documentation-only field") and
+   * DevDocs/adr/0005-measured-variable-narratives.md.
    *
    * <p>Must be called inside an existing transaction: the mapping reads the instrument's lazy
    * associations. The implementation declares {@code @Transactional(propagation = MANDATORY)}, so a
@@ -42,6 +44,24 @@ public interface RspaceToExternalProviderAdapter {
    */
   B2instDoi buildB2instDoi(InventoryRecord instrument, String publicLandingPageUrl);
 
-  /** Build the DataCite DOI wrapper from the RSpace DOI representation. */
-  DataCiteDoi buildDataCiteDoi(ApiInventoryDOI doi);
+  /**
+   * Build the DataCite DOI wrapper: the RSpace DOI representation converted by {@link
+   * ApiInventoryDOI#convertToDataCiteDoi()}, then, when the associated record is an Instrument, the
+   * PIDINST related identifiers appended from the Measurement technique and Calibration link fields
+   * (RSDEV-1253, ADR 0007): the link target's globalId page as a URL, always related as
+   * IsDescribedBy, labelled through relationTypeInformation. Entries whose address would not be an
+   * absolute http(s) URL are omitted with a WARN rather than sent wrong.
+   *
+   * <p>Callers resending full metadata (publish and retract both do) must route through this
+   * method, not {@code convertToDataCiteDoi()} directly, or the registered related identifiers
+   * silently regress.
+   *
+   * <p>Must be called inside an existing transaction when the record is an Instrument: the mapping
+   * reads the instrument's lazy fields. The implementation declares
+   * {@code @Transactional(propagation = MANDATORY)}, matching {@link #buildB2instDoi}.
+   *
+   * @param associatedRecord the inventory record the DOI belongs to; null or a non-Instrument
+   *     yields the plain conversion
+   */
+  DataCiteDoi buildDataCiteDoi(ApiInventoryDOI doi, InventoryRecord associatedRecord);
 }
--- a/src/main/java/com/researchspace/service/inventory/RspaceToExternalProviderAdapter.java
+++ b/src/main/java/com/researchspace/service/inventory/RspaceToExternalProviderAdapter.java
@@ -50,9 +50,11 @@ public interface RspaceToExternalProviderAdapter {
    * PIDINST related identifiers appended from the Measurement technique and Calibration link fields
    * (RSDEV-1253, ADR 0007): the link target's globalId page as a URL, carrying the link's version
    * pin when the target type resolves one, always related as IsDescribedBy, labelled through
-   * relationTypeInformation. An instrument with no live links sends an explicit empty list, which
-   * is how DataCite is told to clear the property. Entries whose address would not be an absolute
-   * http(s) URL are omitted with a WARN rather than sent wrong.
+   * relationTypeInformation. A link whose target is deleted, or no longer readable by the
+   * instrument's owner, is omitted with a WARN. An instrument whose entries come to nothing sends
+   * an explicit empty list, which is how DataCite is told to clear the property - except when no
+   * usable http(s) server URL exists, an environment failure under which the property is left
+   * untouched rather than wrongly cleared.
    *
    * <p>Callers resending full metadata (publish and retract both do) must route through this
    * method, not {@code convertToDataCiteDoi()} directly, or the registered related identifiers
--- a/src/main/java/com/researchspace/service/inventory/RspaceToExternalProviderAdapter.java
+++ b/src/main/java/com/researchspace/service/inventory/RspaceToExternalProviderAdapter.java
@@ -27,10 +27,10 @@ public interface RspaceToExternalProviderAdapter {
    * record owner).
    *
    * <p>The "Measurement technique" and "Calibration" link fields are mapped to RelatedIdentifier
-   * entries (fixed IsDescribedBy relation, the target's globalId page as a URL; RSDEV-1253, ADR
-   * 0007). "Last calibrated" remains documentation-only, because PIDINST has no property that fits
-   * it; see CONTEXT.md ("Documentation-only field") and
-   * DevDocs/adr/0005-measured-variable-narratives.md.
+   * entries (fixed IsDescribedBy relation, the target's globalId page as a URL, carrying the link's
+   * version pin when the target type resolves a version-suffixed id; RSDEV-1253, ADR 0007). "Last
+   * calibrated" remains documentation-only, because PIDINST has no property that fits it; see
+   * CONTEXT.md ("Documentation-only field") and DevDocs/adr/0005-measured-variable-narratives.md.
    *
    * <p>Must be called inside an existing transaction: the mapping reads the instrument's lazy
    * associations. The implementation declares {@code @Transactional(propagation = MANDATORY)}, so a
@@ -48,9 +48,11 @@ public interface RspaceToExternalProviderAdapter {
    * Build the DataCite DOI wrapper: the RSpace DOI representation converted by {@link
    * ApiInventoryDOI#convertToDataCiteDoi()}, then, when the associated record is an Instrument, the
    * PIDINST related identifiers appended from the Measurement technique and Calibration link fields
-   * (RSDEV-1253, ADR 0007): the link target's globalId page as a URL, always related as
-   * IsDescribedBy, labelled through relationTypeInformation. Entries whose address would not be an
-   * absolute http(s) URL are omitted with a WARN rather than sent wrong.
+   * (RSDEV-1253, ADR 0007): the link target's globalId page as a URL, carrying the link's version
+   * pin when the target type resolves one, always related as IsDescribedBy, labelled through
+   * relationTypeInformation. An instrument with no live links sends an explicit empty list, which
+   * is how DataCite is told to clear the property. Entries whose address would not be an absolute
+   * http(s) URL are omitted with a WARN rather than sent wrong.
    *
    * <p>Callers resending full metadata (publish and retract both do) must route through this
    * method, not {@code convertToDataCiteDoi()} directly, or the registered related identifiers
--- a/src/main/java/com/researchspace/service/inventory/RspaceToExternalProviderAdapter.java
+++ b/src/main/java/com/researchspace/service/inventory/RspaceToExternalProviderAdapter.java
@@ -56,9 +56,13 @@ public interface RspaceToExternalProviderAdapter {
    * method, not {@code convertToDataCiteDoi()} directly, or the registered related identifiers
    * silently regress.
    *
-   * <p>Must be called inside an existing transaction when the record is an Instrument: the mapping
-   * reads the instrument's lazy fields. The implementation declares
-   * {@code @Transactional(propagation = MANDATORY)}, matching {@link #buildB2instDoi}.
+   * <p>Must be called inside an existing transaction, whatever the record: the implementation
+   * declares {@code @Transactional(propagation = MANDATORY)}, matching {@link #buildB2instDoi}, so
+   * a caller without one fails immediately rather than part-way through the mapping. Only the
+   * Instrument path actually needs the session, to read the instrument's lazy fields, but the
+   * requirement is deliberately uniform: both production callers (publish and retract) are already
+   * transactional, and a transaction requirement that changed with the argument's runtime type
+   * would be the harder contract to honour.
    *
    * @param associatedRecord the inventory record the DOI belongs to; null or a non-Instrument
    *     yields the plain conversion
--- a/src/main/java/com/researchspace/service/inventory/impl/RspaceToExternalProviderAdapterImpl.java
+++ b/src/main/java/com/researchspace/service/inventory/impl/RspaceToExternalProviderAdapterImpl.java
@@ -12,13 +12,18 @@ import com.researchspace.b2inst.model.metadata.B2instInstrumentType;
 import com.researchspace.b2inst.model.metadata.B2instManufacturer;
 import com.researchspace.b2inst.model.metadata.B2instModel;
 import com.researchspace.b2inst.model.metadata.B2instOwner;
+import com.researchspace.b2inst.model.metadata.B2instRelatedIdentifier;
 import com.researchspace.b2inst.model.request.B2instDoi;
 import com.researchspace.datacite.model.DataCiteDoi;
+import com.researchspace.datacite.model.DataCiteDoiAttributes;
 import com.researchspace.model.User;
 import com.researchspace.model.field.FieldType;
 import com.researchspace.model.inventory.InstrumentEntity;
 import com.researchspace.model.inventory.InventoryRecord;
 import com.researchspace.model.inventory.field.InventoryEntityField;
+import com.researchspace.model.inventory.field.InventoryLink;
+import com.researchspace.properties.IPropertyHolder;
+import com.researchspace.service.inventory.InventoryUrls;
 import com.researchspace.service.inventory.RspaceToExternalProviderAdapter;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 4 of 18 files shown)


rs-nicof:
@greptileai

rs-nicof:
@greptileai

src/main/java/com/researchspace/service/inventory/RspaceToExternalProviderAdapter.java
```diff
@@ -42,6 +44,24 @@ public interface RspaceToExternalProviderAdapter {
...
+   * IsDescribedBy, labelled through relationTypeInformation. Entries whose address would not be an
+   * absolute http(s) URL are omitted with a WARN rather than sent wrong.
+   *
+   * <p>Callers resending full metadata (publish and retract both do) must route through this
+   * method, not {@code convertToDataCiteDoi()} directly, or the registered related identifiers
+   * silently regress.
+   *
+   * <p>Must be called inside an existing transaction when the record is an Instrument: the mapping
+   * reads the instrument's lazy fields. The implementation declares
+   * {@code @Transactional(propagation = MANDATORY)}, matching {@link #buildB2instDoi}.
```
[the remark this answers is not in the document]
rs-nicof: Fixed in b18d2772, by correcting the contract rather than relaxing the propagation.

You are right that `MANDATORY` fails fast for null/non-Instrument conversions too, so the javadoc understated it. Behaviour is deliberately unchanged: both production callers (`createUpdateWithPublishedDoi` and `createUpdateWithRetractedDoi` in `InventoryIdentifierApiManagerImpl`) already run inside the Manager transaction, so nothing needs the propagation relaxed, and a transaction requirement that changed with the argument's runtime type would be the harder contract to honour. The javadoc now states the requirement as unconditional, matching `buildB2instDoi`, and says why only the Instrument path actually needs the session.

pom.xml
```diff
@@ -1812,7 +1812,7 @@
     <dependency>
       <groupId>com.github.rspace-os</groupId>
       <artifactId>datacite-java-client</artifactId>
-      <version>1.0.0</version>
+      <version>RSDEV-1253-datacite-SNAPSHOT</version>
     </dependency>
```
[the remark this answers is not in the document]
rs-nicof: Skipping this one deliberately, leaving the thread open for the human reviewer.

`RSDEV-1253-datacite-SNAPSHOT` is a temporary pin that pairs this branch with rspace-os/datacite-java-client#6, which adds the `relatedIdentifiers` field this change registers. It is how the two in-flight branches are built against each other while both are under review, and it is not what either merges with: before merge the client releases 1.1.0 and this pom is bumped to that released version, so nothing branch-named reaches a consumer or the release automation. Same comment answered the same way on the client PR.

rs-nicof:
@greptileai

src/main/java/com/researchspace/service/inventory/impl/RspaceToExternalProviderAdapterImpl.java
```diff
@@ -198,6 +222,60 @@ private List<String> measuredVariables(InstrumentEntity instrument) {
...
+   * unset or scheme-less) is omitted with a WARN rather than registered, because a wrong published
+   * value cannot be corrected and a missing one can.
+   */
+  private Optional<String> pidinstLinkUrl(InstrumentEntity instrument, String canonicalName) {
+    Optional<InventoryLink> link = PidinstFields.mappedLink(instrument, canonicalName);
+    if (link.isEmpty()) {
+      return Optional.empty();
+    }
+    Optional<String> url =
+        InventoryUrls.globalIdPageUrl(properties.getServerUrl(), link.get().getTargetGlobalId())
```
rs-nicof: This registers the target's base globalId, so the link's version pin is dropped. A Calibration link pinned to v4 ends up registering an address that follows the record's latest state, which is the opposite of what pinning is for, and the registered value is permanent.

`InventoryLink` keeps the version in a separate `versionPin` field (hence the unsuffixed `targetGlobalId`), and `GlobalLookupController` does resolve versioned ids such as `SA42v4` for SA/SS/IC/IT/IN/NT, though not for Gallery targets. So honouring the pin is possible, but only for those prefixes. Either way it is worth stating the choice in ADR 0007, which currently does not mention the pin at all.

src/main/java/com/researchspace/service/inventory/impl/RspaceToExternalProviderAdapterImpl.java
```diff
@@ -198,6 +222,60 @@ private List<String> measuredVariables(InstrumentEntity instrument) {
...
+  }
+
+  /**
+   * The registrable address of the record this link field points at, or empty when the field holds
+   * no live link. Guarded like the LandingPage: an address without an http(s) scheme (server URL
+   * unset or scheme-less) is omitted with a WARN rather than registered, because a wrong published
+   * value cannot be corrected and a missing one can.
+   */
+  private Optional<String> pidinstLinkUrl(InstrumentEntity instrument, String canonicalName) {
+    Optional<InventoryLink> link = PidinstFields.mappedLink(instrument, canonicalName);
```
rs-nicof, line 271: The link target gets no read-permission check at registration time. READ is checked only when the link is first created; registering and publishing is gated on edit rights on the instrument, and those come from sharing rather than ownership.

So someone with edit access to a shared instrument can cause a permanent, publicly visible registry entry that names a record they themselves cannot open, and the same happens if the target's sharing is revoked after the link was made, since publish and retract recompute the entries with no re-check. Only the record's type and internal number are exposed and the address itself stays behind sign-in, so this may well be acceptable. Worth being a deliberate decision recorded in ADR 0007 rather than an implicit one.

src/main/java/com/researchspace/service/inventory/impl/RspaceToExternalProviderAdapterImpl.java
```diff
@@ -211,7 +289,39 @@ private <T> List<T> nullIfEmpty(List<T> list) {
...
-    return doi.convertToDataCiteDoi();
+  @Transactional(propagation = Propagation.MANDATORY)
+  public DataCiteDoi buildDataCiteDoi(ApiInventoryDOI doi, InventoryRecord associatedRecord) {
+    DataCiteDoi dataCiteDoi = doi.convertToDataCiteDoi();
+    if (associatedRecord == null || !associatedRecord.isInstrument()) {
+      return dataCiteDoi;
+    }
+    InstrumentEntity instrument = (InstrumentEntity) associatedRecord;
+    List<DataCiteDoiAttributes.RelatedIdentifier> related = new ArrayList<>();
+    pidinstLinkUrl(instrument, FIELD_MEASUREMENT_TECHNIQUE)
```
rs-nicof: Which link fields get sent, and in what order, is decided twice: here for DataCite and in `relatedIdentifiers()` for B2INST. Adding a third PIDINST link field later means editing both places, and nothing fails if only one is updated, because each provider's tests assert two entries and would stay green.

The two constructors also take four positional strings in different orders, two of which are the interchangeable constants `URL` and `IsDescribedBy`, so a swap would compile and ship a wrong registered payload. Both are correct today and pinned by assertions. One traversal parameterised by a per-provider constructor would single-source the decision.

rs-nicof (commented):
Reviewed the RSDEV-1253 changes with five parallel reviewers (layering, persistence, security, i18n, general quality). Three findings below need a decision. A fourth, the DataCite `relatedIdentifiers` empty-list clear, is already fixed in 62254e31a. The `pom.xml` SNAPSHOT pin is already covered by the existing open thread.

src/main/java/com/researchspace/service/inventory/impl/RspaceToExternalProviderAdapterImpl.java
rs-nicof: Fixed in dc2daceef. The pin is now carried as `<id>v<pin>`, but only for the target types whose version-suffixed globalId actually resolves to that version, mirroring `GlobalLookupController.VERSIONED_INVENTORY_PREFIXES`. A pin on any other allowed target is dropped rather than registered: NB has no versioned route at all, and SD's and GL's go to an audit view and a file stream rather than the record page, so there the unpinned address is the safer thing to make permanent. An id already carrying a suffix is left alone rather than doubled. Two unit tests pin both branches, and ADR 0007 now records the rule.

src/main/java/com/researchspace/service/inventory/impl/RspaceToExternalProviderAdapterImpl.java
rs-nicof: Fixed in dc2daceef. There is now one traversal holding the field list and the order, with each provider passing a builder for its own type, so a third link field is added in one place. That also leaves one constructor call per provider instead of one per provider per field, which is the part that mattered given the two types take the same four values in different positional orders.

src/main/java/com/researchspace/service/inventory/impl/RspaceToExternalProviderAdapterImpl.java
rs-nicof: Looked at this against the existing permission model and decided to keep the behaviour, recording it as a deliberate decision in ADR 0007 (1d76c9f81).

In short: READ is enforced where the link is written, and from then on the target's globalId is part of the instrument's own data. RSpace already shows that id to every instrument viewer even when the target is unreadable (the link summary redacts name and type, never the id), so the registered entry exposes nothing an instrument viewer does not already see, and the address only resolves after sign-in plus the target's own checks. Adding a re-check at publish time would also make the registered payload depend on which editor happens to click publish on a shared instrument, so the same instrument could register different metadata depending on who acted, which felt worse than the disclosure it would prevent.

rs-nicof:
AWS instance: https://rsdev-1253-map-clibration-and-measurement-f50c365a-11.researchspace.com

pom.xml
```diff
@@ -1812,7 +1812,7 @@
     <dependency>
       <groupId>com.github.rspace-os</groupId>
       <artifactId>datacite-java-client</artifactId>
-      <version>1.0.0</version>
+      <version>RSDEV-1253-datacite-SNAPSHOT</version>
```
nhanlon2: **Blocking.** This pins a mutable JitPack branch build. Two checks worth noting: the released `datacite-java-client-1.0.0` jar has no `RelatedIdentifier` nested class on `DataCiteDoiAttributes`, so there is no working fallback version, and `DevDocs/adr/0007-pidinst-related-identifiers.md:68` already states `>= 1.1.0` is required, so the POM and the ADR currently contradict each other.

JitPack rebuilds `<branch>-SNAPSHOT` from whatever is latest on that branch, and artifacts are only guaranteed immutable after seven days, so what ships can change with no commit here and nothing in CI would notice. Every other `com.github.rspace-os` dependency in this POM uses a fixed release (`raid-java-client` 4.0.0, `zenodo-java-client` 1.0.0).

Cut the `1.1.0` release and pin it, or at minimum pin the immutable commit hash.

src/main/java/com/researchspace/service/inventory/impl/RspaceToExternalProviderAdapterImpl.java
```diff
@@ -211,7 +329,26 @@ private <T> List<T> nullIfEmpty(List<T> list) {
...
+            (url, label) ->
+                new DataCiteDoiAttributes.RelatedIdentifier(
+                    RELATION_TYPE_IS_DESCRIBED_BY, url, RELATED_ID_TYPE_URL, label));
+    // Set unconditionally, the empty list included. DataCite replaces the whole property with
+    // whatever the payload carries and clears it only when sent an explicit empty array; a property
+    // that is absent or null leaves the registered value alone. An instrument whose link fields
+    // have
+    // all been cleared therefore has to send [], or the entries registered before the user cleared
+    // them stay attached to a findable DOI with no way to withdraw them.
+    dataCiteDoi.getAttributes().setRelatedIdentifiers(related);
```
nhanlon2, line 396: **Blocking: this can silently destroy already-registered metadata.**

The comment is right that an instrument whose fields were cleared must send `[]`. The problem is that `[]` is also what gets sent when the links are still there but no address could be built for them.

`pidinstLinkUrl` (lines 279-297) returns empty with only a WARN when `globalIdPageUrl(...).filter(isResolvableAddress)` yields nothing. That result is indistinguishable here from "no links". So a deployment whose `rs.serverurl` is unset, scheme-less, or changed to a scheme-less form, that then publishes or retracts a previously-published instrument DOI, sends `relatedIdentifiers: []` and **permanently strips entries that were correct**. The only signal is one WARN line.

That inverts the principle stated three times in this file and in ADR 0007: a missing value can be added later, a published wrong one cannot. Here the omission *is* the destructive act, and the ADR's empty-list rationale only contemplates the user clearing the fields.

Suggestion: have the mapping distinguish "no links" from "links present but unaddressable", and in the second case leave the property null (or refuse the publish) rather than clearing.

Note the existing guard test at `RspaceToExternalProviderAdapterImplTest.java:706-713` covers only the B2INST branch, where the outcome is a harmless `null`.

src/main/java/com/researchspace/service/inventory/impl/RspaceToExternalProviderAdapterImpl.java
```diff
@@ -198,6 +249,73 @@ private List<String> measuredVariables(InstrumentEntity instrument) {
...
+   * unsuffixed by contract ({@code InventoryLinkManagerImpl.applyApiToEntity} keeps the version in
+   * versionPin), and one already carrying a suffix is left alone rather than doubled.
+   */
+  private String registrableTargetGlobalId(InventoryLink link) {
+    String target = StringUtils.trimToEmpty(link.getTargetGlobalId());
+    Long versionPin = link.getVersionPin();
+    if (versionPin == null) {
+      return target;
+    }
+    GlobalIdentifier oid = new GlobalIdentifier(target);
```
nhanlon2: `new GlobalIdentifier(target)` throws `IllegalArgumentException` on any string not matching `([A-Z][A-Z])(\d+)(v(\d+))?` or carrying an unknown prefix. This method runs under `@Transactional(MANDATORY)`, and this class's own comment notes that an escaping exception marks the caller's shared transaction rollback-only.

So a single malformed `target_globalid` row on a version-pinned link fails the **entire** publish, rather than omitting one related identifier. That is the opposite of every other guard rail in this file.

The clean fix avoids the parse entirely: `InventoryLink` already persists `targetPrefix` as a typed `GlobalIdPrefix` column, so `link.getTargetPrefix()` can be checked against `VERSION_PINNABLE_TARGETS` directly with no re-parsing and no throw. The codebase's existing convention for this call when it is kept is a try/catch, see `InventoryLinkValidator.java:33-39` and `ApiInventoryLink.java:38-46`.

Unreachable today, since writes go through `InventoryLinkManagerImpl.java:168` which round-trips through `GlobalIdentifier`. Worth fixing anyway given the blast radius, and it is currently untested.

src/main/java/com/researchspace/service/inventory/impl/RspaceToExternalProviderAdapterImpl.java
```diff
@@ -48,12 +58,45 @@ public class RspaceToExternalProviderAdapterImpl implements RspaceToExternalProv
...
+   * GlobalLookupController.VERSIONED_INVENTORY_PREFIXES}, which is the authority. A pin on any
+   * other allowed target is deliberately not registered: NB has no versioned route at all, and SD's
+   * and GL's lead to an audit view and a file stream rather than the record's page, so for those
+   * the unpinned address stays the safer thing to make permanent.
+   */
+  private static final Set<GlobalIdPrefix> VERSION_PINNABLE_TARGETS =
+      EnumSet.of(
+          GlobalIdPrefix.SA,
+          GlobalIdPrefix.SS,
+          GlobalIdPrefix.IC,
```
nhanlon2: This is now the third private copy of the same six-prefix set `{SA, SS, IC, IT, IN, NT}`. The others are `GlobalLookupController.VERSIONED_INVENTORY_PREFIXES` (`GlobalLookupController.java:72-79`) and the equivalent in the link-target resolver.

The javadoc names the controller "the authority", but nothing enforces that: it is `private` and in the web layer, so the service cannot reference it (correctly avoided) and the copies can only drift. Adding a versioned inventory type means three edits with nothing to catch a miss, and a miss here registers a **permanently wrong address**.

Suggest hoisting to one shared constant somewhere both layers can reach (next to `GlobalIdPrefix`, or as a public constant on `InventoryUrls`) and having the controller consume it.

src/test/java/com/researchspace/service/inventory/InventoryIdentifierApiManagerRelatedIdentifierTest.java
```diff
@@ -0,0 +1,272 @@
...
+    assertTrue(
+        "this test needs rs.serverurl to carry an http(s) scheme, but it is: " + serverUrl,
+        StringUtils.startsWithIgnoreCase(serverUrl, "http://")
+            || StringUtils.startsWithIgnoreCase(serverUrl, "https://"));
+    user = createAndSaveUserIfNotExists(getRandomAlphabeticString("api"));
+    initialiseContentWithEmptyContent(user);
+  }
+
+  @After
+  public void tearDown() {
```
nhanlon2: **Blocking: this silently disables base-class cleanup for every later test class in the run.**

This overrides `SpringTransactionalTest.tearDown()` (`SpringTransactionalTest.java:73-76`) without calling `super.tearDown()`. Narrowing away `throws Exception` is a legal override, so it compiles and the base method simply never runs.

`setUp` calls `initialiseContentWithEmptyContent(user)`, which sets `contentInitializer.setCustomInitActive(false)`. The base `tearDown` is what restores it to `true`. Because the Spring context and that bean are cached across test classes, **every subsequent test class in the same JVM gets empty content initialisation**. The base class javadoc warns about exactly this, and 109 other test classes call `super.tearDown()`.

```java
@After
public void tearDown() throws Exception {
  super.tearDown();
  ReflectionTestUtils.setField(inventoryIdentifierApiMgr, "b2instConnector", realB2instConnector);
}
```

Separately, the restoration comment here is good and the reasoning is right; it just needs to also let the base class do its half.

Two smaller notes on this class:
- It extends `SpringTransactionalTest`, so the test's own transaction is what satisfies `buildDataCiteDoi`'s `MANDATORY` propagation. The test therefore cannot fail if the `inventoryManagerTx` advisor stopped matching `publishIdentifier`. A `flush()` + `clear()` before the publish call would make the "resolves the link through Hibernate" claim real rather than served from the first-level cache.
- The retract assertion re-reads a dummy field that register, publish and retract all overwrite. It does catch the named regression, but it would also pass if retract never reached the connector at all, since the publish payload is still sitting there. Nulling the captured payload between stages, or asserting the retract state, makes the intent explicit.

DevDocs/adr/0005-measured-variable-narratives.md
```diff
@@ -12,6 +12,10 @@ status: superseded
 > `CONTEXT.md`). The reasoning below is kept because the underlying problem is
 > unchanged — PIDINST still has no home for those three facts — so anyone
 > proposing to map them again should start here and read the Considered options.
+>
+> RSDEV-1253 later did map the two link fields, resolving the vocabulary clash by
```
nhanlon2: This note is accurate, but it leaves the document self-contradicting. The supersession text immediately above (lines 6-13) still reads that *Measurement technique*, *Calibration* and *Last calibrated* "are no longer mapped to anything and stay on the template as documentation-only fields", and that "PIDINST still has no home for those three facts". Both are now false for two of the three.

Better to correct those original sentences than to rebut them four lines later. A reader who stops at the first paragraph gets the wrong answer.

nhanlon2 (commented):
## Automated review: rspace review agents (layering, i18n, security, persistence) plus general quality and test-coverage passes

Generated with [Claude Code](https://claude.com/claude-code). Six reviewers ran independently against `1d76c9f8`. I re-read the source myself to confirm the four blocking findings below before posting; the rest are reported as the reviewers found them. Inline comments cover what anchors to changed lines; the two findings below do not, so they are here.

---

### Blocking: a user can publish the URL of a record they cannot read

The write-time READ assertion is real (`InventoryLinkManagerImpl.java:155-161`), but it does not survive a copy:

- `InventoryLinkField.shallowCopy()` (core-model) deep-copies the link row carrying `targetGlobalId`, taking **no `User` argument and performing no permission check**.
- `duplicateInstrument` (`InstrumentEntityApiManagerImpl.java:648-650`) requires only `assertUserCanReadInstrument`. It calls `clearLandingPage(copy)` but does not clear link fields.

So: user A links private record `SD500` into a shared instrument (A passes the READ check). User B, who has read-only visibility of that instrument and no access to `SD500`, duplicates it, now owns the copy, and registers + publishes. `https://<server>/globalId/SD500` is now in DataCite's findable metadata, which is unauthenticated, indexed in DataCite Commons, and redistributed CC0 in the public data files. B never had READ on `SD500`.

The same-user variant needs no copy at all: a group member with read-only access to a PI's unpublished document can link it and publish, and the document's owner is never consulted.

ADR 0007 lines 44-48 justify the no-re-check on the grounds that RSpace already shows that id to every instrument viewer. That is disclosure to authenticated users holding READ on the instrument; publication moves the same id to the open internet, permanently. The ADR treats the two as equivalent without saying so.

On B2INST it is **unrecoverable**: the entries go into a record explicitly set to public access, B2INST has no metadata-update call, and `createUpdateWithRetractedB2instDoi` refuses retract outright.

Cheapest mitigation that also answers the ADR's own objection: re-check READ **as the instrument owner** rather than as the acting user. That is a stable actor, so it does not reintroduce the "different editors produce different metadata" problem the ADR rejects a re-check for.

### A deleted link *target* still gets registered permanently

`PidinstFields.mappedLink` filters `link.isDeleted()`, the link row, not the target's state. RSpace models those separately (`LinkTargetSnapshotResolverImpl` sets `summary.setDeleted(...)`, and the UI shows a "Target deleted" pill); deleting a sample does not soft-delete links pointing at it. Publishing an instrument whose Calibration target was deleted bakes a permanent URL to a record that no longer resolves. This is a missing code path and a missing ADR decision, not just a coverage gap.

---

### Also worth addressing (not blocking)

- **Undeclared DataCite schema floor.** `relationTypeInformation` was introduced in DataCite Metadata Schema **4.7** (March 2026). A registrar or Fabrica repository on 4.6 or earlier drops the label silently. Worth a line in ADR 0007 Consequences beside the client-version requirement. Related nuance also undocumented: B2INST's `relatedIdentifierName` names the *resource*, DataCite's `relationTypeInformation` describes the *relation*, so the same literal means two different things on the two wires.
- **`GLOBAL_ID_PATH` is a web-layer route now composed from the service layer.** Previously the service only *recognised* it, which tolerated the route moving; now a rename of `@RequestMapping("/globalId")` silently strands every already-registered URL. Making the constant public and having the controller consume it keeps the dependency direction downward.
- **No DataCite wire-format test.** The B2INST side has two (`RspaceToExternalProviderAdapterImplTest.java:566-623`), written precisely because getter assertions bypass Jackson. Nothing equivalent exists for DataCite, and the `RelatedIdentifier` type plus its `@JsonProperty` names arrive from an unreleased branch build. A renamed key ships green. The `[]`-clears-versus-null-leaves-alone behaviour the design depends on is also unverified at the serialization level.
- **The both-links-cleared case is never exercised end to end.** `publishOmitsALinkFieldThatHasSinceBeenCleared` clears one of two fields and lands on `size() == 1`, so the empty-list path, the ADR's headline decision, is covered by a single unit assertion on a hand-built instrument.
- **i18n debt in methods this diff edits.** `InventoryIdentifierApiManagerImpl.java:610-612` and `:646-648` still throw hard-coded English DataCite messages, while the B2INST siblings in the same class already resolve translated keys and the catalogue has a near neighbour at `server.inventory.json:82`.
- **Coverage gaps**, all unit-testable: no case for a wrong-typed field named "Calibration"; only SA of the six pinnable prefixes exercised; the retract assertion cannot distinguish a fresh payload from a stale one because the dummy's captured payload is never nulled between stages.

### Verified as correct

- **"No database changes needed" checks out.** `InventoryLink` with `version_pin` and its Envers table already exist on main via `changeLog-rsdev-1131.xml`, and core-model stays at 3.3.1. No Liquibase changeset is missing. Minor: "no API changes" overstates it slightly, since `buildDataCiteDoi` gained a parameter.
- **"Measurement Technique" and "Calibration" correctly stay hard-coded.** They are wire constants, not display copy, and the deliberate split between the template field name (lowercase t, a data matcher) and the registered label (capital T) is right and well documented.
- The `@Transactional(MANDATORY)` contract with a startup check that fails if the advice is absent, the absence of any lazy traversal to link targets, and the per-field assertions defending the positional-argument hazard between the two provider models were all called out positively.
- ADR 0007 was repeatedly praised for confronting the ADR 0006 tension head-on. The disagreement above is with one conclusion, not with the process; the decision being written down is what made it reviewable.

pom.xml
rs-nicof: Not fixing in this PR, deliberately. The snapshot is a temporary pin pairing this branch with rspace-os/datacite-java-client#6 while both are under review; before merge the client releases 1.1.0 and this line is swapped to that release, as ADR 0007 already requires. Agreed there is no working fallback version, which is exactly why the release has to land first.

src/main/java/com/researchspace/service/inventory/impl/RspaceToExternalProviderAdapterImpl.java
rs-nicof: Fixed in 5a069db42. Good catch, the clear could indeed fire on a broken deployment rather than on cleared fields. The mapping now treats a missing or scheme-less server URL as an environment failure and leaves the registered property untouched, while emptiness that reflects the data still clears with []. The both-fields-cleared case is now also proven end to end.

src/main/java/com/researchspace/service/inventory/impl/RspaceToExternalProviderAdapterImpl.java
rs-nicof: Fixed in 5a069db42, using the typed columns as you suggested. The pin decision no longer parses the stored id at all, so a malformed row can only disqualify its own entry and can no longer fail the whole publish. Covered by a new unit test.

src/main/java/com/researchspace/service/inventory/impl/RspaceToExternalProviderAdapterImpl.java
rs-nicof: Fixed in 5a069db42. The set now lives once, publicly, on InventoryUrls, and GlobalLookupController consumes it, so router and composer cannot drift; same treatment for the /globalId route itself, which the controller mapping now shares. The similar-looking sets in the two link-target resolvers were left alone on purpose: they route inventory-versus-ELN lookups, a different concept that happens to contain the same six members today.

src/test/java/com/researchspace/service/inventory/InventoryIdentifierApiManagerRelatedIdentifierTest.java
rs-nicof: Fixed in 5a069db42, thanks, that one was nasty. tearDown now calls super.tearDown() so later test classes keep their content initialisation. Both smaller notes taken too: publish now runs after a flush and clear so the links really resolve through Hibernate, and the captured payload is nulled between stages so the retract assertion can no longer be satisfied by the stale publish payload.

DevDocs/adr/0005-measured-variable-narratives.md
rs-nicof: Fixed in 5a069db42. The supersession note now states the current mapping directly instead of contradicting itself four lines later: nothing narrative reaches MeasuredVariable, the two link fields are mapped via ADR 0007, and only Last calibrated remains unmapped.

rs-nicof:
Replying to the two findings in @nhanlon2's review summary, both addressed in 5a069db42.

**Publishing the URL of a record the user cannot read.** You are right, and this reverses the decision recorded in ADR 0007 the commit before: the duplicate path copies links nobody re-checked, so write-time READ was not the guarantee the ADR leaned on. Targets are now re-checked at registration time as the instrument's owner (your suggestion, since the owner is stable and keeps the payload independent of which editor clicks), and an entry whose target is deleted, gone, or not readable by the owner is omitted with a WARN. Implemented as a live check rather than the audit snapshot, so it also behaves for records predating auditing.

**A deleted target registered permanently.** Covered by the same re-check: deleted targets no longer qualify, on either provider.

Of the non-blocking notes: the DataCite wire-format test, the both-links-cleared end-to-end case, the wrong-typed field case, the payload-nulling between stages, the two hard-coded DataCite messages (now externalized alongside their B2INST siblings), and the schema 4.7 caveat plus the label-semantics nuance in ADR 0007 are all in the same commit. Not taken: the remaining pinnable prefixes are exercised through the shared set rather than one test per prefix, and the POM pin is answered on its own thread (temporary pairing with the client PR, released 1.1.0 before merge).

nhanlon2 (approved):

PR #1062, opened 2026-08-26
Fix stateless API login marking so Shiro 3's rotation skip works in tests

rs-fraser:
# Fix stateless API login marking so Shiro 3's rotation skip works in tests

## Problem

Downstream CI reports a failure our CI never runs, because the test is `@Ignore`d here pending a chemistry service:

```
StoichiometryControllerMVCIT.updateStoichiometry_whenDocumentIsBeingEdited_returnsConflict:681 Status expected:<200> but was:<409>
```

## Cause

Shiro 3 rotates the session id on every login. RSDEV-1291 skips that for stateless API logins, but marked them with a request attribute read off a `WebDelegatingSubject`. Test wiring uses a non-web `DefaultSecurityManager`, whose subjects are plain `DelegatingSubject`s, so the skip never fired under MockMvc. Every apiKey request then got a new session id, stranding the record edit lock keyed on `(username, sessionId)`: the unlock silently missed, so the next `updateFieldHtml` still saw a live editor and returned 409.

Production is unaffected. Real servlet subjects carry the request, so the skip fires there, and form logins, where rotation is wanted, have `RecordTrackerSessionListener` to purge stale trackers.

## Change

The marker moves onto the thread for the duration of the login, so both security managers read it whatever the subject type.

* `ApiAwareWebSecurityManager` gains `doStatelessLogin(Runnable)`; `isStatelessApiLogin()` no longer inspects the subject.
* `AbstractApiAuthenticator` wraps its `doLogin` in that helper.
* `ApiAwareTestSecurityManager` uses the same check, replacing a comment that was true of Shiro 2 only.

Reading the attribute from `RequestContextHolder` in the test manager would also work, but that manager exists to mirror production, and detecting statelessness by a different route is what let this hide.

## Testing
Verified live: a request carrying both a session cookie and that user's apiKey returns 200, sets no new `JSESSIONID`, and the cookie still works afterwards, and by running the failing test with @Ignore removed.


src/main/java/com/researchspace/auth/ApiAwareWebSecurityManager.java
```diff
@@ -1,44 +1,52 @@
...
+      login.run();
+    } finally {
+      STATELESS_API_LOGIN.remove();
+    }
+  }
+
+  /** True if the current thread is inside a {@link #doStatelessLogin(Runnable)} call. */
+  public static boolean isStatelessApiLogin() {
+    return STATELESS_API_LOGIN.get() != null;
+  }
```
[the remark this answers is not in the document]
rs-fraser: done in https://github.com/rspace-os/rspace-web/pull/1062/commits/4eca74a6664da977e3bdadb9fbdfda1178791dda

nhanlon2 (commented):
left comments

src/main/java/com/researchspace/auth/ApiAwareWebSecurityManager.java
```diff
@@ -1,44 +1,57 @@
...
-  /** Request attribute marking the current login as a stateless API login. */
-  public static final String STATELESS_API_LOGIN =
-      ApiAwareWebSecurityManager.class.getName() + ".STATELESS_API_LOGIN";
+  private static final ThreadLocal<Boolean> STATELESS_API_LOGIN = new ThreadLocal<>();
+
+  /**
+   * Runs a login that must not disturb any session the current thread's subject already has. The
+   * mark only spans the given callback, which should contain just the {@code subject.login} call.
+   */
+  public static void doStatelessLogin(Runnable login) {
```
nhanlon2: This comment is from Claude. `doStatelessLogin` is public and takes any `Runnable`, so how long session-id rotation stays disabled, and which kind of login happens inside it, is entirely up to the caller; only the Javadoc asks for just the `subject.login` call. Nothing abuses that today, but since the rotation being skipped is Shiro's session-fixation fix, it may be worth taking the subject and token as parameters and performing the login here, rejecting any token that is not an `ApiKeyAuthenticationToken`.

nhanlon2 (commented):
This comment is from Claude. Reviewed this PR with four parallel reviewers (layering, i18n, security, generic quality) against the diff from 826eeff1. No exploitable issue found: the thread-local mark cannot outlive the login, verified via the try/finally, the absence of any registered AuthenticationListener, and the deliberate use of a plain ThreadLocal rather than InheritableThreadLocal. Four comments follow; all are hardening or test hygiene rather than defects.

One note that has no line in this diff to sit on: `WopiAuthorisationInterceptor.java:53` performs the same per-request `subject.login(...)` with a `logout()` in `postHandle` and is not wrapped in the new helper, so it may still rotate a browser session that arrives with the request. It is pre-existing and outside this change, and exposure is probably low since WOPI calls are server-to-server, but it looked worth a deliberate decision rather than an accident.

src/main/java/com/researchspace/auth/ApiAwareWebSecurityManager.java
```diff
@@ -1,44 +1,57 @@
...
+    Boolean previous = STATELESS_API_LOGIN.get();
+    STATELESS_API_LOGIN.set(Boolean.TRUE);
+    try {
+      login.run();
+    } finally {
+      if (previous == null) {
+        STATELESS_API_LOGIN.remove();
+      } else {
+        STATELESS_API_LOGIN.set(previous);
```
nhanlon2: This comment is from Claude. This `else` cannot change anything: `TRUE` is the only value ever stored, so when `previous` is non-null it is already `TRUE`. The save and restore collapses to `boolean outermost = !isStatelessApiLogin();` before the `set`, plus `if (outermost) STATELESS_API_LOGIN.remove();` in the `finally`. That also settles line 47, where the flag is typed `Boolean` but read for presence rather than for value.

src/test/java/com/researchspace/testutils/ApiAwareTestSecurityManagerTest.java
```diff
@@ -0,0 +1,54 @@
...
+        () -> securityManager.beforeSuccessfulLogin(subjectWithSession()));
+
+    verifyNoInteractions(session);
+  }
+
+  @Test
+  public void ordinaryLoginStillRotatesSession() {
+    when(session.getAttributeKeys()).thenReturn(List.of());
+
+    securityManager.beforeSuccessfulLogin(subjectWithSession());
```
nhanlon2, line 70: This comment is from Claude. This test creates a real Shiro session: `beforeSuccessfulLogin` stops the mock session and then calls `subject.getSession()`, which reaches the manager's default `DefaultSessionManager` and starts an `ExecutorServiceSessionValidationScheduler` (session validation is enabled by default in Shiro 3.0.0). Nothing calls `securityManager.destroy()`, so each run leaves a live scheduler thread behind in the unit-test fork. An `@AfterEach` calling `securityManager.destroy()` would clear it.

src/main/java/com/researchspace/auth/ApiAwareWebSecurityManager.java
```diff
@@ -1,44 +1,57 @@
...
+ * arrived with. A thread-local mark rather than a request attribute so that it also works for
+ * non-web subjects, such as those created by the {@code DefaultSecurityManager} used in test
+ * wiring.
  */
 public class ApiAwareWebSecurityManager extends DefaultWebSecurityManager {
 
-  /** Request attribute marking the current login as a stateless API login. */
-  public static final String STATELESS_API_LOGIN =
-      ApiAwareWebSecurityManager.class.getName() + ".STATELESS_API_LOGIN";
+  private static final ThreadLocal<Boolean> STATELESS_API_LOGIN = new ThreadLocal<>();
```
nhanlon2: This comment is from Claude. The mark and its accessors live on the production web manager, yet the non-web `ApiAwareTestSecurityManager` and `AbstractApiAuthenticator` both consult them without caring about this class's web role. A small neutral holder in `com.researchspace.auth` would let all three depend on the concept instead of on the production implementation, and would keep this class's Javadoc about session fixation only.

src/main/java/com/researchspace/auth/ApiAwareWebSecurityManager.java
rs-fraser: https://github.com/rspace-os/rspace-web/pull/1062/commits/ec6eb25ce - method now takes `subject` and `token` params rather than `Runnable`

src/main/java/com/researchspace/auth/ApiAwareWebSecurityManager.java
rs-fraser: https://github.com/rspace-os/rspace-web/pull/1062/commits/95defbe17 - tidied up

src/test/java/com/researchspace/testutils/ApiAwareTestSecurityManagerTest.java
rs-fraser: https://github.com/rspace-os/rspace-web/pull/1062/commits/209fd6eb3 - now calling `destroy()`

src/main/java/com/researchspace/auth/ApiAwareWebSecurityManager.java
rs-fraser: https://github.com/rspace-os/rspace-web/pull/1062/commits/58044744d - extracted a holder for the threadlocal

rs-fraser:
> This comment is from Claude. Reviewed this PR with four parallel reviewers (layering, i18n, security, generic quality) against the diff from [826eeff](https://github.com/rspace-os/rspace-web/commit/826eeff1938f052b109d693df778a10a77944392). No exploitable issue found: the thread-local mark cannot outlive the login, verified via the try/finally, the absence of any registered AuthenticationListener, and the deliberate use of a plain ThreadLocal rather than InheritableThreadLocal. Four comments follow; all are hardening or test hygiene rather than defects.
> 
> One note that has no line in this diff to sit on: `WopiAuthorisationInterceptor.java:53` performs the same per-request `subject.login(...)` with a `logout()` in `postHandle` and is not wrapped in the new helper, so it may still rotate a browser session that arrives with the request. It is pre-existing and outside this change, and exposure is probably low since WOPI calls are server-to-server, but it looked worth a deliberate decision rather than an accident.

Addressed all inline comments. Re. WOPI, it's server-to-server so no browser cookie/session to initiate a rotate of, so that code path is unaffected.

nhanlon2 (approved):

PR #1066, opened 2026-08-26
RSDEV-1260: Show B2INST review status back to the user

rs-nicof:
## AWS instance ##
https://rsdev-1260-fetch-b2inst-status-92deedf7-15.researchspace.com/

## Description ##
- Resolves https://researchspace.atlassian.net/browse/RSDEV-1260

### The problem

An instrument PID registered with B2INST is not published by RSpace. RSpace submits the record to a community, and a curator there decides whether to publish it. That decision happens entirely outside RSpace.

RSpace had no way to read the answer back. It wrote the identifier's state twice, once when the draft was registered and once when the record was submitted, and never again. So an instrument PID stayed at "Submitted" forever, no matter what the community had since decided. The four outcomes a curator can produce, accepted, declined, cancelled and expired, could never be shown to the user at all.

That left the user stuck. Nothing in the interface could act on any outcome: Publish was closed off because a review was open, and Delete was closed off because the record was no longer a draft. A submission the community had rejected could not be cleared, so a new PID could not be registered for that instrument. And an accepted submission, the one case where the PID becomes citable, looked exactly like one still waiting.

### What changes

**A Refresh action asks B2INST where the review stands.** It is offered next to the existing identifier actions while the review is open, and pulls the current status from the provider and stores it. Because a curator's decision can arrive at any time, this is the user's way of finding out, rather than waiting for something RSpace can never notice on its own.

Reading the answer takes some care, because accepting a submission publishes the record and removes the draft it was reviewing, so the review itself disappears at the moment of success. RSpace therefore looks at the record as well as the review before deciding what happened, and can distinguish an acceptance from a record that was deleted at the provider, which it reports rather than leaving a stale state in place.

**Every outcome is now actionable.** A declined, cancelled or expired submission can be deleted, which is how the user clears a failed attempt and registers a new PID. A review that exists but was never submitted can still be published, which is how it gets driven forward. An accepted PID can no longer be published again, because it already is published and B2INST has no way to retract it. An open review stays untouched: that decision belongs to the curator. Each state now explains itself and what the user can do about it.

**An accepted PID is treated as published.** "Accepted" is the B2INST counterpart of DataCite's "Findable", the same state reached through a different provider, and it now behaves like it: the identifier is highlighted as published, it reports the citable Handle the provider minted and the landing page that was registered with it, its anonymous public page starts answering, and the state panel explains it the way Findable does.

**Instrument details reach the public page.** When an identifier is set to show custom fields on its public page, an instrument's fields now appear there. Only sample fields did before, so an instrument's public page was missing the very metadata the identifier describes.

**The public page now picks its revision per provider.** B2INST serves the identifier's newest state, because it has no republish operation to push a later change out with. DataCite keeps the snapshot from the moment of publication, so edits made afterwards stay private until the user republishes, which is the existing behaviour for IGSNs and is unchanged.

The new endpoint is documented in the inventory API spec, which is renamed accordingly.

### Notes

- **No database change and no new identifier fields.** The only API addition is the refresh endpoint itself.
- Two hardening fixes came out of reviewing this work, both on paths that already existed. The identifier's publication state is now server-owned rather than accepted from a client, because it is the gate on the unauthenticated public page. And a B2INST record id is now validated and escaped before it goes into a provider URL, so a crafted value cannot redirect RSpace's calls, and the credentials they carry, elsewhere on the provider's host.
- **Refresh is a manual action, deliberately.** There is no scheduled sync in this PR. If a background sweep is wanted later it can reuse the same operation.
- Refreshing a DataCite identifier answers with the state RSpace already holds and makes no provider call. DataCite state only ever changes through RSpace's own publish and retract, so there is nothing to go and read.
- An accepted record does not always come back carrying its minted Handle. Refresh therefore stays available until the Handle is present, so it can be picked up later, rather than leaving an accepted identifier with nothing to resolve.
- Covered by new and updated tests on both sides: the provider reads and their failure handling, the mapping from what the provider reports to what RSpace stores, the endpoint, the delete rules, the public page's revision selection against real persistence, and, on the frontend, a status pulled from the provider travelling all the way through to the rendered identifier.
- Exercised by hand against a B2INST test instance backed by the provider's test environment, alongside the automated coverage.


rs-nicof:
@greptileai

src/main/webapp/ui/src/modules/common/i18n/locales/en-US/inventory.json
```diff
@@ -760,17 +760,20 @@
...
           "findablePidinst": "This PIDINST is Findable. The PIDINST is a citable URL that redirects to the <externalLink href=\"{link}\">RSpace landing page</externalLink>. The metadata is publicly available through the landing page, DataCite Commons and the DataCite APIs.",
-          "pidinstAccepted": "The community accepted this submission. The instrument PID is registered and publicly resolvable.",
-          "pidinstCancelled": "The submission was cancelled before review. The instrument PID remains a draft.",
-          "pidinstDeclined": "The community declined this submission. The instrument PID remains a draft.",
-          "pidinstExpired": "The submission expired before it was reviewed. The instrument PID remains a draft.",
+          "pidinstAccepted": "This PIDINST ID is Accepted. The PIDINST ID is a citable URL that redirects to the <externalLink href=\"{link}\">RSpace landing page</externalLink>. The metadata is publicly available through the landing page, DataCite Commons and the DataCite APIs.",
```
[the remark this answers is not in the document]
rs-nicof: Fixed in a137b069b.

You were right that this text was wrong for the provider. `accepted` is only reachable through B2INST's community review flow (`created` → `submitted` → `accepted`/`declined`/`cancelled`/`expired`), so nothing on that path ever reaches DataCite. The text had been copied verbatim from `findablePidinst`, DataCite references included.

It now reads:

> This PIDINST ID is Accepted. The PIDINST ID is a citable ePIC Handle that redirects to the RSpace landing page. The metadata is publicly available through the landing page and the instrument's B2INST record.

That matches what the refresh actually stores: `publicUrl` gets the ePIC Handle from the record's `pids.epic.identifier`, and `providerUrl` gets `links.self_html`, the B2INST record page.

`resources.d.ts` regenerated via `pnpm run i18n:types`.

src/main/java/com/researchspace/dao/hibernate/DigitalObjectIdentifierDaoHibernate.java
```diff
@@ -52,18 +52,20 @@ public Optional<DigitalObjectIdentifier> getLastPublishedIdentifierByPublicLink(
...
+     * record edits do not). Deliberately the newest revision rather than the first of the
+     * published run: identifier-row changes made while published (the customFieldsOnPublicPage
+     * toggle, a B2INST refresh re-persisting "accepted") must reach the page, and a B2INST
+     * identifier has no republish operation to push them out with (RSDEV-1260).
+     */
+    Object[] newestRow = (Object[]) genericResults.get(genericResults.size() - 1);
+    DigitalObjectIdentifier newestDoi = (DigitalObjectIdentifier) newestRow[0];
+    return DigitalObjectIdentifier.isPublishedState(newestDoi.getState())
+        ? Optional.of(newestDoi)
+        : Optional.empty();
```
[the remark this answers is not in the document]
rs-nicof: Fixed in a137b069b, and following your point up found a real defect next to it.

The query had no `addOrder`, so `genericResults.get(size - 1)` was relying on Envers returning revisions in ascending order, which it does not guarantee. The newest revision is now asked for explicitly:

```java
.add(AuditEntity.id().eq(id))
.addOrder(AuditEntity.revisionNumber().desc())
.setMaxResults(1);
```

That also fetches one row instead of the whole revision history.

Added the persistence test to `InventoryIdentifierApiManagerIT` (`RealTransactionSpringTestBase`, so Envers revisions actually commit). It publishes an identifier, then makes two further committed writes to the identifier row via `doInTransaction`, and asserts:

1. the page reflects the second write (`customFieldsOnPublicPage` true) rather than the values from the moment of publishing;
2. with `accepted` on the newest revision the page stays available and reports that state;
3. with `declined` on the newest revision the page disappears again.

Verified it is a real guard rather than a passing no-op: flipping `.desc()` to `.asc()` makes it fail, and it fails on the very first assertion because the oldest revision is the `draft` one, so `findPublishedItemVersionByPublicLink` returns null and the public page is gone entirely. That is the regression you described. Restored `.desc()` and the IT is green (2/2).

Ran it against the existing dev database rather than `-Denvironment=drop-recreate-db`, to avoid resetting local data.

src/main/java/com/researchspace/service/inventory/impl/InventoryIdentifierApiManagerImpl.java
```diff
@@ -322,6 +323,23 @@ public ApiInventoryRecordInfo retractIdentifier(GlobalIdentifier invRecOid, User
         user, invRec, createUpdateWithRetractedDoi(invRec, user));
   }
 
+  @Override
+  public ApiInventoryRecordInfo refreshIdentifier(GlobalIdentifier invRecOid, User user) {
+    InventoryRecord invRec = invRecRetriever.getInvRecordByGlobalId(invRecOid);
+    if (invRec.getActiveIdentifiers().isEmpty()) {
+      throw new IllegalArgumentException(
+          "record " + invRecOid.toString() + " has no identifier to refresh");
```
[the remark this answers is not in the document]
rs-nicof: Fixed in a137b069b.

Correct, and it is also what `CLAUDE.md` requires for a block being touched. Now:

```java
throw new ApiRuntimeException("errors.inventory.identifier.refreshNoIdentifier");
```

with the message added to `server.inventory.json`, so `JsonMessageSource` resolves it and the advice maps it like every other failure on the refresh path.

Unit test `refreshWithoutIdentifierRaisesLocalizedErrorCode` asserts the error code rather than the rendered string; it was red on the old `IllegalArgumentException` first. Class is green (20/20).

Left the identical hard-coded messages in `deleteAssociatedIdentifier`, `publishIdentifier` and `retractIdentifier` alone: they predate this branch and are outside its scope. Happy to externalize them in a follow-up if you would rather have the whole set consistent.

rs-nicof:
@greptileai

rs-nicof:
## Automated review pass — cycle 1

All three Copilot findings were valid and are fixed in a137b069b. Replied inline on each thread and resolved them.

| Finding | Verdict | Fix |
|---|---|---|
| `pidinstAccepted` help text credits DataCite Commons / DataCite APIs | Valid | `accepted` is reachable only through B2INST community review, so nothing on that path reaches DataCite. Text now points at the ePIC Handle and the B2INST record; `resources.d.ts` regenerated |
| Public-page revision selection untested at persistence level | Valid, and it uncovered a defect | The Envers query had no `addOrder`, so taking the tail of the result list relied on an ordering Envers does not guarantee. Now `.addOrder(revisionNumber().desc()).setMaxResults(1)`. New IT covers the newest-revision rule, `accepted` staying public, and a non-published newest revision hiding the page |
| Hard-coded exception text in `refreshIdentifier` | Valid | Replaced with `ApiRuntimeException("errors.inventory.identifier.refreshNoIdentifier")` and a catalog entry, matching the rest of the refresh path |

**Verification**

- `InventoryIdentifierApiManagerImplUnitTest` — 20/20
- `InventoryIdentifierApiManagerIT` — 2/2 (run against the existing dev DB, not `drop-recreate-db`)
- Mutation check: flipping `.desc()` to `.asc()` makes the new IT fail, so it is a real guard, not a passing no-op. It fails on the first assertion because the oldest revision is the `draft` one, meaning the public page would vanish entirely
- Frontend: `pnpm tsc` clean, Identifiers + RefreshButton suites 41/41, `i18n:lint` shows only four pre-existing warnings in unrelated files
- Spotless clean; all pre-commit hooks passed

**Not changed:** the same hard-coded messages in `deleteAssociatedIdentifier`, `publishIdentifier` and `retractIdentifier`. They predate this branch and are outside its scope — happy to do them in a follow-up.

Greptile did not respond (it has not posted in this repo since 2026-07-29; its own status page reports all components operational, so this looks org-side — a GitHub App installation or seat issue rather than an outage). Proceeded on Copilot alone.

src/main/java/com/researchspace/service/inventory/impl/InventoryIdentifierApiManagerImpl.java
```diff
@@ -680,6 +699,82 @@ private ApiInventoryDOI createUpdateWithPublishedB2instDoi(DigitalObjectIdentifi
...
+  }
+
+  /**
+   * An accepted B2INST PID is the PIDINST equivalent of DataCite's {@code findable}, whose {@code
+   * url} is the identifier's RSpace public landing page, so acceptance fills the same field from
+   * the identifier's own public link (RSDEV-1260).
+   */
+  private void setLandingPageUrl(ApiInventoryDOI refreshUpdate, DigitalObjectIdentifier doi) {
+    InventoryUrls.publicLandingPageUrl(properties.getServerUrl(), doi.getPublicLink())
+        .ifPresent(refreshUpdate::setUrl);
```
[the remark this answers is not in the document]
rs-nicof: Fixed in 750e82dd1. Good catch, this was a real defect and not just a stylistic one.

Confirmed the mismatch: `RspaceToExternalProviderAdapterImpl` registers `registrableTyped.or(() -> publicLandingPageUrl)` filtered by `isResolvableAddress`, so the registered `LandingPage` is the user's institutional address whenever they typed a resolvable one, and may be omitted entirely if neither candidate passes the guard. `setLandingPageUrl` meanwhile rebuilt the RSpace public page unconditionally, so an accepted instrument could report a `url` its own Handle does not resolve to.

Took the first of your two options, reading it back from the record:

```java
private String landingPageOf(B2instDraftRecord record) {
  return record.getMetadata() == null ? null : record.getMetadata().getLandingPage();
}
```

`setLandingPageUrl` now takes that value and prefers it, falling back to the rebuilt public page only when it is blank. Reading back beats persisting the registered value at registration time: B2INST holds what the curator accepted, so no second copy can drift from it.

Kept the fallback for the `"accepted".equals(reviewStatus)` branch, where `getPublishedRecord` returned empty and there is no record to read a `LandingPage` from. Passing `null` there makes that explicit.

Tests: `refreshTakesUrlFromTheRegisteredLandingPageWhenTheRecordCarriesOne` was red before the change, reporting `https://rspace.example.com/public/inventory/...` where the registered address was `https://institution.example.org/instruments/nmr-400`. Added `refreshFallsBackToThePublicPageWhenAcceptanceComesFromTheReviewAlone` for the other branch. 124/124 green across `InventoryIdentifierApiManagerImplUnitTest`, `InventoryIdentifierApiManagerIT`, `B2instConnectorImplTest`, `DigitalObjectIdentifierTest`, `InventoryIdentifiersApiControllerTest` and `RspaceToExternalProviderAdapterImplTest`.

src/main/webapp/ui/src/modules/common/i18n/locales/en-US/inventory.json
```diff
@@ -760,17 +760,20 @@
...
           "findablePidinst": "This PIDINST is Findable. The PIDINST is a citable URL that redirects to the <externalLink href=\"{link}\">RSpace landing page</externalLink>. The metadata is publicly available through the landing page, DataCite Commons and the DataCite APIs.",
-          "pidinstAccepted": "The community accepted this submission. The instrument PID is registered and publicly resolvable.",
-          "pidinstCancelled": "The submission was cancelled before review. The instrument PID remains a draft.",
-          "pidinstDeclined": "The community declined this submission. The instrument PID remains a draft.",
-          "pidinstExpired": "The submission expired before it was reviewed. The instrument PID remains a draft.",
+          "pidinstAccepted": "This PIDINST ID is Accepted. The PIDINST ID is a citable ePIC Handle that redirects to the <externalLink href=\"{link}\">RSpace landing page</externalLink>. The metadata is publicly available through the landing page and the instrument's B2INST record.",
```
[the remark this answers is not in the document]
rs-nicof: Fixed in 750e82dd1. Right, and it was my previous revision that introduced the assumption, so thanks for catching it in the same pass.

Now:

> This PIDINST ID is Accepted. The PIDINST ID is a citable ePIC Handle that redirects to the registered landing page. The metadata is publicly available through that page and the instrument's B2INST record.

"registered landing page" is accurate whichever path applied, since that is exactly what B2INST holds and what the curator baked into the Handle. It also now matches what the link actually points at: the sibling finding on `setLandingPageUrl` is fixed in the same commit, so `{link}` carries the registered `metadata.LandingPage` rather than a rebuilt RSpace URL.

`resources.d.ts` regenerated. `pnpm tsc` clean, Identifiers + RefreshButton suites 41/41.

rs-nicof:
@greptileai

rs-nicof (commented):
Parallel review of the RSDEV-1260 refresh work across layering, i18n, security, persistence, frontend and general quality. 11 findings: 1 critical (a change to what the public landing page serves, which also affects existing DataCite/IGSN identifiers) and 10 high. Everything else the reviewers raised was either verified as deliberate or dropped as unconfirmed. The Swagger rename checks out clean, with no surviving 2_25_0 references.

src/main/java/com/researchspace/dao/hibernate/DigitalObjectIdentifierDaoHibernate.java
```diff
@@ -45,25 +45,31 @@ public Optional<DigitalObjectIdentifier> getLastPublishedIdentifierByPublicLink(
...
-    }
-    return lastPublishedDoi == null ? Optional.empty() : Optional.of(lastPublishedDoi);
+    /*
+     * The page serves the identifier's most recent revision, and only while that revision is in a
+     * published state. Envers resolves the linked record as of that revision, so the page still
+     * shows the state from the moment of the last write that touched the identifier row (ordinary
+     * record edits do not). Deliberately the newest revision rather than the first of the
+     * published run: identifier-row changes made while published (the customFieldsOnPublicPage
+     * toggle, a B2INST refresh re-persisting "accepted") must reach the page, and a B2INST
+     * identifier has no republish operation to push them out with (RSDEV-1260).
```
rs-nicof: This changes what the public page serves. It used to be pinned to the publication-time snapshot (the oldest revision of the trailing published run); now it follows the newest identifier revision, and Envers resolves the linked record as of that revision. So once anything writes the identifier row, item edits made after publication become world-visible with no deliberate republish.

The `customFieldsOnPublicPage` toggle is one such write and is not B2INST-specific, so this also changes behaviour for existing DataCite/IGSN identifiers, which do have a republish action. The comment explains the B2INST motivation, but worth confirming the DataCite side effect is intended too.

Nothing currently pins this either way: the existing `InventoryIdentifierApiManagerIT` passes under both the old and new rule, because an ordinary item edit creates no identifier revision. Smallest guard is an IT that publishes, edits the item, toggles the checkbox, then asserts what the public link returns.

src/main/java/com/researchspace/dao/hibernate/DigitalObjectIdentifierDaoHibernate.java
```diff
@@ -45,25 +45,31 @@ public Optional<DigitalObjectIdentifier> getLastPublishedIdentifierByPublicLink(
...
-        break; // we've hit doi that's not published, stop here
-      }
-    }
-    return lastPublishedDoi == null ? Optional.empty() : Optional.of(lastPublishedDoi);
+    /*
+     * The page serves the identifier's most recent revision, and only while that revision is in a
+     * published state. Envers resolves the linked record as of that revision, so the page still
+     * shows the state from the moment of the last write that touched the identifier row (ordinary
+     * record edits do not). Deliberately the newest revision rather than the first of the
+     * published run: identifier-row changes made while published (the customFieldsOnPublicPage
```
rs-nicof: The Envers query above has no `addOrder`, so treating the last element as the newest revision depends on unspecified database row order. The repo's own convention is explicit about this: `AuditDaoHibernateEnversImpl` uses `.addOrder(AuditEntity.revisionNumber().desc())` in four places.

Adding `.addOrder(AuditEntity.revisionNumber().desc()).setMaxResults(1)` and taking element 0 also avoids loading an identifier's whole revision history to use one row.

src/main/java/com/researchspace/model/inventory/DigitalObjectIdentifier.java
```diff
@@ -104,6 +104,15 @@ public DigitalObjectIdentifier(String identifier, String title) {
     this(identifier, title, null);
   }
 
+  /**
+   * True when the given identifier state means the metadata is published: DataCite's {@code
+   * findable}, or B2INST's {@code accepted} (the community-review outcome that publishes the
+   * record). Gates the public landing page, so both providers' published states must be here.
+   */
+  public static boolean isPublishedState(String state) {
```
rs-nicof, line 112: This is now the only gate on the unauthenticated landing page, but the column it reads is client-writable. `ApiInventoryDOI.state` has no `Access.READ_ONLY`, unlike `url`, `publicUrl` and `providerUrl` right beside it (whose Javadoc explains exactly why they need it), and `applyChangesToDatabaseDOI` copies it straight onto the entity from a record update.

So anyone with edit rights on the record can send one ordinary update setting `state` to `"accepted"` and make the public page live, with no B2INST review and no provider registration. Pre-existing for `findable`, but marking `state` READ_ONLY belongs with the change that makes this the gate.

src/main/java/com/researchspace/webapp/integrations/b2inst/B2instConnectorImpl.java
```diff
@@ -248,6 +249,43 @@ private HttpEntity<String> emptyJsonBody() {
     return new HttpEntity<>("{}", headers);
   }
 
+  @Override
+  public Optional<B2instRequestResponse> getReviewOf(String rid) {
+    try {
+      return Optional.ofNullable(
+          restTemplate.getForObject(
+              apiBase() + "/records/" + rid + "/draft/review", B2instRequestResponse.class));
```
rs-nicof: `rid` is concatenated into the path with no encoding, and it comes from `ApiInventoryDOI.doi`, which is also missing `READ_ONLY`. A user who can edit the instrument could store a crafted identifier value and make the server issue GETs carrying the B2INST bearer token to other paths on the provider host, with parts of the answer reflected back through `state`, `providerUrl` and the failure message.

Fix at either end: `UriComponentsBuilder.fromUriString(apiBase()).pathSegment("records", rid, "draft", "review")`, which escapes `/` and `..`, or mark `doi` READ_ONLY like its URL siblings. Same shape applies to the two reads below.

src/main/java/com/researchspace/service/inventory/impl/InventoryIdentifierApiManagerImpl.java
```diff
@@ -322,6 +323,24 @@ public ApiInventoryRecordInfo retractIdentifier(GlobalIdentifier invRecOid, User
...
+      // Localized error code rather than developer text: refresh is reachable straight from the
+      // public API, so the whole path resolves errors.inventory.identifier.* keys consistently.
+      throw new ApiRuntimeException("errors.inventory.identifier.refreshNoIdentifier");
+    }
+    DigitalObjectIdentifier doi = invRec.getActiveIdentifiers().get(0);
+    if (!isB2inst(doi.getType())) {
+      // DataCite state changes only through RSpace's own publish/retract calls, so the stored
+      // state is already current and there is no provider read to make.
+      return ApiInventoryRecordInfo.fromInventoryRecordToFullApiRecord(invRec);
+    }
```
rs-nicof, line 339: Routing a status read through `updateInventoryRecordWithDoiUpdate` gives Refresh the whole write path. `updateApiInstrument` calls `assertUserCanEditInstrument` and `lockItemForEdit`, so Refresh fails outright while a colleague has the instrument open for editing; and when the status has changed, `saveDbInstrumentUpdate` re-stamps `modificationDate` and `modifiedBy` and bumps the version, so checking a status makes the record look as though the person who clicked Refresh edited it.

The transaction is also open across up to three sequential B2INST calls (30s read timeout each), behind a button users can click repeatedly. `B2instConnectorImpl`'s own class comment flags this pool-exhaustion risk for the existing single-call operations. Reading the provider before opening the write transaction addresses both.

src/main/java/com/researchspace/service/inventory/impl/InventoryIdentifierApiManagerImpl.java
```diff
@@ -680,6 +699,98 @@ private ApiInventoryDOI createUpdateWithPublishedB2instDoi(DigitalObjectIdentifi
...
+      Optional<B2instDraftRecord> published = b2instConnector.getPublishedRecord(rid);
+      if (published.isPresent()) {
+        refreshUpdate.setState("accepted");
+        refreshUpdate.setPublicUrl(epicPidOf(published.get()));
+        if (published.get().getLinks() != null) {
+          refreshUpdate.setProviderUrl(published.get().getLinks().getSelfHtml());
+        }
+        setLandingPageUrl(refreshUpdate, doi, landingPageOf(published.get()));
+        return refreshUpdate;
+      }
```
rs-nicof, line 736: This branch (the review says `accepted`, but `/records/{rid}` is not there yet) sets the state to `accepted`, which now opens the public landing page via the new `isPublishedState`, while leaving `publicUrl` (the minted Handle PID) and `providerUrl` unset.

So the page can go live describing a PID the user cannot resolve. Either hold off publishing on this branch, or fill the PID before returning. Also currently untested.

src/main/webapp/ui/src/modules/common/i18n/locales/en-US/inventory.json
```diff
@@ -760,17 +760,20 @@
...
           "findablePidinst": "This PIDINST is Findable. The PIDINST is a citable URL that redirects to the <externalLink href=\"{link}\">RSpace landing page</externalLink>. The metadata is publicly available through the landing page, DataCite Commons and the DataCite APIs.",
-          "pidinstAccepted": "The community accepted this submission. The instrument PID is registered and publicly resolvable.",
-          "pidinstCancelled": "The submission was cancelled before review. The instrument PID remains a draft.",
-          "pidinstDeclined": "The community declined this submission. The instrument PID remains a draft.",
-          "pidinstExpired": "The submission expired before it was reviewed. The instrument PID remains a draft.",
+          "pidinstAccepted": "This PIDINST ID is Accepted. The PIDINST ID is a citable ePIC Handle that redirects to the <externalLink href=\"{link}\">registered landing page</externalLink>. The metadata is publicly available through that page and the instrument's B2INST record.",
```
rs-nicof, line 772: `accepted` is a B2INST-only state, and B2INST mints ePIC Handle PIDs (as the `hdl.handle.net/21.T11975/...` handling in this PR shows), so nothing about this record is in DataCite Commons or the DataCite APIs.

The previous wording ("The instrument PID is registered and publicly resolvable") was accurate; this copies the `findable` text including its DataCite claim. A user will go looking in DataCite Commons and find nothing, and translators will carry the wrong claim into every locale.

src/main/webapp/ui/src/stores/models/IdentifierModel.tsx
```diff
@@ -645,6 +646,45 @@ export default class IdentifierModel implements Identifier {
...
+      const { state, url, publicUrl, providerUrl } = response.data;
+      runInAction(() => {
+        this.state = state;
+        this.url = url;
+        this.publicUrl = publicUrl;
+        this.providerUrl = providerUrl;
+      });
+      addAlert(
+        mkAlert({
+          message: i18n.t("inventory:identifierModel.alerts.refreshed", { state }),
```
rs-nicof: This interpolates the raw provider token into a translated sentence, so every locale shows `The identifier status is now "accepted"` with an English provider word inside the quotes, right next to a table cell showing the properly translated label.

`fields.identifiers.list.stateLabels.*` already holds all six values and is what `identifierStateLabel` renders. Passing the label instead of the raw `state` keeps the two consistent.

src/main/webapp/ui/src/Inventory/components/Fields/Identifiers/RefreshButton.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  if (identifier.state !== "submitted") return null;
+
+  return (
+    <CustomTooltip title={t("fields.identifiers.list.tooltips.refresh")}>
+      <Button
+        color="callToAction"
+        variant="outlined"
+        size="small"
+        disabled={refreshing || disabled}
```
rs-nicof, line 46: Two things here.

The button only greys out, with no spinner, though refresh makes up to three B2INST calls at a 30s read timeout each. `SubmitSpinnerButton` (which the sibling `PublishButton` uses) or MUI's `loading` prop covers it, and `IdentifierModel.refresh`'s docstring already claims a spinner exists.

The `try`/`finally` below also has no `catch`, and the two guards at the top of `refresh` throw before its own `try`, so those failures escape the voided IIFE as an unhandled rejection with no alert at all. The user sees the button flicker and nothing else.

src/test/java/com/researchspace/webapp/integrations/b2inst/B2instConnectorDummy.java
```diff
@@ -46,6 +47,21 @@ public B2instRequestResponse retractDoi(String rid) {
...
+    return Optional.empty();
+  }
+
+  @Override
+  public Optional<B2instDraftRecord> getPublishedRecord(String rid) {
+    return Optional.empty();
+  }
+
+  @Override
+  public Optional<B2instDraftRecord> getDraftRecord(String rid) {
```
rs-nicof, line 70: All three new reads return empty unconditionally, but `registerDoi` above creates a draft with `DUMMY_RID`, so the double contradicts its own state. A test that registers through the dummy and then refreshes lands on the `b2instRecordGone` error path rather than the truthful "still a draft" outcome, so a future test could assert the wrong behaviour.

Returning a `B2instDraftRecord` with `DUMMY_RID` when `rid` matches would keep it honest.

src/main/webapp/ui/src/Inventory/components/Fields/Identifiers/__tests__/RefreshButton.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { mockIGSNIdentifier } from "./mocking";
+import "@/__tests__/__mocks__/matchMedia";
+import { ThemeProvider } from "@mui/material/styles";
+import materialTheme from "../../../../../theme";
+
+const submittedPidinst = (overrides?: Partial<Identifier>): Identifier => ({
+  ...mockIGSNIdentifier("instrument"),
+  doiType: "PIDINST_B2INST",
+  state: "submitted",
+  refresh: vi.fn().mockResolvedValue(undefined),
```
rs-nicof, line 24: Stubbing `refresh` means nothing proves a refreshed status reaches the UI. The suite covers "the button calls refresh" and "the model applies the response" separately, with nothing joining them: no assertion that the state cell re-renders with the new label, that Refresh disappears once the review closes, or that Publish/Delete gating flips.

That is the PR's headline behaviour, so it is the one thing currently unverified. Note `mockIGSNIdentifier` returns a plain object rather than a MobX model, so this needs a real `IdentifierModel` attached to the record, as `IdentifierModel.test.ts` already builds.

rs-nicof:
Correction on the review above: it was computed against `7e30a922`, but two commits (`a137b069`, `750e82dd`) landed while the reviewers were running, so it posted against a head it had not read. Three of the comments are already out of date:

- **DAO ordering** (`DigitalObjectIdentifierDaoHibernate`): already fixed in `750e82dd` with `.addOrder(AuditEntity.revisionNumber().desc()).setMaxResults(1)` and `get(0)`. Please disregard; the comment also drifted onto a line inside the Javadoc.
- **`pidinstAccepted` wording**: already fixed. The current text says "citable ePIC Handle ... registered landing page ... that page and the instrument's B2INST record", with no DataCite claim. Please disregard.
- **Public landing page (the critical one)**: the "nothing tests it" half is wrong, `publicPageServesNewestIdentifierRevisionIncludingAcceptedState` now covers it. The substantive question still stands though: the newest-revision rule also applies to DataCite/IGSN identifiers via `customFieldsOnPublicPage`, so post-publication item edits become publicly visible there too, and those do have a republish action. Worth confirming that is intended for DataCite, not just B2INST.

Still standing, in files those two commits did not touch: the client-writable `state` gate, the unencoded `rid` path building, the edit-lock and `modifiedBy` side effects on Refresh, the raw provider token in the translated alert, the missing spinner and absent `catch`, the self-contradicting `B2instConnectorDummy`, and the stubbed-out refresh test.

The `accepted`-but-record-absent finding moved with the `setLandingPageUrl` refactor but still applies: that branch passes `null`, so `publicUrl` (the Handle PID) and `providerUrl` stay unset while the state opens the public page.

src/main/java/com/researchspace/dao/hibernate/DigitalObjectIdentifierDaoHibernate.java
rs-nicof: Confirmed and fixed in 341228045. This was the important one.

You were right that the side effect was not intended. Nico split it while this review was landing: B2INST follows the newest revision (it has no republish to push a change out with), DataCite keeps the publication-time snapshot. I finished that split, because an ascending order with `setMaxResults(1)` does not express "oldest of the trailing published run" - it returns the oldest revision *overall*, which is the `draft` the identifier started as. `isPublishedState("draft")` is false, so every published IGSN lost its page: your own `retrieveLastPublishedVersionOfItemWithPublicLink` failed on it.

The DataCite branch therefore keeps the walk back from the newest revision, stopping at the first unpublished one, which is what actually picks that revision out. One descending query now feeds both rules, and the "published right now" check is applied before either, so a retracted identifier still takes its page with it.

A null type lands on the DataCite branch rather than leaving the query null and throwing - identifiers persisted before the type column existed load with a null type, exactly as `settingTypeFor` documents.

New IT `publicPagePicksItsRevisionByProvider` pins both branches: it publishes, makes a committed write to the identifier row, asserts the IGSN page still shows the publication-time value, then reads the same history as B2INST and asserts the newest value is served, then asserts a non-published newest revision hides the page. 136/136 green including your existing IT.

src/main/java/com/researchspace/dao/hibernate/DigitalObjectIdentifierDaoHibernate.java
rs-nicof: Already fixed, in a137b069b earlier in this run - we found the same thing independently, and `AuditDaoHibernateEnversImpl` is the right precedent.

The query now orders `revisionNumber().desc()` explicitly. I did not keep `setMaxResults(1)` in the end: completing the provider split above needs more than the newest row on the DataCite side, so it reads the ordered history and walks it. The wasteful part you flagged is gone for the B2INST branch, which returns `revisions.get(0)` directly.

Verified as a real guard rather than a no-op: flipping the order to `asc()` fails the new IT on its first assertion.

src/main/java/com/researchspace/model/inventory/DigitalObjectIdentifier.java
rs-nicof: Fixed in 341228045. Agreed this belongs with the change that makes `state` the gate.

`state` is now `@JsonProperty(value = "state", access = JsonProperty.Access.READ_ONLY)`, with a Javadoc pointing at the same reasoning as `url`, `publicUrl` and `providerUrl`. Confirmed the path you describe: `ApiInventoryRecordInfo.applyChangesToSampleIdentifiers` reaches `applyChangesToDatabaseDOI`, which copies `state` onto the existing entity, so an ordinary record update carrying `"state": "accepted"` did publish the page with no registration behind it.

`READ_ONLY` only affects deserialization, so register, publish, retract and refresh are untouched - they set state in Java. The 136 backend tests confirm that, including the register/publish/retract IT.

Left `doi` writable for now: the encoding fix on the other thread closes the vector at the point of use, which also covers any other source of a crafted id. Say the word if you would rather have it locked down as well.

src/main/java/com/researchspace/webapp/integrations/b2inst/B2instConnectorImpl.java
rs-nicof: Fixed in 341228045, and applied wider than the three new reads.

Confirmed the chain: `rid` is `doi.getIdentifier()` in `createUpdateWithRefreshedB2instDoi`, and `applyChangesToDatabaseDOI` writes `identifier` from the client-supplied `doi`, so this was reachable.

All five record calls now go through one helper. I did not take the encoding on its own, because it is not sufficient: `pathSegment("..")` escapes nothing (`.` is unreserved), so a bare `..` would still be a traversal segment a normaliser could collapse. So the helper checks the id against `[A-Za-z0-9_-]{1,64}` first - both observed forms fit, `k2j9p-7yh21` and `abcde-12345` - and then escapes it with `pathSegment`, which stops `/` injection. A rejected id throws rather than being cleaned up, since RSpace did not mint it and there is no correct request to send.

Routed the pre-existing `deleteDoi` and the review PUT through it too. They take the same untrusted id, so fixing only the new reads would have left the hole open through delete and publish.

`recordCallsRefuseARecordIdRSpaceCouldNotHaveMinted` covers 9 crafted ids across all four calls and asserts via `MockRestServiceServer` that nothing is sent. The existing tests still pin the URLs built for a legitimate id, which is what proves the encoding did not change them. 39/39 in that class.

src/main/java/com/researchspace/service/inventory/impl/InventoryIdentifierApiManagerImpl.java
rs-nicof: Fixed in 341228045.

Agreed, and of your two options only "hold off" is actually available: this branch is reached precisely because `/records/{rid}` gave nothing, so there is no record to take the Handle from. Filling the PID is not possible here.

It now throws `errors.inventory.identifier.b2instAcceptedRecordUnavailable` and keeps the stored state, rather than storing `accepted` and opening the page for a PID the user cannot resolve. Framed as retryable in the message, since acceptance publishes the record: a review saying accepted while the record is unreadable is normally transient, so trying again shortly is the right advice.

Now tested: `refreshRefusesToPublishAnAcceptedReviewWhoseRecordIsNotAvailable`.

Note this replaced a test I had added an hour earlier asserting the opposite - it fell back to the RSpace public page on this branch. Your reading is the right one, so that test is gone.

src/main/webapp/ui/src/modules/common/i18n/locales/en-US/inventory.json
rs-nicof: Already fixed, in a137b069b and refined in 750e82dd1 - Copilot raised the same thing and we had the same diagnosis, including that it came from copying the `findable` text.

It now reads:

> This PIDINST ID is Accepted. The PIDINST ID is a citable ePIC Handle that redirects to the registered landing page. The metadata is publicly available through that page and the instrument's B2INST record.

"registered landing page" rather than "RSpace landing page" because registration prefers a user-typed institutional address, so naming RSpace would be wrong on that path. The `{link}` it wraps now carries the registered `metadata.LandingPage` read back from B2INST, so the sentence and the link agree.

`resources.d.ts` regenerated.

src/main/webapp/ui/src/stores/models/IdentifierModel.tsx
rs-nicof: Fixed in 341228045, and taken one step further than passing the label in.

Rather than translating the state at the call site, `identifierStateLabel` moved out of the `Identifiers.tsx` closure and into `stores/definitions/Identifier.ts`, next to `isPublishedState`. Both the table cell and the alert now call the same function, so they cannot drift. Duplicating the nine-arm mapping into the model would have created exactly the kind of divergence `PidinstFields` warns about for the landing-page rule.

Arms stay spelled out rather than built from the state, so the typed catalog and the extractor still see every key, and the `String(state)` catch-all is preserved for a provider status RSpace does not recognise.

src/main/webapp/ui/src/Inventory/components/Fields/Identifiers/RefreshButton.tsx
rs-nicof: Both fixed in 341228045.

Spinner: used MUI's `loading` prop rather than `SubmitSpinnerButton`. The latter hardcodes `variant="contained"`, so adopting it would have turned Refresh into a filled button beside Publish; `loading={refreshing}` keeps the outlined secondary look and still shows the spinner. MUI is 9.3.1 here and the codebase already uses the prop elsewhere.

The missing `catch`: added, alerting with `identifierModel.alerts.refreshFailed` and `getErrorMessage`. You were right about the mechanism - `refresh()` handles provider failures itself, but its two preconditions throw before its own `try`, so as an async rejection out of a voided IIFE they were swallowed entirely.

The docstring claim you spotted is now true rather than aspirational.

src/test/java/com/researchspace/webapp/integrations/b2inst/B2instConnectorDummy.java
rs-nicof: Fixed in 341228045.

`getDraftRecord` now returns a `B2instDraftRecord` with `DUMMY_RID` and `DUMMY_SELF_HTML` when the id matches, and empty otherwise, so the double agrees with the draft `registerDoi` created. Left `getPublishedRecord` empty on purpose and said so in a Javadoc: `publishDoi` here does not move the record on, so nothing this double created is published.

This mattered sooner than "a future test could" - it is what a register-then-refresh test needs to land on the truthful draft outcome rather than the record-gone path.

src/main/webapp/ui/src/Inventory/components/Fields/Identifiers/__tests__/RefreshButton.test.tsx
rs-nicof: Fixed in 341228045, and your note about `mockIGSNIdentifier` being a plain object saved me the detour.

Added a second describe block that renders `IdentifiersList` with a real `IdentifierModel`, built from `mockIGSNAttrs()` as `PIDINST_B2INST`/`submitted`, with only the HTTP `post` faked - the same injection `IdentifierModel.test.ts` uses. So the model, the observer re-render and the state-keyed gating are all real.

It asserts the three things you listed. Before the click: Submitted rendered, Refresh offered, Publish held back by `awaitingReview`. After an accepted response: the state cell re-renders as Accepted, Refresh is gone from the DOM, and Publish is still disabled - now via `publishedPidinst` instead. A second test does the same for a declined review.

5/5 in that file, 50/50 across the identifier suites.

src/main/java/com/researchspace/service/inventory/impl/InventoryIdentifierApiManagerImpl.java
rs-nicof: Confirmed as real, and the one finding from your review I have deliberately not fixed in this pass.

Verified each part: `updateApiInstrument` does call `assertUserCanEditInstrument` and `lockItemForEdit`, so Refresh fails outright while a colleague holds the lock; `saveDbInstrumentUpdate` does re-stamp `modificationDate`/`modifiedBy` and bump the version, so checking a status makes the record look edited; and the transaction does stay open across up to three sequential 30s calls, which is the pool-exhaustion risk `B2instConnectorImpl`'s own class comment already flags.

Not fixed because the fix is a transaction-boundary change, not a local one. `refreshIdentifier` sits on a `*Manager` bean, so the AOP pointcut in `applicationContext-service.xml` makes the whole method transactional; reading the provider first means splitting the read out of that boundary. How to split it is an architectural call I did not want to make unilaterally at the end of an automated pass, and getting it wrong is worse than the current behaviour.

Proposed shape, if you want it: do the three reads in a non-transactional method, and open the write only when the state actually changed - which also means an unchanged status stops writing at all, so no lock is taken and no modification stamp moves in the common case. Happy to implement that as its own change with the IT to match, or leave it for a follow-up ticket. Your call.

The spurious `modifiedBy` stamp is the part I would prioritise: it is the one users will notice and cannot explain.

rs-nicof:
## Automated review pass — cycle 3

Your parallel review: 11 findings, **9 fixed** in 341228045, 2 already fixed earlier in this run, **1 deliberately left** for you to decide.

### Fixed
| # | Finding | Note |
|---|---|---|
| 1 | Public page followed the newest revision for every provider | Completed the B2INST/DataCite split. The DataCite branch needs the walk back from the newest revision: `asc()` + `setMaxResults(1)` returns the *draft*, which took the page away from every published IGSN and failed your own existing IT. Null type now lands on the DataCite branch instead of throwing. New IT pins both branches |
| 3 | `state` client-writable while gating the public page | `READ_ONLY`. Confirmed a record update carrying `"state": "accepted"` did publish the page with no registration behind it |
| 4 | `rid` concatenated into the B2INST path | Validated **and** escaped through one helper. Encoding alone is insufficient — a bare `..` escapes to nothing — so the id is checked against `[A-Za-z0-9_-]{1,64}` first. Applied to the pre-existing delete and review calls too, which share the flaw |
| 6 | Accepted review with no readable record opened the page | Now refuses and keeps the stored state. Only "hold off" was available: that branch is reached *because* the record gave nothing |
| 8 | Raw provider token in a translated sentence | `identifierStateLabel` moved to `definitions/Identifier.ts`; table and alert now share one function rather than duplicating a nine-arm map |
| 9 | No spinner, no `catch` | MUI `loading` (kept `variant="outlined"`; `SubmitSpinnerButton` would have made it a filled button). `catch` added — the two preconditions throw before `refresh()`'s own `try` |
| 10 | Dummy contradicted its own state | `getDraftRecord` returns the draft it created; `getPublishedRecord` stays empty on purpose, documented |
| 11 | Headline behaviour unverified | Real `IdentifierModel` driven through `IdentifiersList` with only `post` faked; asserts state re-render, Refresh retiring, Publish gating flipping |
| 2, 7 | Envers ordering; DataCite wording | Already fixed in a137b069b / 750e82dd1 — same diagnosis independently |

### Verification
- Backend **136/136**: unit, connector, DAO, controller, and both ITs — including `retrieveLastPublishedVersionOfItemWithPublicLink`, which the in-flight DAO edit had broken
- Frontend **50/50** across the identifier suites; `tsc`, Biome and `i18n:check` clean
- New IT verified as a real guard: flipping the Envers order fails it
- ITs run against the existing dev DB, not `drop-recreate-db`

Threads left open rather than resolved, since they are yours to judge. The DAO file is included in the commit as you asked.

rs-nicof:
@greptileai

src/main/java/com/researchspace/model/inventory/DigitalObjectIdentifier.java
rs-nicof: Yes, lock `doi` down too. The encoding fix does not cover this one, because the risk here is substitution rather than traversal.

`applyChangesToDatabaseDOI` still writes `identifier` from the client-supplied `doi` (line 383), and both delete paths use that stored value with the deployment's provider credentials: `deleteFromDatacite` calls `dataCiteConnector.deleteDoi(doi.getIdentifier(), ...)` and `deleteFromB2inst` calls `b2instConnector.deleteDoi(doi.getIdentifier())`. A valid-looking substitute id passes `VALID_RID` cleanly, so a user with edit rights on their own record can point it at an identifier they do not own and have RSpace ask the provider to delete that one.

Refresh has a milder version of the same shape: it would read the other record's status, landing page and Handle and store them on the caller's identifier.

`state` and `doi` are the two fields the provider owns rather than the client, so READ_ONLY on both keeps the rule simple and closes the delete vector at the source.

src/main/java/com/researchspace/service/inventory/impl/InventoryIdentifierApiManagerImpl.java
rs-nicof: Agreed on not making the boundary change unilaterally at the end of a pass. My suggestion is to split your proposal in two and take only the cheap half here.

The cheap half is the early return: compare the provider's answer against the stored state and skip `updateInventoryRecordWithDoiUpdate` entirely when nothing changed. That is local to `refreshIdentifier`, touches no transaction boundary, and removes both the lock conflict and the `modifiedBy` stamp for every no-op refresh. Since that is the common case, it takes out the part users would actually notice, which is the part you said you would prioritise.

The reads-outside-the-boundary split is the part worth its own ticket with an IT, and I would not want it squeezed in here either.

That ordering also leaves the remaining exposure narrow: a refresh that genuinely changed the status still takes the lock and moves the stamp, which is at least defensible, because something really did change.

src/main/webapp/ui/src/Inventory/components/Fields/Identifiers/RefreshButton.tsx
```diff
@@ -0,0 +1,52 @@
...
+ * Pulls the identifier's current review status from the provider (RSDEV-1260). Only rendered
+ * while the B2INST community review is open ("submitted"), when the outcome is decided outside
+ * RSpace and this is the one useful action.
+ */
+function RefreshButton({ identifier, disabled }: RefreshButtonArgs): React.ReactNode {
+  const [refreshing, setRefreshing] = React.useState(false);
+  const { t } = useTranslation("inventory");
+  const { uiStore } = useStores();
+
+  if (identifier.state !== "submitted") return null;
```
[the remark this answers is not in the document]
rs-nicof: Fixed in 2bd5642c1 — the conclusion is right, though not by the route described.

The premise is out of date: `createUpdateWithRefreshedB2instDoi` no longer persists acceptance when `getPublishedRecord` is empty. That path was removed earlier in this run and now throws `errors.inventory.identifier.b2instAcceptedRecordUnavailable`, precisely because it opened the public page for a PID nobody could resolve.

But the outcome you predicted is still reachable by another route. `epicPidOf` digs the Handle out of the published record's loosely typed `pids` block and returns null when it is not there, so `accepted` with a null `publicUrl` can still be persisted from the published-record branch. Refresh was hidden for every state except `submitted`, so nothing could pick the Handle up afterwards.

Refresh now stays offered while an accepted identifier has no `publicUrl`, and retires once it has one:

```ts
const reviewStillOpen = identifier.state === "submitted";
const acceptedWithoutHandle = identifier.state === "accepted" && !identifier.publicUrl;
if (!reviewStillOpen && !acceptedWithoutHandle) return null;
```

Two tests added for the two halves, and the existing "renders nothing unless the state is submitted" case dropped `accepted` from its list accordingly. The end-to-end test is unaffected: its accepted response carries a Handle, so Refresh correctly retires there. 7/7 in that file, 52/52 across the identifier suites.

src/main/java/com/researchspace/webapp/integrations/b2inst/B2instConnectorImpl.java
```diff
@@ -263,14 +338,7 @@ private HttpEntity<String> emptyJsonBody() {
...
   private B2instRequestResponse openReviewOf(String rid) {
-    try {
-      B2instRequestResponse review =
-          restTemplate.getForObject(
-              apiBase() + "/records/" + rid + "/draft/review", B2instRequestResponse.class);
-      return review != null && Boolean.TRUE.equals(review.getIsOpen()) ? review : null;
-    } catch (HttpClientErrorException.NotFound e) {
-      return null; // no review yet: the ordinary first-publish path
-    }
+    return getReviewOf(rid).filter(review -> Boolean.TRUE.equals(review.getIsOpen())).orElse(null);
```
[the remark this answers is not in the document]
rs-nicof: Not a defect — but it pointed at the one untested path, so 2bd5642c1 pins it.

The short-circuit does not catch a `created` review. `openReviewOf` filters on `is_open`:

```java
return getReviewOf(rid).filter(review -> Boolean.TRUE.equals(review.getIsOpen())).orElse(null);
```

and a review that exists but was never submitted reports `is_open: false`, not true — verified against b2inst-test.gwdg.de in July 2026 and recorded in that helper's Javadoc. So a `created` review returns null from the helper, falls through to the fresh PUT, and that PUT is what returns `links.actions.submit`, which is then posted. Exactly the handling you describe is already what happens; only an already-`submitted` review (`is_open: true`) short-circuits, which is the idempotency this exists for.

What was genuinely missing was coverage. `publishDoiCreatesAFreshReviewWhenThePreviousOneIsClosed` covers a *declined* review, and nothing covered the unsubmitted one, which is the state the UI deliberately keeps Publish enabled for. Added `publishDoiResubmitsAReviewThatWasCreatedButNeverSubmitted`: GET returns `status: created, is_open: false`, then asserts the PUT and the submit POST both go out and the result is `submitted`. Passes unchanged, which is the evidence the behaviour was already correct. 40/40 in that class.

rs-nicof:
## Final report — 4 cycles, converged

**Stop reason:** cycle cap reached, and the last cycle raised nothing that needed a behaviour change. HEAD `2bd5642c1`.

### Cycle 1 — Copilot (3 findings, all valid, all fixed in a137b069b)
- `pidinstAccepted` credited DataCite Commons for a B2INST record
- Envers query took the tail of an unordered result list — **also a latent defect**, now ordered explicitly
- Hard-coded exception text in `refreshIdentifier`, now a localized error code

### Cycle 2 — Copilot (2 findings, both valid, both fixed in 750e82dd1)
- `url` rebuilt from the RSpace server URL after acceptance instead of the registered `metadata.LandingPage` — an accepted instrument could report a URL its own Handle does not resolve to
- The accepted-state text then had to say "registered landing page", since the address may be institutional

### Cycle 3 — parallel review (11 findings: 9 fixed, 2 already fixed, 1 left) in 341228045 / f53fdb626
Public-page revision split completed; `state` made `READ_ONLY`; B2INST record ids validated and escaped (including the two pre-existing calls that shared the flaw); accepted-with-no-record no longer publishes; state label shared between table and alerts; Refresh given a spinner and a `catch`; test double made honest; end-to-end UI test added.

**Notable:** the in-flight DAO edit had broken `retrieveLastPublishedVersionOfItemWithPublicLink`. `asc()` + `setMaxResults(1)` returns the *draft* the identifier started as, not the oldest revision of the trailing published run, so every published IGSN lost its page. The DataCite branch keeps the backwards walk; a null type now lands there rather than leaving the query null.

### Cycle 4 — Copilot (2 findings) in 2bd5642c1
- Accepted PID with no Handle could never be refreshed — **fixed.** The stated cause was stale (that backend path now refuses), but `epicPidOf` returning null reaches the same outcome
- `created` review supposedly unable to retry — **not a defect.** `openReviewOf` filters on `is_open`, and an unsubmitted review reports `is_open: false`, so it already falls through and resubmits. It was the one untested path, so it now has a test that passes unchanged

### Deliberately not fixed (1)
**Refresh routed through the write path** (`InventoryIdentifierApiManagerImpl.java:339`). Verified in full: the edit lock, the `modificationDate`/`modifiedBy` re-stamp, the version bump, and the transaction held across three 30s provider calls. Left because it is a transaction-boundary change on a `*Manager` bean, which is an architectural call rather than a review fix. Proposed shape is in the thread; the spurious `modifiedBy` stamp is the part worth prioritising.

### Verification
Backend 136/136 plus 40/40 on the connector; frontend 52/52 across the identifier suites; `tsc`, Biome and `i18n:check` clean; Spotless clean; every pre-commit hook passed. The new IT was confirmed to be a real guard by flipping the Envers order and watching it fail. ITs ran against the existing dev DB, never `drop-recreate-db`.

### Bot status
- **Copilot:** reviewed in all four cycles
- **Greptile:** never responded. Last posted in this repo on 2026-07-29; its status page reports every component operational, so this looks org-side — a GitHub App installation or seat issue — rather than an outage. Worth someone checking the installation. Convergence was judged on Copilot alone

Two threads left open, both yours: the `state` finding (fixed, left for you to confirm) and the one above.

All code changed must follow the AI and Third Party code policy in Drata.

src/main/webapp/ui/src/Inventory/components/Fields/Identifiers/__tests__/RefreshButton.test.tsx
```diff
@@ -0,0 +1,158 @@
...
+    render(
+      <ThemeProvider theme={materialTheme}>
+        <IdentifiersList activeResult={record} />
+      </ThemeProvider>,
+    );
+
+    // while the review is open: Submitted, Refresh offered, Publish held back
+    expect(screen.getByText("inventory:fields.identifiers.list.stateLabels.submitted")).toBeInTheDocument();
+    const refresh = screen.getByRole("button", { name: "inventory:fields.identifiers.list.refresh" });
+    expect(screen.getByRole("button", { name: /republish|publish/i })).toBeDisabled();
```
tchoi-rs: Avoid regex where possible

src/main/webapp/ui/src/Inventory/components/Fields/Identifiers/__tests__/RefreshButton.test.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+    fireEvent.click(refresh);
```
tchoi-rs: We use `userEvent` instead of `fireEvent` (the React Testing Library skill installed in this repo should cover this)

src/main/webapp/ui/src/Inventory/components/Fields/Identifiers/__tests__/RefreshButton.test.tsx
```diff
@@ -0,0 +1,158 @@
...
+    fireEvent.click(button);
+
+    await waitFor(() => {
+      const refresh = identifier.refresh;
+      expect(vi.mocked(refresh)).toHaveBeenCalled();
+    });
+  });
+
+  test("renders nothing unless the state is submitted", () => {
+    for (const state of ["draft", "created", "declined", "findable"] as const) {
```
tchoi-rs: Use https://vitest.dev/guide/learn/writing-tests.html#parameterized-tests (renders nothing when state is X)

src/main/webapp/ui/src/stores/definitions/Identifier.ts
```diff
@@ -83,6 +84,50 @@ export type PidinstPublishingState = "created" | "submitted" | "accepted" | "dec
...
+ * <p>Shared so the table and the alerts cannot disagree. The state cell and the refresh alert used
+ * to be produced separately, which showed the provider's raw English token inside an otherwise
+ * translated sentence.
+ *
+ * Arms are spelled out rather than built from the state, so the typed catalog and the i18n
+ * extractor both still see every key. The final fallback is deliberate: the server passes an
+ * unrecognised provider status through verbatim, and that must degrade to the raw value rather
+ * than throwing out of a render.
+ */
+export const identifierStateLabel = (state: PublishingState): string => {
```
tchoi-rs: It should be fine to return the translation keys rather than the actual strings (I think)

tchoi-rs (changes requested):

src/main/webapp/ui/src/Inventory/components/Fields/Identifiers/__tests__/RefreshButton.test.tsx
rs-nicof: Done in 4c4fc9fb2 — and you were right that there was no need for it here.

I had reached for `/republish|publish/i` because `PublishButton` swaps its label between `common:actions.republish` and `common:actions.publish`. But `republish` keys off `state === "findable"`, which is false for both states this test covers (`submitted`, then `accepted`), so the name is just `common:actions.publish` in both assertions. Tests run in `cimode`, so that is the literal accessible name.

Both are now exact, hoisted into constants next to `REFRESH` so the intent is visible at a glance. The only remaining regex-ish thing is a `stateLabel(state)` helper that builds the state-label key by interpolation, which is test-local string building rather than a matcher.

Worth noting `Identifiers.test.tsx` still uses `/republish|publish/i` in a couple of places from before this branch. Happy to tidy those in a follow-up if you would like them consistent.

src/main/webapp/ui/src/Inventory/components/Fields/Identifiers/__tests__/RefreshButton.test.tsx
rs-nicof: Done in 4c4fc9fb2. Read the skill properly this time, and it changed more than the one line.

All `fireEvent.click` calls are now `userEvent`, with `userEvent.setup()` before `render` and every interaction awaited, per `user-setup-before-render` and `user-await-interactions`.

Awaiting the click also removed the scaffolding around it. The first test no longer needs `waitFor` at all — the click is awaited, so the `refresh` assertion is just a plain assertion afterwards. And in the integration tests `waitFor` wrapping `getByText` became `await screen.findByText(...)`, which is what `async-findby-over-waitfor` asks for. `waitFor` is gone from the file.

Also picked up two nearby rules while I was in there: `toBeVisible()` over `toBeInTheDocument()` for the present cases, and `queryByRole(...)` over `expect(container).toBeEmptyDOMElement()` for the absent ones, so the file no longer asserts on `container`.

src/main/webapp/ui/src/Inventory/components/Fields/Identifiers/__tests__/RefreshButton.test.tsx
rs-nicof: Done in 4c4fc9fb2.

Now `test.each([...] as const)("renders nothing when the state is %s", (state) => ...)`, matching your suggested wording. Four separate named tests rather than one that stops at the first failure — the loop would also have reported only the first bad state.

That loop predates this branch; I had only removed `accepted` from its array when Refresh started staying available for an accepted PID with no Handle. Converted the whole thing rather than leaving a half-converted file.

`test.each` also matches the existing precedent in `ScopeField.test.tsx` and `MoreInfoLinkedItems.test.tsx`, so it is the house style anyway.

src/main/webapp/ui/src/stores/definitions/Identifier.ts
rs-nicof: You are right, and it is better this way — done in 4c4fc9fb2.

`identifierStateLabel` is now `identifierStateLabelKey` and returns the catalog key, with each caller translating:

- `Identifiers.tsx` uses the component's own `t` from `useTranslation`, so the State cell now follows a language change. Reading the singleton once was a latent bug I had introduced.
- `IdentifierModel` has no component `t`, so it resolves on the shared instance, which is what it already does for every other alert.

Two things I checked before committing, since both could have made this awkward:

1. The key union types through `t()` with no cast. Because the arms are literal strings, TypeScript infers a union of valid catalog keys, which `t()` accepts — `pnpm tsc` is clean. That was my original reason for translating inside the function, and it turns out not to be a reason at all.
2. `definitions/Identifier.ts` no longer imports the i18n singleton, which is the layering improvement I take to be your actual point.

An unrecognised provider status returns `null` rather than a key, and each caller falls back to the raw value — the server passes provider statuses through verbatim, so that path is reachable. That is a one-line fallback duplicated across the two callers; the state-to-key mapping, which is the part that must not drift, stays single-sourced.

55/55 across the identifier suites, `tsc`, Biome and `i18n:check` clean.

rs-nicof:
## Tim's review addressed — 4c4fc9fb2

All four points taken; every one was a fair call.

| Comment | Change |
|---|---|
| Avoid regex | Exact accessible names. `republish` keys off `findable`, so Publish is `common:actions.publish` in both states this test covers — the regex was never needed |
| `userEvent`, not `fireEvent` | `userEvent.setup()` before render, interactions awaited. Awaiting them also let `waitFor` go entirely: `findByText` replaces `waitFor` + `getByText`, and the first test needs no wrapper at all |
| Parameterized tests | `test.each` for the four states that render nothing, so each reports separately instead of stopping at the first |
| Return keys, not strings | `identifierStateLabelKey` returns the catalog key. The table cell now translates with the component's own `t`, so it follows a language change, and `definitions/Identifier.ts` no longer imports the i18n singleton |

On the last one: the key union types through `t()` with no cast, since the arms are literal strings. That was my original reason for translating inside the function, and it turned out not to be a reason — your instinct was right.

Two extra rules picked up from the skill while I was in the file: `toBeVisible()` over `toBeInTheDocument()`, and `queryByRole` over asserting on `container`.

**Verification:** 55/55 across the identifier suites (the `test.each` split raised the count), `tsc`, Biome and `i18n:check` clean, all pre-commit hooks passed.

One thing left for you to decide: `Identifiers.test.tsx` still has `/republish|publish/i` in a couple of places from before this branch. I left them alone to keep this diff to your review, but happy to make them consistent if you would like.

Threads left open for you rather than resolved.

All code changed must follow the AI and Third Party code policy in Drata.

src/main/webapp/ui/src/Inventory/components/Fields/Identifiers/__tests__/RefreshButton.test.tsx
rs-nicof: Followed this through into `Identifiers.test.tsx` as offered — c909b1a80. All ten remaining regex matchers there are now exact keys, so the convention holds across both test files.

Two of them were matching the wrong label, not merely loose:

- `deleteOrRetract.retract` contains the word "delete", so `/delete|retract/i` passed for the Delete button too
- `actions.republish` contains "publish", so `/publish/i` passed for Republish as well

Which is the same trap the existing comment in that file already called out for `/retract/i` — it just had not been applied to the rest.

Getting the exact names right meant working out which label each state actually renders, and the tests passing is what confirms it: the identifier the historical-version test renders is `findable`, so its buttons are **Republish** and **Retract**; the PIDINST states are none of them findable, so theirs are **Publish**, with **Retract** for the open and published reviews (`deletable = state === "draft" || isDeletableClosedReview(id)`) and **Delete** for the closed ones. A wrong key throws "unable to find", so 38/38 green is the evidence rather than just an absence of failures.

Two assertions collapsed into their query as a result:

- the accepted case checked `not.toHaveTextContent(/republish/i)`, which an exact name settles by itself, so it now asserts Republish is absent outright — a clearer statement of the same claim
- the closed-review case no longer needs a separate `toHaveTextContent` for the Delete label

Names are hoisted into `PUBLISH`/`REPUBLISH`/`RETRACT`/`DELETE` constants with a comment explaining why they are exact, mirroring `RefreshButton.test.tsx`.

For the record, your other two test points do not apply to this file: it uses no `fireEvent` and no `waitFor`. It does have a few `expect(container).toHaveTextContent(...)` assertions and some `toBeInTheDocument()` where `toBeVisible()` would be stronger, all predating this branch — I left those alone to keep the diff to what you asked for, but happy to sweep them if you want.

55/55 across the identifier suites; `tsc`, Biome and `i18n:check` clean.

tchoi-rs (approved):
Frontend changes look fine!

mrowe-tan (approved):

PR #1067, opened 2026-08-26
Dbrepo integration

### LINKED ISSUE #1065: [IDEA] DBRepo integration
### Summary

DBRepo is a repository software for relational databases. You can create databases, upload tables, create views and subsets, and add new data in a web gui.  Learn more here: https://dbrepo-project.github.io/dbrepo-docs/latest/

The RSpace integration would allow users to place links to DBRepo datasets in their documents and download the data. The functionality is similar to the filestore integration.

### Motivation

This feature benefits the researchers, who already use DBRepo and want to refer to their data there from RSpace. Already working on it.

### User Story

As a researcher I want to connect RSpace to my DBRepo instance so that I can authenticate and use the integration.
As a researcher I want to browse and select my DBRepo datasets from RSpace so that they can be linked in an RSpace document.

### Must Have

- A new RSpace app for the integration.
- A new editor plugin for inserting links to the DBRepo datasets.

### Should Have

- Clicking the DBRepo link should display an info card about the dataset
- The info card should have buttons for opening the resource in DBRepo and one for downloading the data directly (if applicable)

### Could Have

- Another functionality for the plugin for inserting selected table rows of a DBRepo dataset into the RSpace document.

### Won't Have

_No response_

akos326:
## Description ##
Closes #1065 . Implements the DBRepo integration as a new RSpace app according to the issue.
You can select a database, table, view or subset for insertion in this interface:

image

The collapsible menu items are databases. Tables are identified by their name, views are by their name and query and subsets are by only their query as they (sadly) don't have names.

A DBRepo table info card:

image

The interface for selecting rows:

image

TODO: there is no documentation yet, I need support on how to write and submit one.

## Design decisions
- Users authenticate with username and password, which are stored in UserConnection.
- The GUI of the integration is designed after existing integrations. Selecting a DBRepo dataset is similar to the nextcloud integration and the info card (q.v. the should have section of the issue) is to the filestore info cards.
- Implemented the functionality in the could have section of the issue: selecting rows of a DBRepo tables and views and inserting them into the document in a similar manner to the PyRAT integration. DBRepo subsets are not supported because the DBRepo API doesn't return sufficient metadata about their columns. If at a later time the API is extended, subset support can be easily added to this feature.

## Testing notes
There is a DBRepo test site at https://test.dbrepo.tuwien.ac.at/ which I used for manual testing. Unit tests mock the server.


rs-nicof (commented):

Reviewed the DBRepo integration with six parallel passes (layering, security, persistence, frontend, i18n, general quality) and verified each finding against the code. 15 comments below, covering the outbound HTTP client, error handling, a few user-visible frontend bugs and some untranslated text. The Liquibase changelog, the credential storage and the no-XSS design of the inserted table all came back clean.

src/main/java/com/researchspace/webapp/integrations/dbrepo/DBRepoClient.java
```diff
@@ -0,0 +1,401 @@
...
+  static final String TABLE_TYPE = "table";
+  static final String VIEW_TYPE = "view";
+  private static final String SUBSET_TYPE = "subset";
+  private static final MediaType TEXT_CSV = MediaType.parseMediaType("text/csv");
+
+  private final RestTemplate restTemplate;
+  private final ObjectMapper objectMapper;
+
+  public DBRepoClient() {
+    this(new RestTemplate(), new ObjectMapper());
```
rs-nicof: > Written by Claude, an AI coding agent.

This client has no connect or read timeout, so a DBRepo host that accepts the connection and then never replies holds an RSpace request thread indefinitely. `B2instConnectorImpl` already has a `timeoutBoundedRequestFactory()` you can pass straight into the constructor.

src/main/java/com/researchspace/webapp/integrations/dbrepo/DBRepoClient.java
```diff
@@ -0,0 +1,401 @@
...
+
+  public List<DBRepoDatabaseDTO> listDatabases(String baseUrl, DBRepoCredentials credentials) {
+    String normalizedBaseUrl = normalizeBaseUrl(baseUrl);
+    try {
+      return listDatabasesAt(
+          normalizedBaseUrl + CURRENT_DATABASES_PATH, normalizedBaseUrl, credentials);
+    } catch (HttpStatusCodeException e) {
+      if (HttpStatus.NOT_FOUND.value() == e.getStatusCode().value()) {
+        return listDatabasesAt(
+            normalizedBaseUrl + LEGACY_DATABASES_PATH, normalizedBaseUrl, credentials);
```
rs-nicof: > Written by Claude, an AI coding agent.

This legacy path fallback only applies to listing databases. Expanding a database, previewing rows and downloading CSV all hardcode the `/api/v1` path, so against a server that actually needs the fallback, connecting would appear to work and then everything else would fail. Worth either applying it everywhere or dropping it.

src/main/java/com/researchspace/webapp/integrations/dbrepo/DBRepoClient.java
```diff
@@ -0,0 +1,401 @@
...
+    String normalizedBaseUrl = normalizeBaseUrl(baseUrl);
+    String url =
+        normalizedBaseUrl
+            + "/api/v1/database/"
+            + encodePathSegment(databaseId)
+            + "/"
+            + resourceType
+            + "/"
+            + encodePathSegment(resourceId)
+            + "/data";
```
rs-nicof, line 145: > Written by Claude, an AI coding agent.

The entire CSV is loaded into memory before any of it reaches the browser, with no size limit, so one large table can take a sizeable bite out of the heap and a few concurrent downloads could affect everyone on the instance. Streaming the response through to the servlet output stream avoids that.

src/main/java/com/researchspace/webapp/integrations/dbrepo/DBRepoController.java
```diff
@@ -0,0 +1,292 @@
...
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.client.HttpStatusCodeException;
+
+@Slf4j
+@Controller
+@RequestMapping("/apps/dbrepo")
+public class DBRepoController {
```
rs-nicof: > Written by Claude, an AI coding agent.

This controller does not extend `BaseController`, so nothing here turns an exception into a sensible response. When the DBRepo server is down, or a download URL carries an unsupported resource type, the browser receives a 500 HTML error page where it expects JSON or a file, and the dialog shows a parse failure rather than a message. Extending `BaseController`, or catching per endpoint and mapping to a status, would fix both.

src/main/java/com/researchspace/webapp/integrations/dbrepo/DBRepoController.java
```diff
@@ -0,0 +1,292 @@
...
+      @RequestParam String dbrepoUsername,
+      @RequestParam String dbrepoPassword,
+      Principal principal,
+      Model model) {
+    try {
+      String normalizedUrl = dbRepoClient.normalizeBaseUrl(dbrepoUrl);
+      DBRepoCredentials credentials = new DBRepoCredentials(dbrepoUsername, dbrepoPassword);
+      dbRepoClient.listDatabases(normalizedUrl, credentials);
+      User user = userManager.getUserByUsername(principal.getName());
+      saveUrl(user, normalizedUrl);
```
rs-nicof, line 89: > Written by Claude, an AI coding agent.

The server URL and the credentials are saved in two separate transactions, so if the second one fails the user is left with the new server address paired with the previous username and password, and the next request would send the old credentials to the new host. Saving the credentials before the URL shrinks that window considerably.

src/main/java/com/researchspace/webapp/integrations/dbrepo/DBRepoController.java
```diff
@@ -0,0 +1,292 @@
...
+            databaseId,
+            resourceType,
+            resourceId,
+            details.get().credentials());
+    ResponseHeaders.setContentTypeAndPreventSniffing(response, TEXT_CSV);
+    response.setContentLength(csv.length);
+    response.setHeader(
+        "Content-Disposition",
+        String.format("attachment; filename=\"%s\"", csvFilename(resourceType, resourceId)));
+    response.getOutputStream().write(csv);
```
rs-nicof: > Written by Claude, an AI coding agent.

This hands the browser whatever the configured host sent back, and the default client follows redirects, so someone who points the integration at a host they control can get RSpace to fetch an internal-only URL and download the contents. Disabling redirect following on the client, and checking the response is really CSV, would close that off. This is the only integration endpoint that returns an upstream body verbatim, so it is worth a second look.

src/main/java/com/researchspace/service/impl/IntegrationsHandlerImpl.java
```diff
@@ -689,6 +699,7 @@ private boolean isSingleOptionSetAppConfigIntegration(String integrationName) {
       case DMPTOOL_APP_NAME:
       case DMPASSISTANT_APP_NAME:
       case PYRAT_APP_NAME:
+      case DBREPO_APP_NAME:
```
rs-nicof, line 702: > Written by Claude, an AI coding agent.

Registering DBRepo here routes it through the single-option-set save below, which has no DBRepo case and so falls through to the default. The URL the Apps card sends when you enable or disable the integration is therefore silently discarded. Either add a case, or stop sending the URL on that path since the Connect form is what actually saves it.

src/main/webapp/ui/src/tinyMCE/dbrepo/DBRepo.tsx
```diff
@@ -0,0 +1,911 @@
...
+          },
+        },
+      }}
+    >
+      <DialogTitle sx={{ pb: 1 }}>{t("tinymce.dbrepo.rows.dialogTitle", { name: target.name })}</DialogTitle>
+      <DialogContent sx={{ display: "flex", flexDirection: "column", minHeight: 0, overflow: "hidden", pt: 1 }}>
+        <Stack spacing={1} sx={{ flex: 1, minHeight: 0 }}>
+          <Typography variant="body2" color="text.secondary">
+            {t("tinymce.dbrepo.rows.resourceContext", {
+              name: <b>{target.name}</b>,
```
rs-nicof, line 503: > Written by Claude, an AI coding agent.

`t()` returns a string, so these bold elements get stringified and this subtitle renders as `[object Object] in [object Object]`. `TransRichText` is the supported way to put markup inside a translated string.

src/main/webapp/ui/src/tinyMCE/dbrepo/DBRepo.tsx
```diff
@@ -0,0 +1,911 @@
...
+                  }}
+                  hover
+                  tabIndex={-1}
+                  role="checkbox"
+                  onClick={() => onToggleRow(key, row)}
+                  aria-checked={selected}
+                  selected={selected}
+                >
+                  <TableCell padding="checkbox">
+                    <Checkbox color="primary" checked={selected} />
```
rs-nicof, line 641: > Written by Claude, an AI coding agent.

This checkbox has no accessible name, so a screen reader announces an unlabelled checkbox on every row with no way to tell which row it belongs to. The select-all checkbox just above sets an `aria-label`, and the Omero and Clustermarket result tables point theirs at the row with `aria-labelledby`.

src/main/webapp/ui/src/tinyMCE/dbrepo/DBRepo.tsx
```diff
@@ -0,0 +1,911 @@
...
+}
+
+function resourceApiPath(target: RowTemplateTarget, suffix: "metadata" | "rows"): string {
+  return `/apps/dbrepo/databases/${encodeURIComponent(target.databaseId)}/${target.dbrepoType}/${encodeURIComponent(
+    target.resourceId,
+  )}/${suffix}`;
+}
+
+function rowKey(page: number, index: number, row: DBRepoRow): string {
+  return `${page}:${index}:${JSON.stringify(row)}`;
```
rs-nicof, line 785: > Written by Claude, an AI coding agent.

Selection keys include the page number and row index, but changing rows-per-page resets the page without clearing the selection. The selected count then keeps counting rows whose checkboxes no longer appear ticked, and Insert adds them anyway. Keying on the row contents alone, or on a primary key column, removes the problem.

src/main/webapp/ui/src/tinyMCE/dbrepo/DBRepo.tsx
```diff
@@ -0,0 +1,911 @@
...
+  selectedRows: DBRepoRow[],
+): HTMLTableElement {
+  const dbrepoTable = document.createElement("table");
+  dbrepoTable.setAttribute("data-tableSource", "dbrepo");
+  dbrepoTable.style.fontSize = "1em";
+
+  const linkRow = document.createElement("tr");
+  const linkCell = document.createElement("th");
+  linkCell.style.fontSize = "0.7em";
+  linkCell.appendChild(document.createTextNode("Imported from DBRepo "));
```
rs-nicof, line 831: > Written by Claude, an AI coding agent.

This caption is stamped permanently into the user's document in hard-coded English, including an English date format, so no translation can ever reach it. Worth building it from a translated message with the date formatted for the user's locale.

src/main/webapp/ui/src/tinyMCE/dbrepo/__tests__/DBRepo.test.tsx
```diff
@@ -0,0 +1,431 @@
...
+        url: "https://dbrepo.example/database/db-1",
+      },
+    ]);
+
+    render(<DBRepo />);
+
+    expect(await screen.findByText("Research data")).toBeVisible();
+    editorListeners.get("dbrepo-insert-rows")?.();
+
+    expect(screen.queryByText("workspace:tinymce.dbrepo.rows.tableLabel")).not.toBeInTheDocument();
```
rs-nicof, line 273: > Written by Claude, an AI coding agent.

This assertion cannot fail. That key is only ever used as an `aria-label`, never as visible text, so `queryByText` returns null whether or not the row picker opened. Querying the dialog by role, or by its title, would actually test the behaviour the test name describes.

src/main/webapp/ui/src/modules/common/i18n/locales/en-US/common.json
```diff
@@ -246,6 +246,7 @@
     "createTemplate": "c8sxesdqpy-create-a-template",
     "createTemplateFromSample": "c8sxesdqpy-create-a-template#create_a_template_from_a_sample",
     "dataverse": "h14qd5tvjj-dataverse-integration",
+    "dbrepo": "TODO_DBREPO_DOC_ID",
```
rs-nicof, line 249: > Written by Claude, an AI coding agent.

This placeholder ships as a live link: the docs link on the DBRepo card is built from this slug, so anyone clicking it lands on a 404. It needs the real article id, or the link should come out until there is one.

src/main/webapp/scripts/pages/workspace/editor/public_documentView.js
```diff
@@ -35,6 +35,10 @@ const disableLinkedDocumentsAndLinkedFiles = () => {
     files.each(function (index) {
         $(files[index]).removeAttr('href').css('opacity', 0.3);
     });
+    const dbrepoLinks = $(".dbrepo_link");
+    dbrepoLinks.each(function (index) {
+        $(dbrepoLinks[index]).removeAttr('href').css('opacity', 0.3);
```
rs-nicof, line 40: > Written by Claude, an AI coding agent.

Removing the href hides the link, but the same URL is still on the element as a data attribute, along with the view or subset's SQL query and the database name. Anyone viewing a published document can read all of that from the page source, so those attributes need stripping too.

src/main/webapp/scripts/externalTinymcePlugins/dbrepo/plugin.min.js
```diff
@@ -0,0 +1,62 @@
...
+      ],
+      onAction(dialogApi, button) {
+        if (button.name === "insert") {
+          editor.fire("dbrepo-insert");
+        } else if (button.name === "insertRows") {
+          editor.fire("dbrepo-insert-rows");
+        }
+      },
+      onMessage(dialogApi, details) {
+        const $insertButton = $("button[title='Insert link']");
```
rs-nicof, line 22: > Written by Claude, an AI coding agent.

These lookups match on the English button labels, so as soon as those labels are translated the Insert buttons stop being enabled and the feature quietly breaks. Matching on the button id or name instead would be stable. The labels themselves are also hard-coded English; the `video` plugin next door shows the `RS.msg(...)` pattern.

src/main/webapp/ui/src/eln/apps/integrations/DBRepo.tsx
```diff
@@ -0,0 +1,183 @@
...
+    addAlert(
+      mkAlert({
+        variant: "success",
+        message: t("integrations.dbrepo.alerts.connectSuccess"),
+      }),
+    );
+  });
+
+  const handleDisconnect = async () => {
+    const response = await fetch("/apps/dbrepo/connect", {
```
rs-nicof, line 60: > Written by Claude, an AI coding agent.

An HTTP error is handled below, but a request that never completes (offline, server unreachable, blocked) rejects instead, and the call site discards the promise, so the user gets no error and no change in the UI and cannot tell whether they are still connected. `useDisconnectEndpoint` already does this over the shared axios instance with the error handling, and both Omero and Clustermarket use it.

rs-nicof (commented):

Two follow-ups to the earlier review, both smaller than the rest but easy to fix.

src/main/webapp/images/icons/dbrepo.svg
```diff
@@ -0,0 +1,11 @@
+<svg version="1.2" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 265 265" width="265" height="265">
+	<title>favicon</title>
```
rs-nicof, line 2: > Written by Claude, an AI coding agent.

This is a base64 PNG wrapped in an SVG rather than a real vector, still titled `favicon`, and the identical 9.7 KB file is also checked in as the branding logo, so the two will have to be kept in step by hand. One vector file referenced from both places would be easier to live with.

src/main/webapp/ui/src/modules/common/i18n/locales/en-US/common.json
akos326: I haven't added any documentation yet, because I need help on how to submit one. Once it's done, I'll add the link here.

src/main/webapp/images/icons/dbrepo.svg
akos326: Now I use only one of them.

src/main/webapp/ui/src/eln/apps/integrations/DBRepo.tsx
akos326: done

src/main/webapp/scripts/externalTinymcePlugins/dbrepo/plugin.min.js
akos326: Got rid of the hardcoded labels

src/main/webapp/scripts/pages/workspace/editor/public_documentView.js
akos326: Removed the other propertoes

src/main/webapp/ui/src/tinyMCE/dbrepo/__tests__/DBRepo.test.tsx
akos326: Fixed

src/main/webapp/ui/src/tinyMCE/dbrepo/DBRepo.tsx
akos326: Made it localised

src/main/webapp/ui/src/tinyMCE/dbrepo/DBRepo.tsx
akos326: Changed the key to only use the row content

src/main/webapp/ui/src/tinyMCE/dbrepo/DBRepo.tsx
akos326: Added a meaningful aria-label

src/main/webapp/ui/src/tinyMCE/dbrepo/DBRepo.tsx
akos326: For me, it rendered as expected, but I switched to TransRichText

src/main/java/com/researchspace/service/impl/IntegrationsHandlerImpl.java
akos326: Fixed

src/main/java/com/researchspace/webapp/integrations/dbrepo/DBRepoController.java
akos326: Disabled redirecting, and DBRepoClient now only allows csvs

src/main/java/com/researchspace/webapp/integrations/dbrepo/DBRepoController.java
akos326: Switched the save order

src/main/java/com/researchspace/webapp/integrations/dbrepo/DBRepoController.java
akos326: I extended BaseController

src/main/java/com/researchspace/webapp/integrations/dbrepo/DBRepoClient.java
akos326: Done

src/main/java/com/researchspace/webapp/integrations/dbrepo/DBRepoClient.java
akos326: Removed the fallback.

src/main/java/com/researchspace/webapp/integrations/dbrepo/DBRepoClient.java
akos326: Added a timeout

tilorspace:
Thanks for this! 
Really nice to see this working end to end. What I would like to tighten before it ships is the copy on the three surfaces where a user meets the integration: the gallery card, the picker, and the link details dialog. Most RSpace users will not know DBRepo yet, so these screens are where they decide whether this is for them and what to do next. Everything below is copy and small affordances, no changes to the logic.

**1. The app card should describe the integration, not the product**

Right now the card reads:

> DBRepo A repository software for relational databases.

That is an accurate description of DBRepo, but it tells the user nothing about what enabling the integration lets them do in RSpace. 

Suggested replacement, pick whichever reads best to you, my preference would be B:

> A. Link tables, views and subsets from a DBRepo instance into your RSpace documents.

> B. Insert data from relational databases published in DBRepo directly into your RSpace documents, or link back to the published record.

**2. The picker dialog needs instance context and one line of orientation**

Two things are missing. First, an indication which DBRepo instance the user is on. Second, there is no explanation of how to get from this list to inserted data, and the two buttons at the bottom do not explain themselves. 
Here are some suggestions:

1. Header, under or beside the DBRepo title:

> Connected to {instance name} · {instance-url}

Making the URL a link out to the instance would help people confirm they are in the right place.

2. One-line explainer above the list:

> Select a database to insert a link to it. Expand a database to pick a single table, view or subset and insert its rows into your document.

3. Button states. Insert rows is currently disabled with no indication why, which reads as broken rather than as not-yet. Two options:

Tooltip on the disabled button:

>   Expand a database and select a table or subset to insert its rows.

Or inline helper text next to the buttons, which is more discoverable:
 
>  Insert rows needs a table or subset. Expand a database to choose one.



**3. The link details dialog could use more detail**

This is what a reader clicks months later to find out what the link actually points at, so it is worth adding more detail. 

1.  Add the instance. Same reason as above, and more important here: without it the dialog cannot tell a reader which DBRepo the record lives in.

> Instance:   {instance name} ({instance-url})

2. Drop the duplicated title. "DBRepo link details" appears in the header and again as the first line of the body. The body line can go.

3. Show the identifier. Right now the only way to find out where the link goes is to click through. A visible PID or URL, ideally with a copy affordance, makes this dialog useful for citation rather than just navigation:

> Identifier:   {PID or record URL}

**4. Nice to have: a search field over the database list**

29 databases already needs scrolling, and this list probably grows with the instance. A plain filter box at the top of the picker would help a lot:

Placeholder: "Search databases"
No results: "No databases match '{query}'."

Not a blocker for this PR, but worth a follow up issue if usage picks up..

akos326:
Thanks for the suggestions, @tilorspace! I implemented the first three. At the current scale I don't think the searh field is necessary, but I'm open to adding it later. As for the identifiers, I decided to go with the URLs for now. DBRepo resources only have a PID if their owner registers one. I'm planning to explore this functionality more and add PID support to the integration in a later PR. It will involve displaying PIDs as identifiers and adding them to the export metadata.

The way I see it, the only thing left for me right now is the documentation for which I still need support on how to add one and what the expected format is.

src/main/java/com/researchspace/webapp/integrations/dbrepo/DBRepoClient.java
```diff
@@ -0,0 +1,430 @@
...
+      log.warn("Could not count DBRepo rows at {}", dataUrl, e);
+      return Optional.empty();
+    }
+  }
+
+  private List<DBRepoDatabaseDTO> listDatabasesAt(
+      String url, String normalizedBaseUrl, DBRepoCredentials credentials) {
+    ResponseEntity<JsonNode> response =
+        restTemplate.exchange(
+            url, HttpMethod.GET, new HttpEntity<>(headers(credentials)), JsonNode.class);
```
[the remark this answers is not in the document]
akos326: The credentials are username and password coming from the user and saved by RSpace, like in the omero integration

tilorspace:
> The way I see it, the only thing left for me right now is the documentation for which I still need support on how to add one and what the expected format is.

@akos326 , we don't have a good contribution workflow for the product documentation yet that works with our current documentation solution, but you can simply attach a markdown file including screenshots to the original issue, following a structure like this: https://documentation.researchspace.com/l/en/article/j2z5f5r90q-cloud-store-integrations 

we can then take care of it from there. does that work for you?

src/main/java/com/researchspace/webapp/integrations/dbrepo/DBRepoClient.java
```diff
@@ -0,0 +1,430 @@
...
+
+  private void validateRowsResourceType(String resourceType) {
+    if (!TABLE_TYPE.equals(resourceType) && !VIEW_TYPE.equals(resourceType)) {
+      throw new IllegalArgumentException(
+          "DBRepo row insertion is only supported for tables and views.");
+    }
+  }
+
+  private void validateCsvContentType(MediaType contentType) {
+    if (contentType == null || !TEXT_CSV.isCompatibleWith(contentType)) {
```
rs-nicof: > Written by Claude, an AI coding agent.

`isCompatibleWith` treats a wildcard on either side as a match, so an upstream answering `*/*` or `text/*` passes this check and gets its body streamed through RSpace anyway. `TEXT_CSV.includes(contentType)` is the asymmetric version: it rejects wildcards while still accepting `text/csv;charset=utf-8`. The new test only covers a JSON response, so the wildcard case would not be caught.

src/main/java/com/researchspace/webapp/integrations/dbrepo/DBRepoController.java
```diff
@@ -0,0 +1,292 @@
...
+      @PathVariable String resourceId,
+      Principal principal,
+      HttpServletResponse response)
+      throws IOException {
+    Optional<DBRepoConnectionDetails> details = findConnectionDetails(principal);
+    if (details.isEmpty()) {
+      response.sendError(HttpStatus.UNAUTHORIZED.value());
+      return;
+    }
+    ResponseHeaders.setContentTypeAndPreventSniffing(response, TEXT_CSV);
```
rs-nicof: > Written by Claude, an AI coding agent.

The content type and the attachment filename are now set before the first upstream byte arrives, but the streaming call below can still fail after that. When it does, the user is handed the error page as a file named like a spreadsheet, or a truncated file that looks like a complete export. Checking the content type before any header is set closes the window.

src/main/webapp/scripts/externalTinymcePlugins/dbrepo/plugin.min.js
```diff
@@ -0,0 +1,99 @@
...
+      ],
+      onAction(dialogApi, button) {
+        if (button.name === "insert") {
+          editor.fire("dbrepo-insert");
+        } else if (button.name === "insertRows") {
+          editor.fire("dbrepo-insert-rows");
+        }
+      },
+      onMessage(dialogApi, details) {
+        const $buttons = $(".tox-dialog__footer button");
```
rs-nicof: > Written by Claude, an AI coding agent.

This matches footer buttons across the whole page and then picks them by position, and a few lines down it physically moves every match into the new row. If a second editor dialog is open at the time, this reaches into that dialog's footer and takes its buttons. Scoping the lookup to this dialog's own footer and matching each button on the same label string used to create it keeps the language independence you just added, without depending on their order.

rs-nicof (commented):

Second pass over the fixes. Thanks for working through all of them, most are genuinely done and I have resolved those threads.

Four threads I have left open with a reply explaining why, and three new points below that came in with the latest changes. Worth noting the code scanner has also raised the server-side request forgery point independently.

src/main/java/com/researchspace/webapp/integrations/dbrepo/DBRepoController.java
rs-nicof: > Written by Claude, an AI coding agent.

Thanks, turning redirects off closes the important half of this, which was the stored password being re-sent to whatever host a redirect pointed at.

Two things are still open, and the code scanner has now raised the same area independently. A redirect response is not treated as an error, so its body still reaches the handler: a host that redirects makes the connection look successful and the credentials get stored, while nothing is listed. Treating those responses as errors would cover that and the download case together. Separately, the address itself is still unfiltered, so the server can be pointed at internal-only hosts, which is what the scanner is reporting.

src/main/webapp/ui/src/tinyMCE/dbrepo/DBRepo.tsx
rs-nicof: > Written by Claude, an AI coding agent.

My suggestion was too blunt, sorry. Keying on the contents alone does fix the paging case, but two identical rows now collide: they share a key, ticking one ticks both, the count disagrees with what is on screen, and only one of them gets inserted. Combining the contents with the row's position in the full result set keeps both properties.

Worth knowing as well that the new test passes against the old key too, because the mock returns the same rows at both page sizes and the page never changes, so it would not catch a revert.

src/main/webapp/ui/src/modules/common/i18n/locales/en-US/common.json
rs-nicof: > Written by Claude, an AI coding agent.

Understood, thanks for flagging it rather than leaving it silent. Leaving this open as the reminder.

When the article exists, this just needs the slug from the published help site in the same shape as the neighbouring entries. Someone on the team who owns the help docs should be able to get the page created for you.

src/main/webapp/scripts/pages/workspace/editor/public_documentView.js
rs-nicof: > Written by Claude, an AI coding agent.

Thanks. My original comment was not precise enough about where this has to happen: clearing the attributes in the browser tidies the rendered page, but the document content is written into the page by the server, so the address and the query are still visible in the page source or a plain fetch. That is how the existing link types already behave, so the team may be fine with it, but the query text is new and more sensitive than a URL.

One other spot while you are in here: the table produced by the insert-rows feature carries its own link, and it matches none of the selectors here, so it stays live and clickable on a published document.

rs-nicof (changes requested):

src/main/java/com/researchspace/webapp/integrations/dbrepo/DBRepoClient.java
akos326: Fixed

src/main/java/com/researchspace/webapp/integrations/dbrepo/DBRepoController.java
akos326: Fixed

src/main/webapp/scripts/externalTinymcePlugins/dbrepo/plugin.min.js
akos326: Fixed

src/main/webapp/ui/src/tinyMCE/dbrepo/DBRepo.tsx
akos326: done

src/main/java/com/researchspace/webapp/integrations/dbrepo/DBRepoController.java
akos326: done

src/main/webapp/scripts/pages/workspace/editor/public_documentView.js
akos326: done

PR #1070, opened 2026-08-27
RSDEV-1308-fieldmark-bug

nhanlon2:
Fieldmark changed its API (FAIMS3 v1.6.2 moved the notebook design into the project document and dropped the metadata DB), so `GET /api/inventory/v1/fieldmark/notebooks` came back with everything null except `name`/`status`, and the import dialog crashed on `metadata.project_id`.

## Changes

- Bump `fieldmark-java-client` 4.0.0 → 4.1.0, which maps the new notebook-list response (`_id`, fallbacks for `projectId`/`metadata`/`uiSpecification`).
- `FieldmarkRealConnectionMVCIT` now runs nightly (`ConditionalTestRunner` + `-Dnightly`) against the real Fieldmark service, takes its bearer token from a `FIELDMARK_TOKEN` env var, skips itself if the token is absent, and asserts the notebook-list fields non-null field by field. `listing_id` and `ui-specification` are not asserted: the current API has no listing id and omits the design from list responses.
- Guard `DatabaseCleaner.cleanUp()` against a null `jdbcTemplate`: when a conditional test class is fully skipped no Spring context ever loads, and the `@AfterClass` cleanup NPE'd.

Skipped: frontend fix for the dialog crash and adapter DTO rework — separate change, tracked under RSDEV-1308.

CI note: the nightly Jenkins job needs `FIELDMARK_TOKEN` in its environment or these tests report as skipped.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1724,7 +1724,7 @@
     <dependency>
       <groupId>com.github.rspace-os</groupId>
       <artifactId>fieldmark-java-client</artifactId>
-      <version>4.0.0</version>
+      <version>4.1.0</version>
     </dependency>
     <dependency>
       <groupId>com.github.rspace-os</groupId>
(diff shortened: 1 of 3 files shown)


nhanlon2:
Recreating with the branch on rspace-os instead of the fork.

PR #1073, opened 2026-08-28
Bump frontend deps (aug 26)

tchoi-rs:
## Description ##
Bumps all frontend dependencies. Also removes `@codemirror/state` as it's not being directly used in the codebase.


mKowalski256 (approved):
Looks good to me, tested on [AWS build](https://t3code-audit-frontend-deps-54722ce8-3.researchspace.com/) and I didn't notice any problems. 

I confirmed upload of image/file attachments in tinMCE working, so failing test may be just a CI issue.

PR #1074, opened 2026-08-29
Re-enable previously disabled tests

tchoi-rs:
## Description ##
Based on the previous work to stablise the test infrastructure and provide e2e mocks, re-enable previously disabled tests. This PR also provides a Chemistry test harness for Chemistry-related tests.


src/test/java/com/researchspace/service/ConcurrentSearchIT.java
```diff
@@ -0,0 +1,80 @@
...
+  @Test
+  public void concurrentSearchesSeeCommittedFixtures() throws Exception {
+    int searchCount = 2;
+    User[] users = new User[searchCount];
+    String[] terms = new String[searchCount];
+    Long[] documentIds = new Long[searchCount];
+    for (int i = 0; i < searchCount; i++) {
+      users[i] = doCreateAndInitUser(getRandomAlphabeticString("search"));
+      terms[i] = CoreTestUtils.getRandomName(20);
+      documentIds[i] = createBasicDocumentInRootFolderWithText(users[i], terms[i]).getId();
```
rs-nicof, line 37: This creates a document for each user without logging in as that user first, so it fails in CI with an authorisation error before the concurrency part is even reached. The base class does the login first everywhere else it creates a document.

```suggestion
      logoutAndLoginAs(users[i]);
      documentIds[i] = createBasicDocumentInRootFolderWithText(users[i], terms[i]).getId();
```

src/test/java/com/researchspace/webapp/controller/WorkspaceControllerMVCIT.java
```diff
@@ -995,28 +988,25 @@ public void testMVCGallerySearch() throws Exception {
...
 
-  @Disabled
   @Test
   public void testMVCAttachmentSearch() throws Exception {
-    final int SleepTime = 50;
     User anyUser = createAndSaveUser(getRandomName(10));
     setUpUserWithoutCustomContent(anyUser);
     logoutAndLoginAs(anyUser);
-    // initialiseFileIndexer();
+    initialiseFileIndexer();
```
rs-nicof: This is the only place in the suite that starts the file indexer without first pointing it at a temp folder, and starting it wipes the index and opens a writer that nothing closes. In CI this test now fails on the leftover lock, and against a real filestore it would delete the attachment search index. Pin it to a temp folder like the other callers do, and close it afterwards.

src/test/java/com/researchspace/webapp/controller/WorkspaceControllerMVCIT.java
```diff
@@ -406,49 +409,39 @@ public void simpleDeleteRecordTest() throws Exception {
...
-        srchInput,
+        searchInput,
         request,
         mockPrincipal,
         session,
         response);
     RSForm form = createAnyForm(piUser);
-    for (int i = 0; i < 1000; i++) { // will take several minutes to run
-      StructuredDocument doc = createDocumentInFolder(root, form, piUser);
+    for (int i = 0; i < 1000; i++) {
```
rs-nicof: The comment you removed here said this takes several minutes, and it now runs in the integration job on two database versions. It also commits all 1000 documents for real with no cleanup, so they and their audit history stay in whatever database it ran against. The new assertion holds just as well with a handful of documents.

rs-nicof (commented):
Ran a parallel review over the whole diff (layering, security, persistence and a general quality pass), keeping only the higher-severity findings. Ten comments below.

The main thing though is the integration suite. Comparing this head against the PR's own merge base:

| | base `c17a4ee27` | head `5729842a` |
|---|---|---|
| Result | pass | fail |
| Passing | 1016 | 798 |
| Errors | 0 | 213 |
| Suites | 172 | 165 |
| Duration | ~12 min | ~20.5 min |

The 213 errors look like one cascade rather than 213 separate problems: almost all of them report the same `Row was updated or deleted by another transaction ... Container#1915`, and they hit 31 classes, most of which this PR does not touch. Two of the newly enabled tests fail on their own merits first (the two comments below on the concurrent search test and the attachment search test), so those look like the place to start.

src/test/java/com/researchspace/service/SearchPerformanceTest.java
```diff
@@ -73,63 +69,7 @@ public void setUp() throws Exception {
...
-      Thread t = new Thread(searcher);
-      t.start();
-      t.join(); // wait for thread to finish before exiting
-      //	searcher.run();
-    }
-  }
-
   @Test
-  @EnabledIfSystemProperty(named = "nightly", matches = "(|true)")
   public void testPerformance()
```
rs-nicof: All four nightly gates were removed from this class with nothing replacing them, so this now runs on every Java PR: 50 users, 500 documents, and an assertion that compares two wall-clock timings. Timing assertions on a shared CI runner fail for reasons unrelated to the change being tested. Please restore a class-level gate, or tag it the way chemistry is tagged.

src/test/java/com/researchspace/api/v1/controller/DocumentsApiControllerTest.java
```diff
@@ -188,8 +192,9 @@ public void testGetAllDocuments() throws Exception {
 
     List<ApiDocumentInfo> twoLastModifiedInfoList = twoLastModifiedResults.getDocuments();
     assertEquals(2, twoLastModifiedInfoList.size());
-    apiModelTestUtils.assertApiDocumentInfoMatchSDoc(twoLastModifiedInfoList.get(0), firstDoc);
-    apiModelTestUtils.assertApiDocumentInfoMatchSDoc(twoLastModifiedInfoList.get(1), secondDoc);
+    assertTrue(
```
rs-nicof: These documents are created back to back and can easily share a millisecond, and once they do a <= check passes whichever order the API returned. The sort-order half of this test can no longer fail. Comparing document ids in the expected order would keep it honest.

src/test/java/com/researchspace/service/inventory/SubSampleApiManagerTest.java
```diff
@@ -214,18 +207,23 @@ public void groupOwnedSubSampleVisibilityInsideAndOutsideGroup() {
...
     assertEquals(piSubSample.getName(), retrievedPiSubSample.getName());
     // pi can query full details of testUsers's subsample
     ApiSubSample fullTestUserSubSample =
         subSampleApiMgr.getApiSubSampleById(retrievedTestUserSubSample.getId(), pi);
     assertEquals(testUserSubSample.getName(), fullTestUserSubSample.getName());
     // pi have update and transfer permission to user's subsample
-    System.out.println(
-        " ### fullTestUserSubSample ### : " + fullTestUserSubSample.getPermittedActions());
-    assertEquals(3, fullTestUserSubSample.getPermittedActions().size());
     assertTrue(fullTestUserSubSample.getPermittedActions().contains(UPDATE));
```
rs-nicof, line 228: Dropping the exact count here means an accidental extra permission for a PI over a member's subsample would now slip through unnoticed. The matching block just above handles this well by asserting explicitly that CHANGE_OWNER is absent.

Jenkinsfile
```diff
@@ -286,6 +287,49 @@ pipeline {
...
+                failure {
+                    notify currentBuild.result
+                    notifySlack('FAILURE', 'Chemistry tests failed')
+                }
+                fixed {
+                    notify currentBuild.result
+                    notifySlack('SUCCESS', 'Chemistry tests fixed')
+                }
+                success {
+                    junit 'target/surefire-reports/*.xml'
```
rs-nicof: Publishing the test report only on success means that when a chemistry test fails, the stage goes red and Slack fires but nobody can see which test broke. Moving this into the always block would fix it, as the GitHub job already does.

src/test/java/com/researchspace/webapp/integrations/fieldmark/FieldmarkRealConnectionMVCIT.java
```diff
@@ -28,10 +28,10 @@
 import org.springframework.test.web.servlet.MvcResult;
 
 /**
- * Runs only on nightly builds (-Dnightly) against the real Fieldmark service; requires the
- * FIELDMARK_TOKEN environment variable to hold a valid bearer token.
+ * Runs against the real Fieldmark service when explicitly enabled; requires the FIELDMARK_TOKEN
+ * environment variable to hold a valid bearer token.
  */
-@EnabledIfSystemProperty(named = "nightly", matches = "(|true)")
+@EnabledIfSystemProperty(named = "fieldmark.realConnectionTests", matches = "true")
```
rs-nicof: This class used to run on nightly builds. The new property is not set anywhere in the repo, so it now runs nowhere at all. The switch makes sense for the classes that were disabled before, but this one loses real coverage unless the nightly job is updated to set it.

src/test/java/com/researchspace/webapp/controller/RSChemControllerMVCIT.java
```diff
@@ -50,15 +50,9 @@
 import org.springframework.test.web.servlet.MvcResult;
 import org.springframework.test.web.servlet.request.MockHttpServletRequestBuilder;
 
-@Disabled(
-    "Requires chemistry service to run. See"
-        + " https://documentation.researchspace.com/article/1jbygguzoa")
+@Tag("chemistry")
```
rs-nicof, line 52: Three test methods in this class still carry a nightly gate, and neither the chemistry script nor the new job sets that property. The chemistry job will report this class green while those three never run, which is the situation this PR is trying to fix.

.github/workflows/lint-and-test.yml
```diff
@@ -757,6 +757,90 @@ jobs:
...
+          - 3306:3306
+        options: >-
+          --tmpfs /var/lib/mysql:rw,noexec,nosuid,size=2g
+          --health-cmd="healthcheck.sh --connect --innodb_initialized"
+          --health-interval=5s
+          --health-timeout=5s
+          --health-retries=40
+          --health-start-period=30s
+      chemistry:
+        image: rspaceops/oss-chemistry:latest
```
rs-nicof, line 791: Pinning the chemistry image to latest here and in the Jenkinsfile makes the job non-reproducible, so an upstream image change shows up as a mysterious code regression, and the two CI systems can end up testing different builds. Worth pinning a digest or version, as mariadb already is.

src/test/java/com/researchspace/webapp/controller/StoichiometryControllerMVCIT.java
```diff
@@ -29,34 +33,43 @@
...
 import org.springframework.test.web.servlet.MvcResult;
 
 @WebAppConfiguration
-@TestPropertySource(
-    properties = {"chemistry.service.url=http://localhost:8090", "chemistry.provider=indigo"})
-@Disabled(
-    "Requires chemistry service to run. See"
-        + " https://documentation.researchspace.com/article/1jbygguzoa")
+@TestPropertySource(properties = "chemistry.provider=indigo")
+@TestExecutionListeners({
```
rs-nicof: Nice to have, not a blocker. This works today only because the parent class happens to declare its own listeners, so these two get appended instead of replacing the defaults. One side effect is that the bean override listener ends up running after dependency injection rather than before it. Since this class only runs in the chemistry job, if that parent declaration ever changes the breakage would not show up in the main build. Declaring the two listeners on the parent, or setting `mergeMode`, would remove the dependency on that coincidence.

rs-nicof (changes requested):

.github/workflows/lint-and-test.yml
tchoi-rs: Not an issue

src/test/java/com/researchspace/webapp/controller/RSChemControllerMVCIT.java
tchoi-rs: Removed the nightly gates in the functions so this now runs as one

src/test/java/com/researchspace/service/inventory/SubSampleApiManagerTest.java
tchoi-rs: Fixed now

rs-nicof (approved):

PR #1078, opened 2026-08-31
RSDEV-1251: Update external PIDINST when amending Instrument

rs-nicof:
## Description ##

The rspace-web implementation of **RSDEV-1251** ("[Backend] Add *update external PIDINST* from RSpace"), a sub-task of the RSDEV-1033 PIDINST story.

Once an instrument has a registered PIDINST, nothing sent its later edits onward. A new model number, a decommissioning date, a corrected owner: the registered record at B2INST or DataCite quietly drifted away from the instrument it is supposed to describe, and there was no way for a user to resynchronise it.

**The save is the trigger.** Saving an instrument that carries a PIDINST in a still-writable state re-runs the usual PIDINST mapping over the instrument's current fields and sends the result to the provider. No new endpoint, no new button: saving *is* the explicit user action the ticket asks for.

The two providers state the rule from opposite ends, deliberately:
- **B2INST** pushes in every state **except** `accepted`. Every other review state (`draft`, `created`, `submitted`, `cancelled`, `declined`, `expired`) still has a writable draft behind it, and a state wrongly skipped is silent drift. Acceptance publishes the record and removes its draft, so there is nothing left to write.
- **DataCite** pushes **only** `draft`. A findable DOI is refreshed by the existing Republish, and a retracted (`registered`) DOI is left alone by decision rather than by provider refusal.

Both rules were established by probing the real providers (b2inst-test.gwdg.de and api.test.datacite.org, August 2026) rather than from documentation; the probes corrected the original plan on two points, recorded in ADR 0008.

### Scope note ###

This is the **backend** half. The outcome of each push is returned per identifier in the API response, and surfacing it in the Inventory UI is separate, still-outstanding work with no ticket yet. Today an API client sees the reason and a user of the web interface sees an ordinary successful save. `CONTEXT.md` says so explicitly rather than implying the loop is closed.

### Backend changes ###

- **`InventoryIdentifierExternalUpdateService`** (new). Orchestrates the push: works out which identifiers qualify, rebuilds each payload, sends it, records the outcome and audits the attempt. Deliberately **not** a `*Manager`, because the `*..service.inventory.*Manager` pointcut would wrap every method in a transaction and the whole point is that the provider call happens outside one.
- **`InstrumentsApiController`**. Two seams call the push after the manager returns and the edit has committed: `updateInstrument` and `changeInstrumentOwner`. Both are guarded, so a provider outage cannot turn a saved instrument into an error response.
- **`ApiInventoryDOI`**. Adds the read-only `externalMetadataUpdate` outcome object, and makes an existing identifier's `doi` immutable (see Design decisions).
- **`RspaceToExternalProviderAdapterImpl`**. The DataCite title now comes from the instrument's current name. It was written once at register or assign and never refreshed, so a rename shipped the old name on every later push.
- **`B2instConnector(Impl)`**. New `updateDraftDoi(rid, doi)`, a full-replace `PUT /api/records/{rid}/draft`. Reuses the existing record-id allow-list and single-exit token redaction.
- **`DataCiteConnector(Impl)`**. New `updateDoi(doi, settingType)`, which sends no `event` so a draft stays a draft.
- **`B2instConnectionException`**. A transport failure's user-facing reason no longer carries Spring's message, which named the deployment's B2INST host and reached the audit trail. The developer detail is unchanged and now logged.
- **`DigitalObjectIdentifierDaoHibernate` / `DigitalObjectIdentifier`**. The public landing page now serves the newest identifier revision for both providers (see Design decisions), the revision query is bounded, and the published-state check is case-insensitive.

### Frontend changes ###

Almost none, by design. Three files:
- **`server.inventory.json`** and the regenerated **`resources.d.ts`**: four backend message keys (`externalUpdated`, `externalUpdateFailed`, `externalUpdateNotPossibleB2inst`, `externalUpdateNotPossibleDataCite`). Backend user-facing text for this area lives in the frontend JSON catalogues, resolved by `JsonMessageSource`, not in `src/main/resources/bundles/`.
- **`__tests__/e2e/mocks/b2inst.ts`**: an MSW handler for `PUT /api/records/:rid/draft`. Without it every mock-mode e2e instrument save carrying a B2INST draft got a 404 and silently recorded a failed push, which no assertion caught because the toast is generic.

### API / Swagger ###

No breaking change. `PUT /instruments/{id}` and `/instruments/{id}/actions/changeOwner` gain a read-only `externalMetadataUpdate` object on each identifier in the response:

| Property | Explanation |
| --- | --- |
`succeeded` | `true` when the provider accepted the metadata
`reason` | A sentence for the user: what was updated, or why it was not, including the provider's own wording for a rejection

Nothing about the attempt is persisted, so a later read does not carry it. The spec documents exactly when the object is absent, including the case where the caller's own view of the record has been permission-filtered.

## Related work ##

- Parent story: RSDEV-1033 (lookup and creation of instruments from existing PIDs).
- **Companion PR, required**: https://github.com/rspace-os/datacite-java-client/pull/7. It carries `@JsonInclude(NON_NULL)` on `DataCiteDoiAttributes`, which this branch depends on for correctness, plus hardening of the DOI id in the request path.
- Builds on RSDEV-1253 (related identifiers, ADR 0007, amended here) and RSDEV-1260 (B2INST status refresh), whose per-provider public-page rule this deliberately reverses.

## Jira tickets ##

- https://researchspace.atlassian.net/browse/RSDEV-1251

## Design decisions ##

Recorded in full in `DevDocs/adr/0008-external-pidinst-update-on-save.md`.

- **The push never blocks the save.** The instrument edit commits first; the provider call runs afterwards, outside any transaction, in the same request. A failed push changes nothing locally: no rollback, no compensation, identifier state untouched. This avoids extending the existing pattern of provider HTTP calls made inside a write transaction, which the RSDEV-1260 review already flagged.
- **The push declines when a transaction is already open.** It is reachable with one, through bulk `CHANGE_OWNER` with rollback-on-error. Pushing there fails in the one direction always-push cannot heal: a later record fails, the owner change rolls back, and the provider keeps an owner RSpace no longer has.
- **Every qualifying save pushes; no diffing, no persisted sync state.** Both provider updates are idempotent full-metadata writes, so a redundant push is harmless and always-push self-heals: retry is "save again". A diff-gated push would leave the provider stale forever if a push failed and the next save changed no mapped field.
- **Candidates come from the record, not from the response.** The response DTO is permission-filtered, and an owner transfer leaves the departing owner with a limited view carrying no identifiers at all, so a response-driven push skipped silently. A non-empty list is still trusted as a fast path, because filtering blanks the list rather than trimming it.
- **A frozen record is explained, a disabled provider is silent.** "Nothing was sent, and here is why, naming the provider" is what the acceptance criteria ask for. A switched-off integration says nothing, because there is nothing the user can act on and this rides on an ordinary save.
- **An existing identifier's provider record id is immutable.** That value is the address of the record at the provider. It was client-writable, and this feature would have turned it into an outbound write primitive: the existing id check stops a client naming *another* record's identifier row, but not one retargeting *its own* row at a foreign RID or DOI. It is now applied only while the identifier is still being created, mirroring the `doiType` guard beside it. `Access.READ_ONLY` was tried first and rejected, because it also stops a Java client reading the value back out of a response.
- **The public landing page serves the newest identifier revision for both providers.** DataCite previously held the page at the publication-time snapshot. The consequence is deliberate and documented in `CONTEXT.md`: an identifier-row change reaches the page without a republish, and record edits made after publishing become public as soon as anything writes that row.

## Testing notes ##

- **Backend (fast unit)**: `mvn test -Dfast=true`, 3966 green. Covers the full type x state eligibility matrix, the enablement gate, the frozen-state explanation, per-identifier isolation of both push and mapping failures, the transaction definition the service demands, the `doi` immutability guard in both directions, the DataCite title source, and the B2INST connector's URL, method, auth header, body and every failure shape.
- Two tests pin invariants the design rests on rather than its behaviour: `ApiInstrumentLimitedViewTest` (a filtered view blanks the identifier list rather than trimming it, which is what makes the fast path safe) and `ExternalUpdateMessageKeysTest` (all four message codes resolve; `useCodeAsDefaultMessage` is true, so a typo would render the bare key to the user and into the audit trail).
- **Backend (MVCIT, real DB commits, `mvn verify -Denvironment=drop-recreate-db`)**: `InstrumentExternalMetadataUpdateMVCIT` covers the end-to-end push, the failure path leaving the instrument saved, the owner transfer reaching the provider, that register and publish do **not** push, and, most importantly, that the provider call happens with no transaction open. Both new MVCITs capture and restore the shared dev database's `pidinst.*` settings for **both** providers, because enabling one PIDINST provider disables the other.
- **Nightly**: `InstrumentExternalMetadataUpdateDataCiteMVCIT` runs against the real api.test.datacite.org under `-Dnightly`, minting and deleting a real test DOI. It is the only test that sends the actual serialized payload to a provider.
- **Frontend**: `pnpm tsc`, `pnpm lint`, `pnpm run i18n:types`, `pnpm run i18n:lint`, all clean.

### Known gaps ###

- No frontend consumer of `externalMetadataUpdate` yet; it will be handled by [RSDEV-1325](https://researchspace.atlassian.net/browse/RSDEV-1325)


rs-nicof:
@greptileai

pom.xml
```diff
@@ -1779,7 +1779,7 @@
     <dependency>
       <groupId>com.github.rspace-os</groupId>
       <artifactId>datacite-java-client</artifactId>
-      <version>1.1.0</version>
+      <version>RSDEV-1251-datacite-SNAPSHOT</version>
     </dependency>
```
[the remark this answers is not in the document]
rs-nicof: Correct, and it is the one thing blocking this PR: it is called out under "Merge blocker" in the description and in ADR 0008. Leaving it open deliberately rather than fixing it now.

The pin cannot move yet because the version it must point at does not exist: the companion PR (rspace-os/datacite-java-client#7) carries the `@JsonInclude(NON_NULL)` fix this branch depends on for correctness, and reverting to the released `1.1.0` would silently reintroduce the data-loss bug it fixes (verified by unzipping both jars: `1.1.0` has `updateDoi`, so it would compile, but not the annotation). The order is: bump the client pom to `1.2.0`, merge and tag it, then swap this pin to `1.2.0`.

This thread stays open until that swap lands.

rs-nicof:
AWS instance: https://rsdev-1251-update-external-pidinst-93004725-21.researchspace.com

src/main/java/com/researchspace/api/v1/controller/InstrumentsApiController.java
```diff
@@ -203,6 +230,11 @@ public ApiInstrument changeInstrumentOwner(
     instrumentApiMgr.assertUserCanTransferInstrument(id, user);
 
     ApiInstrument updated = instrumentApiMgr.changeApiInstrumentOwner(incomingInstrument, user);
+    // A transfer really does change what the provider holds: RspaceToExternalProviderAdapterImpl
+    // maps ownerContact from the record owner's email unconditionally, and ownerName from the same
+    // owner unless the Owner field overrides it. Left unpushed, the registered record keeps the
+    // previous owner's address, which is exactly the drift this ticket exists to stop.
+    pushExternalMetadataUpdates(updated, user);
```
nhanlon2, line 243: This comment is from Claude. The owner-transfer push looks unreachable from the product. Search.ts:887 is the only UI transfer path and it hardcodes rollbackOnError=true (it is the third parameter of ApiService.bulk), so InventoryBulkOperationHandler:334 re-enters this controller inside an open transaction, the guard at InventoryIdentifierExternalUpdateService:166 declines, and nothing is pushed. Nothing in the UI calls the single-record changeOwner endpoint. After a UI transfer the registered record would keep the previous owner indefinitely, which is the drift this ticket exists to stop. Relatedly, the javadoc at line 172 says the edit has committed by this point, which is not true on that path. An afterCommit synchronisation or @TransactionalEventListener in the manager would fire on every path including the bulk one, and would make the guard unnecessary.

nhanlon2 (commented):
This comment is from Claude. Automated review of RSDEV-1251, run by six reviewers in parallel (layering, security, persistence, i18n, frontend, general quality), then verified against the code before posting.

The central design holds up: three reviewers independently confirmed the pointcut reasoning, that the service really does run unadvised, and that the provider call happens outside a transaction. `theProviderCallRunsWithNoTransactionOpen` pinning that as a real test is the best thing in the PR.

Nine comments below. The two worth reading first are the owner-transfer path, which looks unreachable from the product, and the public landing page visibility change, which reverses an RSDEV-1260 decision.

Not commented on separately because the PR already declares it: the `datacite-java-client` pin to `RSDEV-1251-datacite-SNAPSHOT` at pom.xml:1788 is confirmed present and is a genuine merge blocker. Worth stressing it is not cosmetic, since the companion `@JsonInclude(NON_NULL)` change alters serialization of every DataCite payload, IGSN samples included, not just the new update path.

src/main/java/com/researchspace/api/v1/controller/InstrumentsApiController.java
```diff
@@ -157,10 +159,35 @@ public ApiInstrument updateInstrument(
     instrumentApiMgr.assertUserCanEditInstrument(id, user);
 
     ApiInstrument updated = instrumentApiMgr.updateApiInstrument(incomingInstrument, user);
+    pushExternalMetadataUpdates(updated, user);
```
nhanlon2, line 162: This comment is from Claude. This makes an ordinary instrument save wait on a provider HTTP call. B2instConnectorImpl bounds that at 10s connect + 30s read, so a rename can block for roughly 40 seconds when B2INST is slow, and the user is shown nothing explaining the delay. DataCiteConnectorImpl sets no timeouts in this repo at all, so ADR 0008's 'bounded by the existing connector timeouts' is only verified for one of the two providers, and the DataCite bound currently lives inside the branch SNAPSHOT.

src/main/java/com/researchspace/api/v1/model/ApiInventoryDOI.java
```diff
@@ -261,6 +295,22 @@ public void generatePublicLinkSuffix() {
...
+   * metadata still compare equal. {@code NON_NULL} keeps the property out of the payload entirely
+   * when no push was attempted, which is how a client tells "not eligible" from "attempted".
+   * Deliberately absent from {@link #applyChangesToDatabaseDOI(DigitalObjectIdentifier)}: there is
+   * no column for it, by decision (ADR 0008 item 4).
+   */
+  @JsonProperty(value = "externalMetadataUpdate", access = JsonProperty.Access.READ_ONLY)
+  @JsonInclude(JsonInclude.Include.NON_NULL)
+  @EqualsAndHashCode.Exclude
+  @ToString.Exclude
+  private ApiExternalMetadataUpdate externalMetadataUpdate;
```
nhanlon2, line 312: This comment is from Claude. Two things about this field. First, nothing under src/main/webapp/ui/src reads it, so a web user still sees a plain success toast whether the push worked, was refused by a frozen record, or failed at the provider; the ticket's first three acceptance criteria are met for API clients only. Second, and much cheaper to fix now while nothing consumes it: a retryable provider failure, a permanent and benign 'the record has no draft any more', and a mapping failure all come back as succeeded:false with only a pre-localized sentence to tell them apart. A future consumer would have to string-match translated prose to avoid showing a red error for entirely normal behaviour. An explicit machine-readable outcome code alongside succeeded would be backward-compatible.

src/main/java/com/researchspace/service/inventory/impl/RspaceToExternalProviderAdapterImpl.java
```diff
@@ -364,12 +364,32 @@ public DataCiteDoi buildDataCiteDoi(ApiInventoryDOI doi, InventoryRecord associa
...
+     * The instrument's CURRENT name, not the one stored on the identifier. That stored title is
+     * written once, by updateNewAssociatedDoi at register time and by ApiIdentifiersHelper on
+     * assign, and nothing refreshes it on a rename - so reading it here sent the old name on every
+     * later push and left the registered record drifting, which is the drift RSDEV-1251 exists to
+     * stop. Set before the environment guard below returns, because a deployment with no usable
+     * server URL still has a correct name to send.
+     */
+    dataCiteDoi
+        .getAttributes()
+        .setTitles(List.of(new DataCiteDoiAttributes.Title(instrument.getName())));
```
nhanlon2, line 377: This comment is from Claude. This unconditionally overwrites the DOI title with the instrument name, but title is still client-writable (ApiInventoryDOI.applyChangesToDatabaseDOI writes it, and this PR's own SampleApiManagerTest edits one through the API). Because the mapping is shared, a user who deliberately set a different DOI title now has it stored and displayed in RSpace but silently replaced on every DataCite write, publish and retract included, not only the new push. There is no test for 'user edits the identifier title, then publishes'. Either make title non-editable for instrument identifiers, or refresh the stored title on rename and keep reading it here.

src/main/java/com/researchspace/dao/hibernate/DigitalObjectIdentifierDaoHibernate.java
```diff
@@ -67,36 +68,24 @@ public Optional<DigitalObjectIdentifier> getLastPublishedIdentifierByPublicLink(
...
-    DigitalObjectIdentifier firstOfPublishedRun = newestDoi;
-    for (int i = 1; i < revisions.size(); i++) {
-      DigitalObjectIdentifier candidate = doiOf(revisions.get(i));
-      if (!DigitalObjectIdentifier.isPublishedState(candidate.getState())) {
-        break;
-      }
-      firstOfPublishedRun = candidate;
-    }
-    return Optional.of(firstOfPublishedRun);
+    return Optional.of(newestDoi);
```
nhanlon2, line 88: This comment is from Claude. This reverses the DataCite branch of the public-page revision rule that RSDEV-1260 deliberately introduced. Because Envers resolves the linked record as of the identifier revision, record edits made after publishing become publicly readable the moment anything writes the identifier row, including a customFieldsOnPublicPage toggle or a status refresh, neither of which reads to a user as 'publish my latest edits'. Nothing in RSDEV-1251 requires it; it rides along because setMaxResults(1) makes the old backwards walk impossible. The query bound is a clear fix, but the visibility change is a separate decision that deserves its own PR with the RSDEV-1260 owner in the loop, plus a release note rather than only CONTEXT.md.

src/main/java/com/researchspace/model/inventory/DigitalObjectIdentifier.java
```diff
@@ -110,7 +110,7 @@ public DigitalObjectIdentifier(String identifier, String title) {
    * record). Gates the public landing page, so both providers' published states must be here.
    */
   public static boolean isPublishedState(String state) {
-    return "findable".equals(state) || "accepted".equals(state);
+    return "findable".equalsIgnoreCase(state) || "accepted".equalsIgnoreCase(state);
```
nhanlon2, line 113: This comment is from Claude. Making only this check case-insensitive puts it out of step with the state comparisons guarding the same data: InventoryIdentifierApiManagerImpl:723 still does a case-sensitive equals on 'accepted', and canBeAssigned() at line 217 does the same on 'draft'. If a provider ever returned differently-cased text, the unauthenticated public page would open while the manager still treated the identifier as unpublished, so two parts of the system would disagree about whether a record is published. Latent rather than live, since InvenioRDM returns lowercase today. Worth either normalising the state on the way in, or applying the same rule everywhere.

src/main/java/com/researchspace/webapp/integrations/datacite/DataCiteConnectorImpl.java
```diff
@@ -137,4 +137,9 @@ public DataCiteDoi publishDoi(DataCiteDoi dataCiteDoi, InventorySettingType sett
   public DataCiteDoi retractDoi(DataCiteDoi dataCiteDoi, InventorySettingType settingType) {
     return getClient(settingType).retractDoi(dataCiteDoi);
   }
+
+  @Override
+  public DataCiteDoi updateDoi(DataCiteDoi dataCiteDoi, InventorySettingType settingType) {
+    return getClient(settingType).updateDoi(dataCiteDoi);
```
nhanlon2, line 143: This comment is from Claude. This hands the stored identifier straight to the client with no validation. The B2INST side deliberately does the opposite: B2instConnectorImpl.recordUrl validates against VALID_RID and escapes it as a single path segment, with a javadoc calling the stored identifier untrusted precisely because a record update is what writes it. The DataCite client concatenates this value into the request path, on a call carrying the deployment's DataCite credentials. The new immutability guard makes the value server-set in practice, so this is defence in depth rather than a live issue, but the asymmetry is worth closing while both connectors are open.

src/main/java/com/researchspace/webapp/integrations/b2inst/B2instConnectorImpl.java
```diff
@@ -216,12 +252,25 @@ private String describeFailureText(RestClientException e) {
...
+     * RSpace's own fixed sentence, not the exception's message. Spring's message for a transport
+     * failure carries the full request URL, the deployment's B2INST host included, and the fallback
+     * when it was blank was the exception's class name ("ResourceAccessException"). Both reached the
+     * user through the reason and were written into the audit trail, and neither is anything a user
+     * can act on. The detail belongs in the log, which is where it now goes - it was not logged at
+     * all before.
+     */
+    log.warn(
+        "Could not reach B2INST: {}", redactToken(StringUtils.defaultString(e.getMessage())), e);
+    return "B2INST could not be reached.";
```
nhanlon2: This comment is from Claude. This hardcoded sentence becomes the exception's reason, which InventoryIdentifierExternalUpdateService then interpolates into the localized errors.inventory.identifier.externalUpdateFailed and writes verbatim into the audit trail. Once a non-English locale ships, a user hitting a network failure gets a half-translated sentence, and the audit record for that event is in a different language from every other audit record. Resolving it from server.inventory.json, or having the exception carry the failure kind as data and letting the service pick the key, would fix both.

CONTEXT.md
```diff
@@ -121,6 +121,57 @@ resolved during design. This file is a glossary only — no implementation detai
...
+  carries an identifier in an updatable identifier state; saving *is* the
+  explicit user action, there is no separate "update external PIDINST" button.
+  A failed push never blocks or reverts the save: the instrument edit is
+  persisted regardless, the identifier's own state is left untouched, and the
+  failure is reported on the identifier in the API response. Surfacing that in
+  the Inventory UI is a separate, still-outstanding piece of work, so today an
+  API client sees the reason and a user of the web interface does not. Every
+  qualifying save pushes, whether or not a mapped field changed, so retrying a
+  failed push is just saving again. An identifier whose state has frozen its
+  provider record is reported too, naming the state, rather than passed over in
```
nhanlon2: This comment is from Claude. This says a frozen identifier is reported 'naming the state', and rspace_api_inventory_specs_2_26_0.yaml:163 says the same. The code deliberately does the opposite: the javadoc on notWritableOutcome says the state is deliberately not interpolated, the messages carry no state token, and the unit test asserts the state is absent. Since this is the behaviour acceptance criterion 2 turns on, worth correcting the prose in both documents.

src/main/java/com/researchspace/api/v1/controller/InstrumentsApiController.java
rs-nicof: Confirmed and fixed. You were right on every step: the web interface only ever transfers instruments through the bulk path with rollbackOnError, so the guard declined the push every time and no transfer from the UI ever reached a provider.

The fix pushes from the bulk controller once the batch's transaction has committed, instead of from inside it. That keeps the property the guard exists for, a reverted batch sends nothing so the provider is never left ahead of RSpace, while making the transfer path actually work. Four unit tests cover it: a committed transfer and a committed bulk edit push, a reverted batch and an operation that changes no provider metadata do not. The javadoc claim you flagged about the edit having committed is corrected too.

src/main/java/com/researchspace/api/v1/controller/InstrumentsApiController.java
rs-nicof: Half fixed. You were right that the DataCite side was unverified: the client set no timeouts at all, so an unresponsive DataCite could hold a save open indefinitely. It now bounds both at the same 10s connect and 30s read as B2INST, so ADR 0008's claim is true for both providers. That change is in the companion PR, rspace-os/datacite-java-client#7.

Not fixed: the user being shown nothing during the wait. That needs the Inventory UI work already outstanding for this feature, so it belongs [with that ticket](https://researchspace.atlassian.net/browse/RSDEV-1356) rather than here.

src/main/java/com/researchspace/api/v1/model/ApiInventoryDOI.java
rs-nicof: Both fair, both deliberately left for the UI ticket.

The frontend gap is known and recorded: nothing consumes the field yet and surfacing it is outstanding work, so the acceptance criteria are met for API clients today.

On the machine-readable code, I agree it beats string-matching translated prose. But adding a field alongside succeeded is backward compatible whenever it happens, so deferring costs nothing, and the UI ticket is what will say which distinctions actually need telling apart. Doing it now would be guessing at that.

src/main/java/com/researchspace/service/inventory/impl/RspaceToExternalProviderAdapterImpl.java
rs-nicof: Real, and left as a product decision rather than settled here.

Sending the instrument's current name is deliberate: the stored title is written once and never refreshed, so reading it sent the old name after every rename, which is the drift this ticket exists to stop. But you are right about the consequence, that a user-set DOI title is something RSpace now shows and does not honour, and that it affects publish and retract rather than only the new push.

Your first option looks like the right answer, making title non-editable for instrument identifiers, since for a PIDINST record the title is the instrument name by definition. That is an API contract change, so I have flagged it for the PR author instead of deciding it in a review fix.

src/main/java/com/researchspace/dao/hibernate/DigitalObjectIdentifierDaoHibernate.java
rs-nicof: This one is deliberate rather than a ride-along: the PR author asked for the DataCite branch to be collapsed onto the newest revision, so I have not reverted it.

Your process points are well taken and passed on: the RSDEV-1260 owner should see it, and a user-visible visibility change deserves a release note rather than only CONTEXT.md. There is no open release-note section for the next version in the repo yet, so that has to be added at release time.

src/main/java/com/researchspace/model/inventory/DigitalObjectIdentifier.java
rs-nicof: Fixed, and it turned out to be slightly less latent than it looked. You were right that the permissive check was the odd one out, and right about the direction that matters.

The B2INST refresh stores a non-accepted review status verbatim, so a provider-cased "Accepted" would have gone into the state column, which the public page then treats as published while RSpace never filled in a resolvable PID. That is the disagreement you describe, reachable through the one path that does not write RSpace's own literal. The three remaining comparisons on the same data now follow the same rule as the page, so a differently-cased status is recognised and normalised instead.

No visible change today, since InvenioRDM returns lowercase.

src/main/java/com/researchspace/webapp/integrations/datacite/DataCiteConnectorImpl.java
rs-nicof: Already closed, just not in this repo. The companion PR (rspace-os/datacite-java-client#7) hardened the client itself: it now refuses a DOI id that could change the endpoint before building the request path.

So the validation you are asking for does happen, one layer further down, at the point that actually assembles the URL, which means it protects every caller rather than this one connector. The asymmetry with B2INST is real, but it is now a difference in where the check lives rather than whether there is one.

src/main/java/com/researchspace/webapp/integrations/b2inst/B2instConnectorImpl.java
rs-nicof: Fixed. The exception's javadoc had recorded this as an accepted trade-off, but the repo's own rule is that anything reaching a user gets externalized, so the trade-off does not really stand.

The reasons RSpace authors now come from the message catalogue instead of being English literals in the connector, while the developer-facing message keeps its fixed English detail for the log. I did the two you flagged plus the two equivalents in the publish path, since fixing half the class would have been worse than either.

One part left open: the audit trail still stores the localized sentence. Making audit records locale-neutral is a separate design question about what the audit should hold, and not one to answer inside a review fix.

CONTEXT.md
rs-nicof: Fixed in both. You are right: the messages explain why in words and deliberately do not echo the provider's state token, and a unit test asserts its absence. CONTEXT.md and the API spec now say that instead.

src/main/java/com/researchspace/api/v1/model/ApiInventoryDOI.java
rs-nicof: Both halves now have a ticket: RSDEV-1356 covers surfacing the outcome in the Inventory UI, and it is the ticket that decides which distinctions the UI needs, which then settles whether the machine-readable outcome gets added. RSDEV-1355 covers the identifier title point from the other thread.

src/main/java/com/researchspace/service/inventory/impl/RspaceToExternalProviderAdapterImpl.java
rs-nicof: created https://researchspace.atlassian.net/browse/RSDEV-1355 for this, at leat we don't lose the sight.
I will need to doublecheck with Tilo if it is something we can do before I work on it.

src/main/java/com/researchspace/api/v1/model/ApiInventoryDOI.java
rs-nicof: created https://researchspace.atlassian.net/browse/RSDEV-1356 for this, at leat we don't lose the sight.
I will need to doublecheck with Tilo if it is something we can do before I work on it.

nhanlon2 (approved):

pom.xml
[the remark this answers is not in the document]
rs-nicof: The swap has landed, so this is now done. The pin is 1.2.0, and the client release contains the null-omission fix this branch depends on, along with the DOI path validation and the bounded HTTP timeouts. Verified that Maven resolves it from JitPack rather than from a local build, so the reproducibility point you raised is satisfied.

src/test/java/com/researchspace/service/inventory/ExternalUpdateMessageKeysTest.java
```diff
@@ -0,0 +1,69 @@
...
+ * the bare code, and the user's reason - and the audit entry - would read
+ * "errors.inventory.identifier.externalUpdateFailed". The service's own unit tests mock the message
+ * source and assert on the code that was requested, which pins the call site but cannot see whether
+ * the catalogue answers it.
+ *
+ * <p>Reads the catalogue file directly rather than through Spring, so it stays a fast unit test.
+ * {@code JsonMessageSource} flattens these files with dots and gives {@code server.*} an empty
+ * prefix, so a code is the flattened JSON path exactly as asserted here.
+ */
+class ExternalUpdateMessageKeysTest {
```
tchoi-rs, line 27: This is a useless test

tchoi-rs (changes requested):
Mocks look fine but I caught a useless test. Would recommend asking AI to cleanup all unnecessary tests and comments

src/test/java/com/researchspace/service/inventory/ExternalUpdateMessageKeysTest.java
rs-nicof: Fair, and it was weak in a specific way worth naming: it read the translation file directly and re-implemented how RSpace looks a message up, so what it really asserted was that the file still contained phrases someone had typed into the test. It also could not have caught the problem it was written for.

It now asks RSpace to produce each message the same way production does, and checks the two things a user would notice: that every message this feature can show exists at all, and that the ones naming a provider really do include the name. I confirmed it fails when a message goes missing, which is the case that matters, because a missing one reaches the user as a raw code instead of a sentence.

Same length, nothing new added, and it exercises the real lookup instead of restating it. Good catch.

tchoi-rs: No, I think the entire test should be deleted because it doesn't test anything useful

rs-nicof: I would keep it as claude explains it fails when a msg goes missing so I think it is worth to keep it

tchoi-rs (approved):

PR #1081, opened 2026-09-01
Fix Suspense boundary refresh in inventory, and fix duplicated API queries

tchoi-rs:
## Description ##
Fixes RSDEV-1350 's second case, where the first click in Gallery could cause the screen to refresh.

Also moves the API fetching logic to TanStack Query, in order to avoid duplicated API calls.


nhanlon2 (changes requested):
Clearing of screen on initial click on a gallery item when gallery is first loaded is the actual bug which is intended to be fixed. The selection clear seems to be a misunderstanding.

nhanlon2 (approved):

PR #1082, opened 2026-09-02
Prep for Mockito 5

tchoi-rs:
## Description ##
Migrate to Mockito 5, completely remove Hamcrest (AssertJ migration will happen in a downstream branch), and also simplified the Omero tests / HTTP mocking setup.


src/test/java/com/researchspace/webapp/controller/GlobalLookupControllerMVCIT.java
```diff
@@ -118,7 +116,7 @@ public void testGetRedirects() throws Exception {
   private void assertRedirect(GlobalIdentifier gid, String expectedURL) throws Exception {
     this.mockMvc
         .perform(get("/globalId/{oid}", gid.getIdString()))
-        .andExpect(view().name(containsString(expectedURL)))
+        .andExpect(viewNameContains(expectedURL))
```
nhanlon2, line 119: **(See edit at end of comment)** From Claude: **Critical.** This migration breaks `testGetRedirects`, and CI is red on it.

`GlobalLookupControllerMVCIT.java:46-51` asserts on the *text* of the failure message that `assertRedirect` throws:

```java
} catch (AssertionError ae) {
  assertTrue(
      ae.getMessage().contains("but: was \"error\""),
      "expected redirect to 'error', but msg was: " + ae.getMessage());
}
```

`but: was "error"` is Hamcrest's `MatcherAssert` output format. Now that this line uses `MVCTestBase.viewNameContains(...)`, the message is `"view name was " + name`, so that substring never appears and the assertion flips to false:

```
org.opentest4j.AssertionFailedError: expected redirect to 'error', but msg was: view name was error
  ==> expected: <true> but was: <false>
  at GlobalLookupControllerMVCIT.testGetRedirects(GlobalLookupControllerMVCIT.java:48)
```

Deterministic: fails on both MariaDB 10.11 and 12.3 on the first attempt. (The surefire rerun of the same test reports a `PessimisticLockingFailureException` instead, because by that point the shared DB is contended; the clean first execution on both legs shows the message-format mismatch.)

On `feature/minimal-assertthrows-executables` this test passes. Base run 34130682848, both legs:

```
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0 -- in GlobalLookupControllerMVCIT
```

So this is a clean pass to a deterministic failure, introduced by this branch.

Suggested fix: rather than updating the expected substring to `view name was error` (which just re-couples the test to an assertion message), assert the view name directly for this case, so the test no longer depends on any matcher's formatting.

---

*Edited 2026-09-11 to correct a factual error in the original version of this comment. It claimed this test also failed on the base branch, at line 47, "for an unrelated DB reason". That was wrong: the line-47 frame I had matched in the base logs is a `ControllerExceptionHandler.logErrorFromException` entry for the expected `NB12345v23 unsupported` case, i.e. logging on the passing path, not a failure. The test passes on base. The correction strengthens rather than weakens the finding. The rest of the comment is unchanged.*

src/test/java/com/researchspace/service/impl/UserManagerImplTest.java
```diff
@@ -50,6 +50,11 @@ public class UserManagerImplTest extends BaseManagerMockTestCase {
   public void setUp() throws Exception {
     ReflectionTestUtils.setField(
         userManager, "messages", new MessageSourceUtils(new JsonMessageSource()));
+    // UserManagerImpl declares a UserDao constructor, so @InjectMocks satisfies it by
+    // constructor injection and does not also field-inject. Wire the @Autowired fields the
+    // tests below exercise by hand.
+    ReflectionTestUtils.setField(userManager, "communityDao", communityDao);
+    ReflectionTestUtils.setField(userManager, "analyticsManager", analyticsManager);
```
nhanlon2, line 57: From Claude: **High.** This wires 2 of the 6 `@Autowired` fields, leaving the other 4 null.

`UserManagerImpl` field-injects `communityDao`, `roleDao`, `accountEventDao`, `permissnUtils`, `analyticsManager`, `properties`, `verificationPasswordValidator` and `messages` (`UserManagerImpl.java:74-82`). Since `@InjectMocks` now takes the `UserManagerImpl(UserDao)` constructor path and skips field injection, everything not explicitly set here is null on the instance under test. That includes `roleDao`, even though `@Mock RoleDao roleDao` is declared in this class (it currently only reaches `roleManager`).

The suite passes today because no current test touches those fields. The failure mode is deferred: the next test added to this class that exercises `roleDao`, `permissnUtils`, `properties` or `verificationPasswordValidator` gets a bare NPE with nothing pointing back to injection.

Two options that close it rather than defer it:
- set every `@Autowired` field the class declares, so the comment's "the fields the tests below exercise" is not a moving target; or
- move the dependencies onto the `UserManagerImpl` constructor so `@InjectMocks` covers them and the reflection calls can go away entirely.

pom.xml
```diff
@@ -79,9 +79,11 @@
...
         <groupId>net.bytebuddy</groupId>
         <artifactId>byte-buddy</artifactId>
-        <!-- Keep >= the version Hibernate ORM ships (6.4.4 ships 1.14.11); pinned here for
-             dependencyConvergence between Hibernate and Mockito -->
-        <version>1.14.11</version>
+        <!-- Keep >= the version Hibernate ORM ships (6.4.4 ships 1.14.11) and the version
+             Mockito 5 ships (5.23.0 ships 1.17.7); pinned here for dependencyConvergence
+             between Hibernate and Mockito. byte-buddy-agent comes from mockito-core, so this
+             must track it or the two halves of byte-buddy skew apart. -->
+        <version>1.17.7</version>
```
nhanlon2, line 86: From Claude: **High.** This is a production-runtime bump, not a test-only one.

byte-buddy is pinned here in `<dependencyManagement>`, so 1.14.11 to 1.17.7 applies at compile and runtime, not just under test scope. Hibernate ORM 6.4.4 ships 1.14.11 and uses byte-buddy directly for proxy generation and bytecode enhancement, so this changes a library the running application depends on, inside a PR whose title and description read as test-infrastructure work.

The evidence is reassuring: both Spring Tests jobs pass, several hundred ITs execute, and there is no byte-buddy, `BytecodeProvider` or proxy-related error anywhere in the CI logs. So this is a request for explicit sign-off rather than a defect report.

Worth calling out in the PR description so it is not merged as a test-only change. If you would rather keep the blast radius at test scope, the alternative is letting Mockito bring its own byte-buddy and adding a `dependencyConvergence` exclusion, though that trades one kind of risk for another and the current approach may well be the better one.

nhanlon2 (commented):
From Claude: posting the critical and high-severity findings from a review of this branch. Lower-severity notes were kept out of the PR.

pom.xml
tchoi-rs: Not an issue

nhanlon2 (approved):

PR #1083, opened 2026-09-02
Tighten assertThrows execution to just the line expected to throw

tchoi-rs:
## Description ##
Tighten assertThrows execution to just the line expected to throw, in order to not accidentally hide code lines that are supposed to fail the test.


src/test/java/com/researchspace/api/v1/controller/ShareApiControllerMVCIT.java
```diff
@@ -541,10 +541,13 @@ public void getSharesForSelectionOfRecordsAndFolders() throws Exception {
     assertEquals(4, apiShareSearchResultWithFolders.getTotalHits().intValue());
 
     // attempt to list shares for pi's folder
+    Long piWorkspaceFolderId = piWorkspaceFolder.getId();
```
tchoi-rs, line 544: I don't feel strongly about the increased verbosity, but I'm happy to revert this if needed

src/test/java/com/researchspace/api/v1/controller/ShareApiControllerMVCIT.java
```diff
@@ -541,10 +541,13 @@ public void getSharesForSelectionOfRecordsAndFolders() throws Exception {
...
+    List<Long> sharedItemIds = List.of(piWorkspaceFolderId);
+    MockHttpServletRequestBuilder shareRequest =
+        createShareGetBuilder(sharer, apiKey)
+            .param("sharedItemIds", StringUtils.join(sharedItemIds, ","));
     AuthorizationException authEx =
-        assertThrows(
-            AuthorizationException.class,
-            () -> listSharesWithSharedItemIds(sharer, apiKey, List.of(piWorkspaceFolder.getId())));
+        assertThrows(AuthorizationException.class, () -> mockMvc.perform(shareRequest));
```
rs-nicof: This one picked the wrong line to keep, and CI is already failing on it: `getSharesForSelectionOfRecordsAndFolders:550 Expected org.apache.shiro.authz.AuthorizationException to be thrown, but nothing was thrown.`

The error is handled server-side, so it never comes back out of `perform()`. The old helper only threw because it went on to read the response body, and that is the line that got dropped. Reading the resolved exception instead is the pattern you documented in AGENTS.md, and it keeps the message check below working:

```suggestion
    MvcResult listResult = mockMvc.perform(shareRequest).andReturn();
    AuthorizationException authEx =
        assertInstanceOf(AuthorizationException.class, listResult.getResolvedException());
```

Needs a static import for `assertInstanceOf`.

rs-nicof (commented):
Reviewed the whole sweep (179 files) across six parallel passes: general quality, security and persistence. One issue worth fixing; everything else looked like a faithful conversion, and several hunks recover assertions that were previously dead inside the old lambdas.

src/test/java/com/researchspace/api/v1/controller/ShareApiControllerMVCIT.java
rs-nicof: Worth reverting, though not for the verbosity. The narrower version stops catching the failure it is meant to catch, and it is the one test failing here that is not already failing on the base branch. Detail in [this comment](https://github.com/rspace-os/rspace-web/pull/1083#discussion_r4047469029).

rs-nicof (changes requested):

rs-nicof (approved):

PR #1085, opened 2026-09-02
RSDEV-1354-export-import-inventory-links

nhanlon2:
## Summary

Inventory link fields (RSDEV-1131) now round-trip through CSV export and import.

### Export

Every link field, whether a sample or instrument template field or an extra field on any record type, now exports to its existing column as a single cell:

```
<RelationType> <serverUrl>/globalId/<GID>[vN]
```

for example `IsDerivedFrom https://rspace.example.com/globalId/SA123v2`. The version pin travels as the `vN` suffix. Unset links export as an empty cell. Previously the cell was always empty because link fields hold no data-column value.

### Import (samples and instruments)

- The server-side type suggester proposes **Link** for a column whose non-blank values all parse, so the mapping UI defaults that column to "Custom Field (Link)" and offers Link in the type dropdown. No frontend code change was needed: the Link type, label, icon and help text already existed.
- A Link-typed column becomes the field's link (relation type, base Global ID, version pin) instead of its content. A malformed cell fails that row with a new externalised message and leaves other rows unaffected.
- Only URLs on this server are accepted. A Global ID from another RSpace instance is rejected rather than silently attached to a different local record.
- Targets that are missing or unreadable on this server are still stored. The import sets a JSON-ignored `skipTargetCheck` flag on the link DTO, and the link manager skips only the exists-and-readable check for it. The UI then shows the existing missing or no-access state for that link, matching what happens when a target is deleted or unshared after linking.

Container and subsample import has no custom-field path today, so their link fields export but have nowhere to land on import. That is unchanged by this PR.

### Tests

Fast unit tests (all green locally): `CsvSampleLinkExportTest`, `CsvLinkValueParserTest`, `InventoryImportInventoryEntityFieldCreatorTest`, `InventoryLinkManagerImplSkipTargetCheckTest`, `CsvSampleImporterLinkFieldTest`, `CsvInstrumentImporterLinkFieldTest`, plus a link-column case in `ImportModel/instruments.test.tsx`.

`InventoryImportApiControllerMVCIT.parseAndImportSampleCsvWithLinkColumn` covers parse and import of a dangling pinned link end to end. It compiles but was not run locally.

### DIFF
--- a/src/main/java/com/researchspace/api/v1/model/ApiInventoryLink.java
+++ b/src/main/java/com/researchspace/api/v1/model/ApiInventoryLink.java
@@ -1,12 +1,15 @@
 package com.researchspace.api.v1.model;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import com.researchspace.model.core.GlobalIdentifier;
 import com.researchspace.model.inventory.field.InventoryLink;
 import lombok.Data;
+import lombok.EqualsAndHashCode;
 import lombok.NoArgsConstructor;
+import lombok.ToString;
 
 /** API payload representing a Link extra-field's target/relation metadata. */
 @Data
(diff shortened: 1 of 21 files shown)


nhanlon2:
Closed: head was on the ResearchSpace-ELN fork by mistake. Recreated from the rspace-os branch as https://github.com/rspace-os/rspace-web/pull/1086.

PR #1086, opened 2026-09-03
RSDEV-1354-export-import-inventory-links

nhanlon2:
## Summary

Inventory link fields (RSDEV-1131) now round-trip through CSV export and import. Link columns used to export empty.

### Test ticket
https://researchspace.atlassian.net/browse/RSTEST-461

### Export

Every link field, whether a sample or instrument template field or an extra field on any record type, exports one cell as `<RelationType> <serverUrl>/globalId/<GID>[vN]`, for example `IsDerivedFrom https://rspace.example.com/globalId/SA123v2`. The version pin travels as the `vN` suffix. With no server URL configured the cell holds the relation type alone.

The exporter builds the URL with `InventoryUrls.globalIdPageUrl`. The importer recognises it with the new `InventoryUrls.globalIdPagePrefix`, which normalises the server URL the same way.

### Import (samples and instruments)

- **Link detection.** The type suggester types a column as Link only when at least one value is present and every present value parses (`CsvLinkValueParser`). Otherwise the column falls through to the other types as before, usually text, and the cells import verbatim.
- **Manual override.** Link is now offered in the import column-type menu for every column except reference and attachment columns. Choosing Link for values that do not parse fails the row with `errors.inventory.import.linkValueInvalid` while the CSV is read, and the import stops before anything is saved.
- **Other instances.** Only URLs on this server parse, so a Global ID exported from another RSpace is not attached to an unrelated local record.
- **Multi-template exports.** A row without a given column carries `#N/A`. The suggester and both importers treat that as absent (`InventoryItemCsvExporter.isAbsentCsvValue`), so a mixed-template export re-imports as links.
- **Long values.** A value containing an `http(s)://` URL may be up to 500 characters (was 100) before the suggester forces text, since a long hostname can push a link cell past 100.
- **Messages.** The sample importer's column-count and duplicate-import-id errors are now externalised. The column-count key is shared with instruments and renamed from `instrumentCsvLineUnexpectedColumnCount` to `csvLineUnexpectedColumnCount`.

### Unresolvable link targets

CSV import stores a link whatever state its target is in (`ApiInventoryLink.skipTargetCheck`, set only by the parser). The card decides what to show at read time. A target the viewer cannot resolve, whether it never existed, was hard-deleted, or is not shared with them, reports `readable: false`. The card shows a **No access** pill and disables Open and Info. The tooltip now reads "This item is not available to you: it may not be shared with you, or it may no longer exist".

This does not distinguish "gone" from "not yours", the rule ADR-0002 already applied to ELN targets. Rejecting unresolvable targets at import would tell the importer whether a record they cannot see exists. Because a failed row stops the import, it would also let one bad target discard every other row.

**Consequences for review:**

1. **Trashed targets the viewer can read show "Target deleted".** An inventory target keeps a working Open, since trashed items have a viewer. An ELN target loses Open. An unresolvable target of either kind shows "No access" instead.
2. **Import can store a link to a record the importer cannot read.** The target's owner sees it only if they can also read the importing record, because `findReferencingItems` filters every source by the caller's read permission. The UI picker cannot select an unreadable target, so this applies to import only.
3. **No access now applies to inventory targets too.** Previously the card ignored `readable` for inventory, on the grounds that every user keeps a limited-read view. That is not so: `canUserLimitedReadInventoryRecord` grants it only through container containment, a list of materials, or a template, and an owner can restrict an item further. ADR-0002 is updated.

**Holes the leniency opened, and how each is closed:**

4. **Purged audit history.** Envers rows can be purged while the record lives on. When there is no snapshot, the summary resolver now asks `LinkTargetResolver.readableInventoryTarget` for the live record, so a readable inventory target stays openable and a trashed one reports `deleted: true`. This fallback is inventory only. An ELN target with no snapshot is redacted.
5. **Type-exact resolution.** Samples and sample templates share one table and one numeric id space, and `getInvRecordByGlobalId` resolves `SA` and `IT` through the same lookup. Inventory resolution now requires the resolved record's own prefix to match the requested one, in `targetExistsAndIsReadable` (which gates link create/update and `findReferencingItems`), `targetIsLiveAndReadable`, and the summary fallback. Before this, a readable `SA90` let an API client create a link to an `IT90` that does not exist. ADR-0002 records the rule and says it must be re-checked when a record kind is added.
6. **ELN summaries no longer 500.** For a non-owner, the summary used to check ELN readability through `BaseRecordManager`, whose lookups throw inside transactional `*Manager` proxies. Catching the exception still left the transaction rollback-only, so `getTargetSummary` failed at commit and the card, receiving nothing, showed no pill and kept Open. The summary now decides ELN read permission from the audit snapshot with `IPermissionUtils.isPermitted`, after `refreshCacheIfNotified()` so an unshare applies on that request. Link create/update and `findReferencingItems` still use the `BaseRecordManager` lookup for ELN targets, unchanged by this PR.
7. **Self-links through import.** Link fields are applied before a new record is persisted, so `rejectSelfLink` cannot see the record's id, and the target-exists check that used to reject a future self Global ID is skipped on import. `assertNoSelfLinkAfterSave` now runs right after a sample or instrument is persisted.

`skipTargetCheck` is `@JsonIgnore`, pinned by `ApiInventoryLinkJsonTest`, so API clients cannot set it. All other callers still require a target they can read.

Container and subsample import has no custom-field path, so their link extra fields export but are not re-imported. Unchanged by this PR.

### Known issue, tracked separately

Import failures raised as `ApiRuntimeException` show the raw i18n key instead of the message, because the import paths use `getMessage()` rather than the resolving handler. Pre-existing, affects all four importers, tracked as **RSDEV-1521**.

### Tests

Fast unit: `CsvSampleLinkExportTest`, `CsvInstrumentLinkExportTest`, `CsvLinkRoundTripTest`, `CsvLinkValueParserTest`, `InventoryUrlsTest`, `InventoryImportInventoryEntityFieldCreatorTest`, `CsvSampleImporterLinkFieldTest`, `CsvInstrumentImporterLinkFieldTest`, `InventoryLinkManagerImplSkipTargetCheckTest`, `LinkTargetResolverImplTest`, `LinkTargetSnapshotResolverImplTest`, `SampleApiManagerImplLinkFieldTest`, `ApiInventoryLinkJsonTest`. Frontend: `FieldTypes.test.ts`, `LinkField.test.tsx`, and link-column cases in `ImportModel/instruments.test.tsx`. `CsvLinkRoundTripTest` feeds the exporter's output back through the parser.

`InventoryImportApiControllerMVCIT`:

| Case | Asserts |
|---|---|
| `parseAndImportSampleCsvWithLinkColumn` | A dangling pinned link imports and is stored |
| `parseAndImportSampleCsvWithLinkToExistingVersionedTarget` | A live target pinned at a real version resolves |
| `parseAndImportSampleCsvWithLinkToUnreadableTarget` | A link to another user's record is stored and reads back redacted. The stored row is checked at the DAO layer, below the permission filter |
| `sampleCsvLinkingToItsOwnFutureGlobalIdIsRejected` | A CSV naming the id the new sample will receive is rejected and nothing is persisted |
| `parseAndImportInstrumentCsvWithLinkColumn` | The instrument persistence path |
| `exportedMultiRowSampleCsvWithSentinelLinkColumnReimportsAsLinks` | Two samples exported through the export API re-import as links, `#N/A` cells included |

`LinkTargetSnapshotResolverIT`:

| Case | Asserts |
|---|---|
| `reportsLiveReadableTargetWhoseAuditRowsWerePurgedAsReadable` | A live readable target with purged audit rows stays openable |
| `reportsSoftDeletedTargetWhoseAuditRowsWerePurgedAsDeletedButReadable` | A trashed one reports deleted but readable |
| `reportsTemplateRequestCollidingWithReadableSampleIdAsUnreadable` | A readable sample does not vouch for a template with the same id |
| `reportsDanglingNotebookTargetAsRedactedWithoutPoisoningTheTransaction` | A nonexistent notebook is redacted and the transaction commits |
| `reportsAnotherUsersNotebookAsRedactedWithoutPoisoningTheTransaction` | Same for another user's live notebook |
| `reportsAnotherUsersDeletedNotebookAsRedactedWithoutPoisoningTheTransaction` | Same for another user's trashed notebook |

The permission-cache refresh in item 6 is pinned by `LinkTargetSnapshotResolverImplTest`. The existing `InventoryLinkTargetApiControllerMVCIT.unshareIsReflectedInSummaryWithoutServerRestart` cannot catch its absence, because API-key requests clear the authz cache on every request.


nhanlon2:
@pullfrog

src/main/java/com/researchspace/service/inventory/impl/LinkTargetSnapshotResolverImpl.java
```diff
@@ -74,6 +74,19 @@ public ApiInventoryLinkTargetSummary resolveSummary(
...
+      if (isInventoryPrefix(prefix)) {
+        // No inventory record here at all: never existed on this server (a CSV-imported dangling
+        // link, RSDEV-1354) or its audit rows were purged. Report it as deleted so the card shows
+        // "Target deleted" instead of a working-looking Open.
+        // This is a deliberate, narrow carve-out from ADR-0002, not an application of it: an
+        // existing-but-unreadable inventory record still answers readable=false, so the two
+        // payloads differ and a caller walking ids learns which inventory ids are occupied. Name,
+        // type and owner stay unset either way, so what leaks is the id high-water mark rather
+        // than any content, and that was judged a fair price for not offering a dead Open link.
+        summary.setReadable(true);
```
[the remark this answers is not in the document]
nhanlon2: This is from Claude: fixed in: be59d6ccf

src/main/java/com/researchspace/service/inventory/impl/LinkTargetResolverImpl.java
```diff
@@ -92,6 +92,24 @@ public boolean targetIsLiveAndReadable(GlobalIdentifier target, User user) {
...
+    if (target == null) {
+      return false;
+    }
+    GlobalIdentifier base =
+        target.hasVersionId() ? new GlobalIdentifier(target.getPrefix(), target.getDbId()) : target;
+    if (!INVENTORY_PREFIXES.contains(base.getPrefix())) {
+      return false;
+    }
+    try {
+      inventoryPermissionUtils.getInvRecByGlobalIdOrThrowNotFoundException(base);
```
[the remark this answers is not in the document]
nhanlon2: This is from Claude: fixed in: be59d6ccf

src/main/java/com/researchspace/service/inventory/impl/LinkTargetResolverImpl.java
```diff
@@ -92,6 +92,24 @@ public boolean targetIsLiveAndReadable(GlobalIdentifier target, User user) {
...
   }
 
+  @Override
+  public boolean targetIsKnownMissing(GlobalIdentifier target) {
+    if (target == null) {
+      return false;
+    }
+    GlobalIdentifier base =
+        target.hasVersionId() ? new GlobalIdentifier(target.getPrefix(), target.getDbId()) : target;
+    if (!INVENTORY_PREFIXES.contains(base.getPrefix())) {
```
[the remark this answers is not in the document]
nhanlon2: This is from Claude: fixed in: be59d6ccf

nhanlon2:
@pullfrog

nhanlon2:
@pullfrog

src/main/java/com/researchspace/service/inventory/csvimport/InventoryImportSampleFieldCreator.java
```diff
@@ -68,6 +82,11 @@ public InventoryEntityField getSuggestedSampleFieldForNameAndValues(
     if (isSuggestedFieldForValues(valueSet, new InventoryUriField())) {
       return new InventoryUriField(name);
     }
+    // link: every value has to parse. allMatch on an empty set would be vacuously true, but the
+    // caller returns before here when there are no values
+    if (valueSet.stream().allMatch(linkParser::isParseable)) {
```
nhanlon2: The exporter fills union columns that are absent on a row with `#N/A`, but this new all-values check treats that sentinel as real data. This breaks the advertised round trip for ordinary multi-record exports: an extra-link column is scoped to its connected record, so every other row gets `#N/A`; structured link columns get the same sentinel for records from other templates. The column is then inferred as STRING, and the import UI does not offer LINK as a compatible type for a string column. If an existing LINK template is selected instead, the new sample/instrument import branches call `linkParser.parse("#N/A")` and reject those rows. Please treat the export sentinel as an absent value both during inference and when populating LINK fields, and cover a multi-row export/import where at least one row lacks the link column.

nhanlon2:
@pullfrog

src/main/java/com/researchspace/service/inventory/csvimport/InventoryImportSampleFieldCreator.java
nhanlon2: This is from Claude: fixed in: 8cec63cdc

nhanlon2:
@pullfrog

nhanlon2:
@pullfrog

src/main/java/com/researchspace/service/inventory/impl/InventoryLinkManagerImpl.java
```diff
@@ -152,7 +147,17 @@ private GlobalIdentifier parseAllowedTargetOrThrow(String targetGlobalId) {
...
-  private void assertTargetExistsAndReadable(ApiInventoryLink apiLink, User actor) {
+  /**
+   * Rejects a link whose target does not resolve to a record the actor can READ, on both write
+   * paths. CSV import (RSDEV-1354) is exempt: it stores the link whatever state the target is in,
+   * because distinguishing "gone" from "not yours" is exactly the disclosure ADR-0002 prevents. The
+   * card reports either as "No access" at read time, so nothing is claimed that the viewer cannot
+   * verify, and one unresolvable target does not discard the rest of the import.
+   */
+  private void assertTargetAcceptable(ApiInventoryLink apiLink, User actor) {
+    if (apiLink.isSkipTargetCheck()) {
```
nhanlon2, line 158: > Written by codex, an AI coding agent.

**[P1] Preserve self-link validation on imported creates**

Returning here skips the only check that used to reject a target which does not yet exist. That matters when creating samples and instruments: their link fields are applied before the new record is persisted and receives its ID, so the normal create path relies on target existence to make a future self Global ID impossible. A crafted CSV can target the predictable ID that the imported row will receive; skipTargetCheck accepts it, and rejectSelfLink cannot help because the source record has no ID yet. The saved record then violates the explicit no-self-link invariant. Please retain a post-allocation self-link check for imported records, or otherwise prevent the lenient path from accepting the source records eventual Global ID, and add import coverage for this case.

src/main/java/com/researchspace/service/inventory/impl/LinkTargetSnapshotResolverImpl.java
```diff
@@ -74,6 +74,9 @@ public ApiInventoryLinkTargetSummary resolveSummary(
             ? auditManager.getObjectForRevision(cls, dbId, targetRevisionId)
             : auditManager.getNewestRevisionForEntity(cls, dbId);
     if (snapshot == null || snapshot.getEntity() == null) {
+      // Nothing to resolve: redacted, for every prefix. Nonexistent must look exactly like
```
nhanlon2: > Written by codex, an AI coding agent.

**[P2] Keep Open available for a readable live target with no audit row**

A missing Envers snapshot does not imply that the live target is unresolvable. Audit rows can be absent or purged while the record still exists and is readable; this branch returns readable=false, and the new LinkField logic consequently shows No access and removes Open from a working target. The non-disclosure rule does not require that false negative: checking targetExistsAndIsReadable for the current actor is permission-gated and reveals nothing they cannot already read. Please distinguish a readable live record from the missing/unreadable redacted case here, and cover a live readable target with no audit snapshot.

nhanlon2:
> Written by codex, an AI coding agent.

The PR description is stale in a security-sensitive section. Under Target check on import it says existing unreadable targets are rejected and that the bypass is limited to provably missing targets. The current head does the opposite: assertTargetAcceptable returns immediately for every CSV-imported link, and the updated ADR/tests explicitly require unreadable targets to be stored. Please update the description so reviewers and release notes reflect the actual authorization behavior.

src/main/java/com/researchspace/service/inventory/impl/InventoryLinkManagerImpl.java
nhanlon2: > Written by Claude, an AI coding agent.

Agreed, and the comment on rejectSelfLink said as much: it claimed creation was safe because the target-exists check rejected a future self Global ID, and import no longer runs that check. Fixed in 8855f6079 with assertNoSelfLinkAfterSave, called once the record has an id, on both the sample and instrument create paths. Covered end to end by InventoryImportApiControllerMVCIT.sampleCsvLinkingToItsOwnFutureGlobalIdIsRejected, which imports a CSV naming the id the sample is about to receive and asserts nothing is persisted. The stale comment is corrected too.

src/main/java/com/researchspace/service/inventory/impl/LinkTargetSnapshotResolverImpl.java
nhanlon2: > Written by Claude, an AI coding agent.

Agreed, and this was a regression I introduced: the redacting branch was fine while the No access pill was ELN-only, and became visible when I widened it. Fixed in 8855f6079 - the branch now asks targetExistsAndIsReadable for the current actor before redacting. Missing and unreadable still both redact, so they stay indistinguishable to anyone who cannot read the target. Covered by LinkTargetSnapshotResolverIT.reportsLiveReadableTargetWhoseAuditRowsWerePurgedAsReadable, which deletes the Sample_AUD rows of a live sample and asserts the summary comes back readable.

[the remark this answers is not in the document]
nhanlon2: > Written by Claude, an AI coding agent.

Fixed in 8855f6079. See the reply on the sibling thread for detail.

src/main/java/com/researchspace/service/inventory/csvexport/InventoryItemCsvExporter.java
```diff
@@ -100,6 +116,32 @@ public enum ExportedContentType {
...
   protected @Autowired CsvExportCommentGenerator exportCommentGenerator;
   protected @Autowired MessageSourceUtils messages;
+  protected @Autowired IPropertyHolder properties;
+
+  /**
+   * CSV cell for a link field: {@code "<RelationType> <serverUrl>/globalId/<GID>[vN]"}, or empty
+   * when the field holds no link. The version pin travels as the {@code vN} suffix so a single
+   * Global ID URL carries the whole link.
+   */
+  /** CSV cell for a template-defined field: its link if it holds one, its plain data otherwise. */
```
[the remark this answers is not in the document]
nhanlon2: > Written by Claude, an AI coding agent.

Correct - the csvValueForField helper landed between the link-format javadoc and the method it documented. Reattached in 8855f6079, with the field-specific block staying on csvValueForField.

src/main/java/com/researchspace/service/inventory/impl/InventoryLinkManagerImpl.java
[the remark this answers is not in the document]
nhanlon2: > Written by Claude, an AI coding agent.

Correct reading, and to answer the open question directly: yes, this was a deliberate decision, taken with the PR author, not an oversight.

Distinguishing "gone" from "not yours" at import time is itself the disclosure ADR-0002 exists to prevent - the row succeeding or failing tells the importer whether a record they cannot see exists. Reporting a missing target as deleted also asserts something the server has not proven. And because import is all-or-nothing, rejecting the unreadable case let one bad target discard every other row in the file.

The accepted cost is the one you name: a link to a record the importer cannot read is stored, and appears in that owner referencing-items view. Nothing is disclosed to the importer. The PR description states this explicitly as a change to the previous READ-target requirement, scoped to the import path; ordinary and API callers still require a readable target, and skipTargetCheck remains @JsonIgnore so no client can reach the leniency.

Two consequences of the leniency that were NOT accepted are fixed in 8855f6079: self-link minting via a predictable future id, and a live target with purged audit rows being shown as No access.

Resolving as intended behaviour.

src/main/java/com/researchspace/service/inventory/impl/LinkTargetSnapshotResolverImpl.java
```diff
@@ -74,6 +74,16 @@ public ApiInventoryLinkTargetSummary resolveSummary(
             ? auditManager.getObjectForRevision(cls, dbId, targetRevisionId)
             : auditManager.getNewestRevisionForEntity(cls, dbId);
     if (snapshot == null || snapshot.getEntity() == null) {
+      // No audit history does not mean no record: rows can be purged while the record lives on.
+      // Asking whether this actor can read it is permission-gated, so it reveals nothing they
+      // could not already see, and it keeps Open on a target whose page works.
+      if (linkTargetResolver.targetExistsAndIsReadable(new GlobalIdentifier(prefix, dbId), user)) {
```
nhanlon2: > Written by codex, an AI coding agent.

**[P1] Avoid a rollback-only live lookup when the snapshot is absent**

This fallback reintroduces the transactional failure described by this class’s own isReadable Javadoc. For an ELN target with no audit row—e.g. a dangling NB imported by this PR—targetExistsAndIsReadable calls transactional BaseRecordManager.getByGlobalIdsAndReadPermission; its get(..., includeDeleted=false) throws for a missing/deleted folder, so Spring marks the participating InventoryLinkManager.getTargetSummary transaction rollback-only before LinkTargetResolverImpl catches the exception. The outer call then fails at commit with UnexpectedRollbackException instead of returning the intended redacted “No access” summary. Please make the absent-snapshot fallback use a genuinely non-throwing lookup (or isolate it in a transaction whose rollback cannot poison the caller) and cover a nonexistent/deleted notebook with no audit snapshot through getTargetSummary.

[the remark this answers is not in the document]
nhanlon2: > Written by Claude, an AI coding agent.

This is from Claude: fixed in: b2cad33a55977a2a58630131d80bd1fdfdf165fc

nhanlon2:

@pullfrog

src/main/java/com/researchspace/service/inventory/impl/LinkTargetSnapshotResolverImpl.java
nhanlon2: > Written by Claude, an AI coding agent.

This is from Claude: fixed in: baa6250ca9c418f4538dfcc6b41367842cc37008

nhanlon2:

@pullfrog

src/main/java/com/researchspace/service/inventory/LinkTargetResolver.java
```diff
@@ -23,13 +25,35 @@ public interface LinkTargetResolver {
...
+   * Global ID prefix differs from the requested one is not the target. Empty for a non-Inventory
+   * prefix, since resolving ELN records runs through transactional {@code *Manager} proxies that
+   * throw for missing or deleted records and would mark the caller's transaction rollback-only.
+   *
+   * @param target the parsed link target GlobalID (any version suffix is ignored)
+   * @param user the acting user, whose READ permission decides
+   * @return the record, or empty if it does not exist, is not readable, is of another kind, or the
+   *     prefix is not an Inventory one
+   */
+  Optional<InventoryRecord> readableInventoryTarget(GlobalIdentifier target, User user);
```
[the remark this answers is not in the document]
nhanlon2: > Written by Claude, an AI coding agent.

This is from Claude: fixed in: 0779bc72e219289df289034dc5a551de4d94a543

src/main/java/com/researchspace/service/inventory/impl/LinkTargetResolverImpl.java
```diff
@@ -77,21 +77,47 @@ public boolean targetIsLiveAndReadable(GlobalIdentifier target, User user) {
...
+    if (target == null) {
+      return Optional.empty();
+    }
+    permissionUtils.refreshCacheIfNotified();
+    GlobalIdentifier base =
+        target.hasVersionId() ? new GlobalIdentifier(target.getPrefix(), target.getDbId()) : target;
+    if (!INVENTORY_PREFIXES.contains(base.getPrefix())) {
+      return Optional.empty();
+    }
+    return readableInventoryRecord(base, user);
```
nhanlon2, line 103: > Written by codex, an AI coding agent.

**[P1] Apply the new type-exact check to the authorization path**

This helper correctly rejects a record whose prefix differs from the requested target, but `targetExistsAndIsReadable` still uses `isInventoryReadable`, which resolves by numeric ID without this check. Consequently, if readable sample `SA90` exists, a request for `IT90` can be authorized against that sample because `InventoryRecordRetriever` maps both prefixes through `getSampleIfExists`. A normal API client can then create a link to a missing or unreadable `IT90`, despite the PR contract limiting that bypass to CSV import, and the same mistake gates `findReferencingItems`. Please route the Inventory branch of `targetExistsAndIsReadable` through `readableInventoryRecord` as well (preserving the intended deleted-record semantics), and cover the `IT`/`SA` collision on that method rather than only on the new summary helper.

src/main/java/com/researchspace/service/inventory/impl/LinkTargetSnapshotResolverImpl.java
```diff
@@ -74,6 +75,35 @@ public ApiInventoryLinkTargetSummary resolveSummary(
...
+      // with a snapshot - deleted, but readable and openable in the trash - rather than being
+      // hidden behind "No access". The lookup is permission-gated and type-exact, so an
+      // unreadable record and a sibling sharing the db id (SA/IT) both fall through to redaction.
+      //
+      // Inventory only. The ELN lookup resolves through transactional *Manager proxies
+      // (BaseRecordManager -> FolderManager) whose folderDao.get and read/not-deleted assertions
+      // throw for a missing, unreadable or deleted record. Catching the throw is not enough: it
+      // crosses a transactional proxy and marks the caller's transaction rollback-only, so
+      // getTargetSummary would fail at commit with UnexpectedRollbackException instead of
+      // returning the redacted summary below - the same trap isReadable's javadoc describes, and
```
nhanlon2, line 92: > Written by codex, an AI coding agent.

**[P1] Keep the rollback-only ELN lookup out of the snapshot-present path too**

This guard avoids the transactional ELN lookup only when no audit snapshot exists. A soft-deleted notebook normally still has a `Folder_AUD` snapshot; for a viewer other than the snapshot owner, execution reaches `isReadable` below and calls `targetExistsAndIsReadable`. `BaseRecordManager` then loads the folder with `includeDeleted=false`, throws across its transactional proxy, and marks `getTargetSummary` rollback-only even though `LinkTargetResolverImpl` catches the exception. The endpoint therefore still finishes with `UnexpectedRollbackException` instead of the redacted No-access payload, and the card falls back to leaving Open enabled. This PR makes that readily reachable by importing a link to another user’s deleted notebook. Please use a non-throwing permission path for snapshot-backed ELN targets as well, and add an integration case for a deleted notebook viewed by a non-owner; the current tests cover only a no-snapshot dangling notebook and a deleted notebook viewed by its owner.

nhanlon2:

@pullfrog

src/main/java/com/researchspace/service/inventory/impl/LinkTargetResolverImpl.java
nhanlon2: > Written by Claude, an AI coding agent.

This is from Claude: fixed in: 733fc2a961e56252ac3372c014e0e803cac57d8d

src/main/java/com/researchspace/service/inventory/impl/LinkTargetSnapshotResolverImpl.java
nhanlon2: > Written by Claude, an AI coding agent.

This is from Claude: fixed in: 733fc2a961e56252ac3372c014e0e803cac57d8d

nhanlon2:

@pullfrog please review this pull request.

src/main/java/com/researchspace/service/inventory/impl/LinkTargetSnapshotResolverImpl.java
```diff
@@ -108,6 +148,10 @@ private boolean isReadable(GlobalIdPrefix prefix, Long dbId, Object entity, User
         && owner.getUsername().equals(user.getUsername())) {
       return true;
     }
+    if (!isInventoryPrefix(prefix)) {
+      return entity instanceof BaseRecord baseRecord
+          && permissionUtils.isPermitted(baseRecord, PermissionType.READ, user);
```
nhanlon2, line 159: > Written by codex, an AI coding agent.

**[P1] Refresh revoked permissions before checking the audit snapshot**

The old ELN path went through `targetExistsAndIsReadable`, which deliberately calls `refreshCacheIfNotified()` before checking READ because an unshare otherwise leaves the viewer’s revoked Shiro grant cached. This replacement calls `isPermitted` directly without that refresh. `PermissionUtils.checkPermissions` tests `SecurityUtils.getSubject().isPermitted(adapter)` first, so immediately after an unshare the stale `RECORD:READ` grant can return true even though the newest snapshot ACL no longer contains the viewer; the summary then discloses the target name/type and keeps Open enabled until cache expiry. Please apply the pending permission-cache refresh before this snapshot check and cover an actual share → summary → unshare → summary transition, rather than only mocking `isPermitted`.

nhanlon2: > Written by Claude, an AI coding agent.

This is from Claude: fixed in: a93a6e551dd547cf71dc83c8fc44dd5559d41888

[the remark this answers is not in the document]
nhanlon2: > Written by Claude, an AI coding agent.

This is from Claude: fixed in: a93a6e551dd547cf71dc83c8fc44dd5559d41888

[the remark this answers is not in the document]
nhanlon2: > Written by Claude, an AI coding agent.

This is from Claude: fixed in: a93a6e551dd547cf71dc83c8fc44dd5559d41888

nhanlon2:

@pullfrog please review this pull request.

src/main/java/com/researchspace/service/inventory/impl/LinkTargetSnapshotResolverImpl.java
```diff
@@ -108,6 +150,14 @@ private boolean isReadable(GlobalIdPrefix prefix, Long dbId, Object entity, User
         && owner.getUsername().equals(user.getUsername())) {
       return true;
     }
+    if (!isInventoryPrefix(prefix)) {
+      // an unshare only queues the revocation; Shiro keeps the viewer's RECORD:READ grant until a
+      // permission-checked request applies it. The live lookup this replaced refreshed first, so
+      // skipping it would let a former sharee keep the target's name and type until cache expiry.
+      permissionUtils.refreshCacheIfNotified();
+      return entity instanceof BaseRecord baseRecord
+          && permissionUtils.isPermitted(baseRecord, PermissionType.READ, user);
```
rs-nicof: > Written by Claude, an AI coding agent.

Deciding ELN readability from the audit snapshot can go stale in the permissive direction. Unsharing or unpublishing a document saves it without an auditable delta, so no new revision is written and the newest snapshot still carries the old ACL. For example: publish a document, edit it, then unpublish it, and any user can still fetch its name from the summary endpoint. Could the ACL/public-link half come from the live row (a plain DAO read, no manager proxy), with only the name and type taken from the snapshot?

src/main/java/com/researchspace/api/v1/model/ApiInventoryLink.java
```diff
@@ -24,6 +27,17 @@ public class ApiInventoryLink {
   @JsonProperty("versionPin")
   private Long versionPin;
 
+  /**
+   * Set only by the CSV importer: store the link even when its target no longer exists on this
+   * server, so an exported dangling link re-imports as a dangling link (RSDEV-1354). A target that
+   * does exist is still checked for readability, so the flag cannot forge a link to a record the
```
rs-nicof: > Written by Claude, an AI coding agent.

This comment says an existing target is still checked for readability, but `assertTargetAcceptable` returns before any check when the flag is set, and the PR description (item 2) says that is intended. It also says API clients cannot reach the leniency, but the import endpoint is itself an API endpoint. Please reword it to match the real rule, so nobody later relies on a guard that isn't there.

src/main/webapp/ui/src/Inventory/components/Fields/Link/LinkField.tsx
```diff
@@ -77,15 +77,13 @@ export default function LinkField(props: LinkFieldProps): React.ReactElement {
...
-  // two pills can never co-occur.
-  const noAccess = targetSummary?.readable === false && !targetIsInventory;
-  // deleted inventory items live on in the trash and their viewer works, so
-  // only deleted or unreadable ELN targets lose Open (their routes are just
-  // error pages)
-  const openBlocked = !targetIsInventory && (targetDeleted || noAccess);
+  // a target the viewer cannot resolve gets "No access", whatever its kind. The server
+  // redacts unreadable and nonexistent targets identically (ADR-0002), so this covers both
+  // and the card never claims a record is deleted when it cannot see whether it is.
+  const noAccess = targetSummary?.readable === false;
```
rs-nicof: > Written by Claude, an AI coding agent.

For inventory targets `readable` reflects full READ only (`canUserReadInventoryRecord`), but the item pages also open for limited-read users: items reached through a container, a List of Materials, or a template. Those users now lose Open and get a 'may no longer exist' tooltip for an item they can actually view. Either report limited read as readable in the summary, or keep Open for inventory targets. The template-field card (`LinkFieldValue.tsx`) still exempts inventory, so the two cards now disagree.

rs-nicof (commented):

Parallel review of the CSV link export/import (six reviewers: layering, i18n, security, persistence, frontend, general quality). Three High findings below; transactions, rollback on a rejected self-link, and the SA/IT type check all looked sound.

rs-nicof (changes requested):

rs-nicof:
AWS instance: https://rsdev-1354-export-import-inventory-links-bb57ac5e-28.researchspace.com

src/main/java/com/researchspace/service/inventory/impl/LinkTargetSnapshotResolverImpl.java
nhanlon2: > Written by Claude, an AI coding agent.

This is from Claude: fixed in: 6e27333a0. ELN read permission in the summary now comes from the live row via `RecordDao`/`FolderDao` `getSafeNull` (no manager proxy); name and type still come from the snapshot. New `InventoryLinkTargetApiControllerMVCIT.unshareAfterAnEditIsReflectedInSummary` pins it: share, rename (writes a revision capturing the ACL), unshare, then the former sharee gets `readable: false`.

src/main/java/com/researchspace/api/v1/model/ApiInventoryLink.java
nhanlon2: > Written by Claude, an AI coding agent.

This is from Claude: fixed in: 19467ac4f. The Javadoc now says the flag skips the target check entirely, and that the CSV import endpoint (UI or API) reaches it; only JSON link payloads cannot set it.

src/main/webapp/ui/src/Inventory/components/Fields/Link/LinkField.tsx
nhanlon2: > Written by Claude, an AI coding agent.

This is from Claude: fixed in: c6e388d31. The summary now counts limited read as readable for inventory targets (`viewableInventoryTarget`, used by the no-snapshot fallback and the snapshot path). Link creation and `findReferencingItems` still require full READ. `LinkFieldValue.tsx` no longer exempts inventory targets, so both cards follow one rule. ADR-0002 is updated.

rs-nicof (approved):

tchoi-rs (approved):

PR #1089, opened 2026-09-03
Simplify AGENTS.md

tchoi-rs:
## Description ##
Simplify AGENTS.md, and also add instructions to use Browser Automation on the live Docker Dev Stack for feature verification.


.agents/skills/ponytail/SKILL.md
```diff
@@ -0,0 +1,120 @@
+---
+name: ponytail
+description: >
+  Forces the laziest solution that actually works, simplest, shortest, most
+  minimal. Channels a senior dev who has seen everything: question whether the
+  task needs to exist at all (YAGNI), reach for the standard library before
+  custom code, native platform features before dependencies, one line before
+  fifty. Supports intensity levels: lite, full (default), ultra. Use on ANY
```
rs-fraser: This reads as though ponytail will be active all the time now. Is that correct? That seems a bit overbearing, and could conflict with skills/instructions a dev has locally. I agree we need to something to reduce the ai slop, but not sure this is the best way. Also, I've found ponytail often oversimplifies things, which could be a symptom of our overly-complex codebase. 

At least, before we add instructions that affect every devs workflow, we should come to a consensus as a team.

AGENTS.md
```diff
@@ -247,3 +158,9 @@ Useful references:
 - Security: `DevDocs/DeveloperNotes/SecurityAndPermissions.md`
 - Logging: `DevDocs/DeveloperNotes/Logging.md`
 - CI: `.github/workflows/lint-and-test.yml` and `Jenkinsfile`
+</important>
+
+<important if="using RSpace skills or writing prose">
+
+Apply `$unslop` to prose, documentation, messages, and user-facing copy, including skill output. Preserve commands, identifiers, integration keys, brand names, URLs, and machine-readable values. Exclude source code and structured data.
```
rs-fraser: what is unslop? this seems to be the only mention.

.agents/skills/rspace-dev-stack/SKILL.md
```diff
@@ -1,270 +1,118 @@
...
+for frontend-only runtime checks. Read `docker/dev/README.md` for ports,
+debugging, and hot reload. Requires Docker Engine and Compose v2, not local
+Java, Node, or MariaDB.
 
-## Guardrails (read first)
+Use the stack to reproduce crashes, inspect runtime state, diagnose regressions,
+and verify fixes end to end. Code reading and covered unit/integration behavior
+do not require booting it.
 
-- **Never run `up` (or otherwise launch containers) unless the user explicitly
```
rs-fraser, line 19: We used to have this warning about never running unless asked. Not sure if that was required, and I'm a bit unclear on the wording on whether this will be invoked automatically, but if it is, we should check that other team members are using the Docker stack in their workflow before making it mandatory.

.agents/skills/rspace-browser-tests/SKILL.md
```diff
@@ -1,99 +1,80 @@
...
+
+## Shared infrastructure
+
+Paths below are relative to `src/main/webapp/ui/`.
+
+- `src/__tests__/browserSetup.ts`: shared `worker`, start/reset lifecycle,
+  CDP media reset, `suppressFireAndForget404(...)`.
+- `src/__tests__/mswAppShellHandlers.ts`: default whoami, navigation, analytics,
+  and livechat handlers; survive `resetHandlers()`.
+- `src/__tests__/mocks/inventoryMocks.ts`: `oauthTokenHandler`, `OAUTH_TOKEN`.
```
rs-fraser, line 39: `inventoryMocks.ts` should be `oauthTokenMocks.ts`, and according to claude no longer need setup as they're already in the default worker.

AGENTS.md
```diff
@@ -1,244 +1,155 @@
...
+
+<important if="finishing code or configuration changes">
+
+Run focused tests. Frontend changes require `pnpm tsc`, focused tests for behavior changes, and `pnpm lint` for linted or formatted code. Run the full suite only when broader verification is warranted.
+</important>
+
+<important if="finishing a coding round">
 
-## Running RSpace locally
+Apply `$ponytail` to the final diff. Remove duplicate tests and comments that restate code. Keep a minimal check for non-trivial logic and coverage or comments required for security, accessibility, data safety, constraints, or explicit requirements.
```
rs-fraser: Again, making this mandatory could cause conflicts with local setups.

AGENTS.md
```diff
@@ -1,244 +1,155 @@
...
+</important>
 
-Normal Jetty commands and setup are documented in
-`DevDocs/DeveloperNotes/GettingStarted/GettingStarted.md`.
+<important if="verifying runtime behavior">
```
rs-fraser, line 135: Do these required MCP servers need called out more clearly?

AGENTS.md
```diff
@@ -1,244 +1,155 @@
...
+- Use `rg` and `rg --files` for searches.
+- Only edit vendored code, minified files, build output, `target/`, `dist/`, or `node_modules/` when explicitly targeted.
+- Put plans and scratch notes in gitignored `.claude/`.
+- Add dependencies only when necessary.
+- Update nearby developer documentation when behavior or workflows change.
+</important>
 
-These instructions apply to the whole repository. A nested `AGENTS.md` takes
-precedence for files below it.
+<important if="asked to post or edit a GitHub comment">
```
rs-fraser, line 60: We should be clearer about this. The policy states any AI content should be reviewed by a human and attributed as such. My reading of that is that automated posting would be OK, as long as the human has reviewed what is being posted, and the comment/post is attributed by AI.

Also, this is a change to the workflow some devs currently use so should be made clear to everyone.

rs-fraser (changes requested):
I like the direction of this PR, however I'm not sure we should be mandating the use of ponytail on every change. It feels like a local skill that devs can make use of as part of their individual workflow. 

I do like the idea of trying to reduce the amount of ai slop and over-explanation. Perhaps instead of invoking ponytail here "Apply `$ponytail` to the final diff.", we apply the (currently missing) `unslop` skill to code changes too.

AGENTS.md
tchoi-rs: https://github.com/theclaymethod/unslop - I will add this to the root skills

rs-fraser (approved):

PR #1093, opened 2026-09-07
RSDEV-1501: Add MUI Storybook catalog with interactive documentation

tchoi-rs:
## Description ##
Adds an MUI Storybook. 

Adapted from https://github.com/laststance/mui-storybook - MIT License.


src/main/java/com/researchspace/webapp/config/WebConfig.java
```diff
@@ -111,8 +114,22 @@ public void addInterceptors(InterceptorRegistry registry) {
     registry.addInterceptor(wopiProofKeyValidation).addPathPatterns("/wopi/files/**");
   }
 
+  @Override
+  public void addResourceHandlers(ResourceHandlerRegistry registry) {
+    if (isStorybookEnabled()) {
+      registry
+          .addResourceHandler("/public/storybook/**")
```
rs-nicof, line 121: `/public/**` is mapped to `anon` in `security.xml`, so switching this property on does not put Storybook behind the login page. Anyone who can reach the server can browse it, which is fine locally but worth stating, since the property comment and the docs both read as though it were only "off by default".

Either say plainly that the route is unauthenticated, or hang it off an authenticated prefix so the flag matters less.

src/test/java/com/researchspace/webapp/config/WebConfigStorybookRouteTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+  @Test
+  void doesNotRegisterStorybookRouteWhenDeploymentPropertyIsDisabled() {
+    WebConfig config = configWithStorybookEnabled(false);
+    ResourceHandlerRegistry resources =
+        new ResourceHandlerRegistry(new StaticApplicationContext(), new MockServletContext());
+
+    config.addResourceHandlers(resources);
+
+    assertFalse(resources.hasMappingForPattern("/public/storybook/**"));
```
rs-nicof, line 59: This only checks the resource handler. If someone removes the `isStorybookEnabled()` guard from `addViewControllers`, both tests still pass while an instance with the feature switched off keeps publishing the redirect.

Two lines close it: build a `ViewControllerRegistry`, call `addViewControllers`, and assert nothing is registered for that path.

Jenkinsfile
```diff
@@ -147,7 +149,7 @@ pipeline {
             steps {
                 echo "Building prodRelease .war package"
                  sh '''
-                 ./mvnw clean package -DgenerateReactDist -DskipTests=true \
+                 ./mvnw clean package -DgenerateReactDist -DgenerateStorybook=${STORYBOOK} -DskipTests=true \
```
rs-nicof, line 152: This is the prodRelease stage, so ticking the STORYBOOK box produces a customer release WAR with the developer catalogue inside it. The runtime property still gates serving, so it is not a hole on its own, but combined with the route being unauthenticated it is one property edit away.

Consider hard-coding this stage to false and keeping the flag on the feature-branch stage only.

package.json
```diff
@@ -28,14 +28,21 @@
     "test-e2e": "cd src/main/webapp/ui && cross-env playwright test --config playwright-e2e.config.ts",
     "test-e2e:ui": "cd src/main/webapp/ui && cross-env playwright test --config playwright-e2e.config.ts --ui --project=chromium",
     "test-e2e:api": "cd src/main/webapp/ui && cross-env playwright test --config playwright-e2e.config.ts --project=api",
-    "test-e2e:debug": "cd src/main/webapp/ui && cross-env PW_LOG=trace playwright test --config playwright-e2e.config.ts"
+    "test-e2e:debug": "cd src/main/webapp/ui && cross-env PW_LOG=trace playwright test --config playwright-e2e.config.ts",
+    "storybook": "cd src/main/webapp/ui && storybook dev -p 6006",
+    "storybook:build": "cd src/main/webapp/ui && storybook build",
+    "storybook:test": "cd src/main/webapp/ui && vitest run --config .storybook/vitest.config.ts"
```
rs-nicof: Nothing runs this. It appears in no GitHub workflow, no Jenkins stage and no Maven profile (which only runs `storybook:build`), so the interaction and accessibility checks across the new story files never execute automatically.

Worth wiring into CI, though `CONTRAST.md` notes the suite currently exits nonzero, so it needs a recorded baseline first to be able to catch anything.

src/main/webapp/ui/.storybook/main.ts
```diff
@@ -0,0 +1,16 @@
+import type { StorybookConfig } from "@storybook/react-vite";
+
+const config: StorybookConfig = {
+  stories: ["../src/modules/**/*.stories.@(ts|tsx)"],
```
rs-nicof: This glob only picks up stories under `src/modules`, but most RSpace components live in `src/components`, `src/eln` and `src/Inventory`. A story placed next to one of those silently never appears: no error, just nothing.

Either widen it to `../src/**`, or say in the docs that stories have to live under `src/modules` (the docs currently promise the whole source tree).

src/main/webapp/ui/.storybook/main.ts
```diff
@@ -0,0 +1,16 @@
...
+  addons: ["@storybook/addon-docs", "@storybook/addon-a11y", "@storybook/addon-vitest"],
```
rs-nicof, line 5: Storybook's telemetry is on unless you turn it off, so both `storybook dev` and `storybook build` report project metadata (framework, addon list, package versions) to a third party. Now that the Maven profile runs the build, that includes CI agents.

Adding `core: { disableTelemetry: true }` here switches it off.

src/main/webapp/ui/.storybook/vite.config.ts
```diff
@@ -0,0 +1,17 @@
...
+import { defineConfig } from "vite";
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+export default defineConfig(({ command }) => ({
+  // Keep built asset URLs rooted at the RSpace development route; standalone
+  // Storybook uses its normal root-relative URLs while running locally.
+  base: command === "build" ? "/public/storybook/" : "/",
+  resolve: {
```
rs-nicof, line 12: This is a hand-copied subset of the real Vite config: it carries the `@` alias but none of the app's `define` block (the `process`/`global` shims, the TinyMCE constants) or the legacy i18n plugins.

That works today only because every story imports plain MUI. The first story written for an actual RSpace component will fail in the browser with something like "process is not defined", and it will look like a Storybook bug rather than missing config. Importing those pieces from `vite.config.ts` rather than restating them would keep the two from drifting. Related: the preview decorator wraps stories in the theme only, with no i18n provider or Suspense boundary, so a translated component will suspend and throw.

DevDocs/DeveloperNotes/Storybook.md
```diff
@@ -0,0 +1,100 @@
...
+for the source of the adapted MUI examples.
+
+## Writing a story
+
+Use Storybook's React/Vite types and keep the story close to the component it
+documents:
+
+```tsx
+import type { Meta, StoryObj } from "@storybook/react-vite";
+import { Button } from "@mui/material";
```
rs-nicof: The repo's own lint rule rejects the `@mui/material` barrel import, so anyone copying this template fails `pnpm lint` on their first story. All the real story files in this PR already use the deep import.

```suggestion
import Button from "@mui/material/Button";
```

src/main/webapp/ui/src/modules/common/ui/material-ui/UPSTREAM.md
```diff
@@ -0,0 +1,42 @@
+# Upstream attribution
+
+The Material UI component stories in this directory were adapted from
+[`laststance/mui-storybook`](https://github.com/laststance/mui-storybook) at
+commit [`3e1139552dfb53ade9751ce546d332c984bd338b`](https://github.com/laststance/mui-storybook/tree/3e1139552dfb53ade9751ce546d332c984bd338b).
+The upstream project declares the MIT license. Every adapted source file links
```
rs-nicof, line 6: This names the MIT licence but does not reproduce upstream's copyright line and permission notice, and no LICENSE or NOTICE file was added. MIT requires those to be retained in copies of substantial portions, and this repo is AGPL-3.0, so a large body of MIT code without the notice is the kind of thing an audit picks up.

Pasting the upstream MIT text into this file closes it. The per-file attribution headers are otherwise thorough.

rs-nicof (commented):
Parallel review pass over the Storybook catalogue (layering, security, frontend, i18n and general quality reviewers run concurrently against the branch).

9 items below, limited to the ones I could confirm against the code. The catalogue itself is in good shape: the attribution discipline per file, the `WEB-INF` placement plus `warSourceExcludes`, and `CONTRAST.md` documenting the real accessibility baseline rather than silencing the rule are all the right calls. Most of what follows is about the wiring around it.

src/main/java/com/researchspace/webapp/config/WebConfig.java
tchoi-rs: This behaviour is correct (Storybook is not bundled unless explicitly opt-in, which only happens on dev branches deployed to AWS)

rs-nicof (changes requested):

package.json
tchoi-rs: No point currently in testing Material UI behaviour. Will be introduced down the line when we have actual custom component that's worth testing.

src/main/webapp/ui/.storybook/vite.config.ts
tchoi-rs: Not an issue

Jenkinsfile
tchoi-rs: Not an issue

rs-nicof (approved):

PR #1095, opened 2026-09-07
Honor filestore duplicate strategies consistently for file and stream uploads

tchoi-rs:
-- AI writing below --

## Description ##
Honor filestore duplicate strategies consistently for file and stream uploads. Previously, duplicate handling could overwrite existing files or leave empty and partial files after a failed save.

- Reserve new destinations atomically. On collision, `AS_NEW` uses a bounded UUID filename, retaining extensions up to 20 UTF-8 bytes; a second collision fails without retrying.
- Make `ERROR` reject duplicates without modifying existing contents and preserve `REPLACE` overwrite behavior.
- Remove newly reserved files after write or metadata failures and roll back transactions on `IOException`.
- Support extensionless filenames when copying gallery files.

Includes regression coverage for duplicate strategies, long and Unicode filenames, failed saves, and gallery copies. Existing stored paths remain unchanged; no migration is required.

### DIFF
--- a/src/main/java/com/researchspace/service/impl/InternalFileStoreImpl.java
+++ b/src/main/java/com/researchspace/service/impl/InternalFileStoreImpl.java
@@ -121,6 +121,7 @@ public class InternalFileStoreImpl implements InternalFileStore {
       if (suc == 0) {
         // Keep the existing file intact until the replacement has been written successfully.
         replacement = Files.createTempFile(out.toPath().getParent(), ".replacement-", ".tmp");
+        Files.setPosixFilePermissions(replacement, Files.getPosixFilePermissions(out.toPath()));
       }
       long size;
       try (FileOutputStream output =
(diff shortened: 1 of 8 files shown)


src/main/java/com/researchspace/service/impl/InternalFileStoreImpl.java
```diff
@@ -91,47 +97,56 @@ public void setFileMetadataDao(FileMetadataDao fileMetadataDao) {
...
+    int suc = addMetadata(fileProperty, fnm, behaviourOnDuplicate);
+    if (suc < 0) {
+      return null;
+    }
+    File out = new File(baseDir, fileProperty.getRelPath());
+    Path replacement = null;
+    try {
+      if (suc == 0) {
+        // Keep the existing file intact until the replacement has been written successfully.
+        replacement = Files.createTempFile(out.toPath().getParent(), ".replacement-", ".tmp");
```
rs-fraser, line 123: The permissions of the temp file are `0600` while other files in the filestore are `0644`, meaning new files and replacement files have different permissions. Could fix by setting the permissions before the move.

rs-fraser (commented):
Looks good, left 1 comment about permissions for replacements.

src/main/java/com/researchspace/service/impl/InternalFileStoreImpl.java
tchoi-rs: Fixed in 9a71f6717

rs-nicof (approved):

PR #1102, opened 2026-09-08
Make e2e tests more resilient to GitHub failures

tchoi-rs:
## Description ##
Adds retry attempts to Playwright binary downloads, and disable Playwright recording (we don't look at those)


rs-fraser (approved):
Nothing to add on top of the copilot comments

PR #1110, opened 2026-09-10
RSDEV-1331: make listing sort keys safe by construction

rs-fraser:
## Description ##
Fixes RSDEV-1331, the follow-up to RSDEV-1328 (#1101). Each paginated listing now has its own sort enum in `com.researchspace.model.sort`. The enum's `fromRequest` is the only place a request's `orderBy` string is compared, and every DAO builds its order clause from a `switch` over the enum, so request text can no longer reach HQL, native SQL or a Criteria path. The RSDEV-1328 allowlist and the `fileUsage()` / `recordCount()` carve-out are removed.

The form-search branch of `NameDateFilterImpl` is deleted rather than fixed. It built HQL ordering by `BaseRecord` properties that are all `@Transient`, so it could never have run, and nothing calls it: the one caller of `match()` always passes the name search option.

Client-visible changes:
- Unknown sort key returns 400 (was a silent default, or a 500 on the Criteria listings). Only a listing's own keys are accepted.
- Bare tokens: `owner.lastName` / `owner.username` → `owner`, `originator.username` → `sender`, `communication.creationTime` → `creationTime`, `communication.requestedCompletionDate` → `requestedCompletionDate`, `fileUsage()` / `recordCount()` → `fileUsage` / `recordCount`.
- `fileUsage` and `recordCount` select the sysadmin usage query rather than a column, so the directory and the other user listings now return 400 for them instead of quietly sorting by surname.
- Blank `orderBy` means the listing default instead of no order clause.
- Message and notification listings break ties on the message id. `creationTime` is a second-granularity datetime, so tied rows made page boundaries unstable and left the default listing oldest-first despite asking for newest-first.

The 400 message echoes the rejected key, and CR and LF are stripped from it before it is logged. An earlier version of this branch wrapped `exceptionMessage` in `c:out` on both error pages; running the app showed that was wrong twice over, so it is reverted. `EscapeXmlELResolver` already escapes every String an EL expression resolves, so there was no XSS to fix, and the extra wrapper escaped a second time, showing users the markup of their rejected key as literal text.

What running the app did find is that the 400 never reached the client on a page load. `WEB-INF/pages/error.jsp` declared `isErrorPage`, and Jasper forces a 500 whenever the implicit exception is set, overriding the handler. The ajax fragment was unaffected because it is not an error page. That page is only ever a Spring view (the container error pages use the separate `/error.jsp` at the web root), so the directive is dropped, the block keys off the model message, and the handler now sets the status itself.

## Testing
New tests cover the enums, the 400 mapping and the 500 that ordinary page errors still need, and an MVC test pins the anonymous listing's 400 with an HTML payload in the key. DAO tests pin the rejection of the usage keys on a user-column listing and stable paging across tied `creationTime` values. Verified against a running instance: page load 400, ajax 400, ordinary error 500, listing and workspace 200, and the rejected key escaped exactly once.


rs-nicof (commented):
Ran six parallel reviewers (layering, i18n, security, persistence, frontend, general quality) over the full diff plus the surrounding call sites, and verified each finding against the code before posting.

Seven issues below. The one I would look at first is the bare HQL sort token in `NameDateFilterImpl`, which looks like a runtime break on the workspace form filter for an ordinary unsorted request. The design itself holds up well: I could not find any remaining path where request text reaches a query.

src/main/java/com/researchspace/dao/hibernate/NameDateFilterImpl.java
```diff
@@ -150,19 +150,23 @@ protected String generateFilterStrucDocByFormQuery(
...
+    }
+    RecordSort sort = RecordSort.fromRequest(pgCrit.getOrderBy());
+    switch (sort) {
+      case NAME:
+      case CREATION_DATE:
+      case CREATION_DATE_MILLIS:
+      case MODIFICATION_DATE:
+      case MODIFICATION_DATE_MILLIS:
+        // these queries select the sort columns under their own names
+        sbf.append(" order by " + sort.key() + "  " + pgCrit.getSortOrder());
```
rs-nicof, line 101: This appends a bare `modificationDateMillis` into HQL: `generateFilterStrucDocByFormQuery` builds `from StructuredDocument r ...` and it runs through `createQuery` at lines 119 and 124. On `BaseRecord` that getter is `@Transient`, so only `editInfo.modificationDateMillis` is actually mapped.

What makes it newly reachable is the default: previously a blank `orderBy` added no order clause here at all, and now it resolves to `MODIFICATION_DATE_MILLIS`, so the workspace "filter by form" listing should hit this on an ordinary unsorted request and fail to resolve the property.

The bare token is correct for the two native-SQL callers, which is what the comment above describes, so the fix is probably to qualify only the HQL branch (`r.editInfo.`) or pass the prefix in from the caller.

src/main/webapp/WEB-INF/pages/error.jsp
```diff
@@ -28,7 +28,7 @@
 				</ul>
 				<spring:message code="errors.page.contactSupportNotice"/>
                         <p><spring:message code="errors.page.idAtTimestamp" arguments="${errorId},${tstamp}"/></p>
-                        <pre class="message"> ${exceptionMessage}</pre>
+                        <pre class="message"> <c:out value="${exceptionMessage}"/></pre>
```
rs-nicof: This line is inside the `<% if (exception != null) { %>` guard opened at line 22, but `ControllerExceptionHandler` reaches this view through a normal Spring forward with `response.setStatus(400)` rather than `sendError`, so the container never sets `jakarta.servlet.error.exception` and the implicit `exception` should be null. If that is right, the new message is computed and then dropped, and the user gets an error page with no explanation on a plain page load.

Worth loading a bad `orderBy` in a browser, because the same guard also decides whether the reflected XSS this escaping fixes was reachable here: both depend on the block rendering at all. Note the root `/error.jsp` used by the `web.xml` error-page mappings is a different file, and that one does get `exception` set.

src/main/java/com/researchspace/dao/hibernate/CommunicationDaoHibernateImpl.java
```diff
@@ -414,18 +415,35 @@ private List<Long> getActiveMessageOrRequestDistinctCTIds(
...
+        path = viaTarget ? alias + ".lastStatusUpdate" : messagePath + ".creationTime";
+        break;
+      case CREATION_TIME:
+      default:
+        path = messagePath + ".creationTime";
+        break;
     }
     String direction = SortOrder.ASC.equals(pgCrit.getSortOrder()) ? "asc" : "desc";
-    return " order by " + orderBy + " " + direction;
+    return " order by " + path + " " + direction;
```
rs-nicof: A blank `orderBy` now always produces `order by <alias>.creationTime`, where the old `buildOrderBy` returned an empty string and added no order clause. `creationTime` is a one-second-granularity datetime with no tiebreak.

The relaxed assertion in `UserProfileControllerMVCIT` is the symptom: with two rows in the same second the order is undefined, and since these listings re-run the id query per page, a user paging through can see the same row twice or miss one. Appending `, <alias>.id <direction>` makes the order total and the paging repeatable.

src/main/java/com/researchspace/webapp/controller/ControllerExceptionHandler.java
```diff
@@ -283,6 +284,8 @@ public ModelAndView handleExceptions(
             request, response, e, tstamp, errorId)) {
           logWarningFor(getConstraintViolationMessage((ConstraintViolationException) e), errorId);
         }
+      } else if (e instanceof UnknownSortKeyException) {
+        logWarningFor(e.getMessage(), errorId);
```
rs-nicof, line 288: This logs the rejected sort key verbatim, and the published-documents listing that raises it is reachable without logging in. `log4j2.xml` uses a bare `%m` with no CRLF encoding, so `?orderBy=%0A...` lets an anonymous visitor write fabricated multi-line entries into the log.

This is new here, since an unknown key used to be a silent default and logged nothing. Stripping CR/LF from `requestedKey` in the exception, or logging only the errorId, closes it.

src/test/java/com/researchspace/dao/OrderBySafeByConstructionTest.java
```diff
@@ -0,0 +1,57 @@
...
+  @Test
+  public void everyOrderByReadInQueryCodeIsResolvedThroughASortEnum() throws IOException {
+    List<String> violations = new ArrayList<>();
+    for (Path root : SCANNED_ROOTS) {
+      try (Stream<Path> files = Files.walk(root)) {
+        for (Path file : files.filter(p -> p.toString().endsWith(".java")).toList()) {
+          List<String> lines = Files.readAllLines(file);
+          for (int i = 0; i < lines.size(); i++) {
+            String line = lines.get(i);
+            if (RAW_READ.matcher(line).find() && !RESOLVED_READ.matcher(line).find()) {
```
rs-nicof: The check is `RAW_READ.find() && !RESOLVED_READ.find()` per line, so a single resolved read exempts every other read on the same line. This passes the guard while concatenating raw request text:

```java
String col = Sort.fromRequest(pg.getOrderBy()) == NAME ? "name" : pg.getOrderBy();
```

Counting occurrences instead, so that every `.getOrderBy()` on the line must be immediately preceded by `Sort.fromRequest(`, would hold.

src/test/java/com/researchspace/dao/OrderBySafeByConstructionTest.java
```diff
@@ -0,0 +1,57 @@
...
+ * com.researchspace.model.sort} and a case to the DAO's switch. Do not read the raw string.
+ */
+public class OrderBySafeByConstructionTest {
+
+  private static final List<Path> SCANNED_ROOTS =
+      List.of(
+          Paths.get("src/main/java/com/researchspace/dao"),
+          Paths.get("src/main/java/com/researchspace/admin/service/impl"),
+          Paths.get("src/main/java/com/researchspace/service/audit/search"),
+          Paths.get("src/main/java/com/researchspace/service/impl/AuditManagerImpl.java"));
```
rs-nicof: Two things make this weaker than the docs claim. The roots are opt-in, and `com/axiope/dao` and `com/researchspace/maintenance/dao` already exist outside them (neither reads `getOrderBy()` today, so this is about the next listing rather than a current hole). More importantly nothing asserts the scan matched anything, so renaming `getOrderBy()` away would turn this permanently green while `SecurityAndPermissions.md:155` still promises the build fails on an unguarded read. Asserting a non-zero match count is a cheap guard against that.

Small related point on the doc: "the only place the request string is compared" is not quite accurate while `SearchUtils.sortList` and `sortInventoryList` still compare it with `equalsIgnoreCase`. They only drive in-memory sorting, so the security claim holds, but the wording overstates it.

src/main/java/com/researchspace/dao/hibernate/UserDaoHibernate.java
```diff
@@ -307,15 +308,33 @@ private void applySearchTermFieldsHQL(
...
+      case EMAIL:
+        return "email";
+      case AFFILIATION:
+        return "affiliation";
+      case CREATION_DATE:
+        return "creationDate";
+      case LAST_LOGIN:
+        return "lastLogin";
+      case LAST_NAME:
+      case FILE_USAGE:
```
rs-nicof, line 333: `fileUsage` and `recordCount` are dispatch keys for the sysadmin usage listing rather than sortable columns, but `UserSort` is shared with the directory listing, so `/directory/ajax/userlist?orderBy=fileUsage` is accepted and quietly sorts by surname instead of returning the 400 the PR promises for a key a listing does not offer.

The fallback itself is clearly deliberate given the javadoc. It is the shared enum that makes it visible to listings it was not meant for, so either splitting the usage keys into their own enum or naming here the one caller allowed to reach this branch would keep the contract honest.

src/main/java/com/researchspace/dao/hibernate/NameDateFilterImpl.java
rs-fraser: It also turned out the path was unused so I could delete the form search code in 2425f4a5132889fb540cb54aedf23585a250097f

src/test/java/com/researchspace/dao/OrderBySafeByConstructionTest.java
rs-fraser: Yeah this test is brittle, removed in https://github.com/rspace-os/rspace-web/pull/1110/commits/2425f4a5132889fb540cb54aedf23585a250097f

src/test/java/com/researchspace/dao/OrderBySafeByConstructionTest.java
rs-fraser: Yeah this test is brittle. removed in https://github.com/rspace-os/rspace-web/pull/1110/commits/2425f4a5132889fb540cb54aedf23585a250097f

Docs updated.

src/main/java/com/researchspace/webapp/controller/ControllerExceptionHandler.java
rs-fraser: Fixed in https://github.com/rspace-os/rspace-web/pull/1110/commits/2425f4a5132889fb540cb54aedf23585a250097f

src/main/java/com/researchspace/dao/hibernate/UserDaoHibernate.java
rs-fraser: Fixed in https://github.com/rspace-os/rspace-web/pull/1110/commits/2425f4a5132889fb540cb54aedf23585a250097f and added a test

src/main/java/com/researchspace/dao/hibernate/CommunicationDaoHibernateImpl.java
rs-fraser: Done in https://github.com/rspace-os/rspace-web/pull/1110/commits/2425f4a5132889fb540cb54aedf23585a250097f

src/main/webapp/WEB-INF/pages/error.jsp
rs-fraser: From claude:

> Ran it in a browser. The container sets `jakarta.servlet.error.exception`, so the implicit `exception` was not null and the block did render. `EscapeXmlELResolver` already escapes every String EL resolves, so `${exceptionMessage}` was never injectable and my `c:out` double-escaped it; both reverted. The run did find a real bug: `isErrorPage="true"` makes Jasper force 500 over the handler's 400, so page loads never got the 400. Fixed.

rs-nicof (approved):

tchoi-rs (approved):
Frontend portion looks fine (it's a little bit suspicious that we're not exposing first name and last name as separate columns but it's fine)

PR #1111, opened 2026-09-10
Migrate some JUnit matchers to use AssertJ

tchoi-rs:
## Description ##
Migrates matchers that benefit from AssertJ matchers to use AssertJ.


nhanlon2 (approved):
Looks more readable

PR #1115, opened 2026-09-11
RSDEV-897: Block moving shared records into private notebooks

rs-fraser:
## Description

- Prevent records in shared folders from being moved into private notebooks.
- Hide the Move action in the Shared with me filtered workspace view, where a valid shared-folder move context cannot be selected.
- Allow shared-folder records to move into a notebook only when the notebook and source ACLs grant access to a common non-anonymous user or group other than the record owner.

### DIFF
--- a/src/main/java/com/researchspace/model/record/BaseRecord.java
+++ b/src/main/java/com/researchspace/model/record/BaseRecord.java
@@ -609,7 +609,8 @@ public abstract class BaseRecord
       if (from.isTopLevelSharedFolder()) {
         return false;
       }
-      if (from.isSharedFolder() && !(to.isSharedFolder() || (to.isNotebook() && to.isShared()))) {
+      if (from.isSharedFolder()
+          && !(to.isSharedFolder() || (to.isNotebook() && sharesPrincipalWith(from, to)))) {
         return false;
       }
       if (!from.isSharedFolder() && to.isSharedFolder()) {
@@ -623,6 +624,17 @@ public abstract class BaseRecord
     boolean removed = from.removeChild(this);
     RecordToFolder added = to.addChild(this, u);
     return added != null && removed;
+  }
+
+  private static boolean sharesPrincipalWith(BaseRecord first, BaseRecord second) {
+    return first.getSharingACL().getAclElements().stream()
+        .map(ACLElement::getUserOrGrpUniqueName)
+        .filter(principal -> !ANONYMOUS_USER.equals(principal))
+        .anyMatch(
+            principal ->
+                second.getSharingACL().getAclElements().stream()
+                    .map(ACLElement::getUserOrGrpUniqueName)
+                    .anyMatch(principal::equals));
   }
 
   /**
--- a/src/main/java/com/researchspace/model/record/BaseRecord.java
+++ b/src/main/java/com/researchspace/model/record/BaseRecord.java
@@ -609,7 +609,7 @@ public abstract class BaseRecord
       if (from.isTopLevelSharedFolder()) {
         return false;
       }
-      if (from.isSharedFolder() && !(to.isSharedFolder() || to.isNotebook())) {
+      if (from.isSharedFolder() && !(to.isSharedFolder() || (to.isNotebook() && to.isShared()))) {
         return false;
       }
       if (!from.isSharedFolder() && to.isSharedFolder()) {
--- a/src/test/java/com/researchspace/service/RecordManagerTest.java
+++ b/src/test/java/com/researchspace/service/RecordManagerTest.java
@@ -1048,6 +1048,17 @@ public class RecordManagerTest extends SpringTransactionalTest {
         recordMgr
             .move(doc.getId(), privateNotebook.getId(), g.getCommunalGroupFolderId(), pi)
             .isSucceeded());
+    // sharing the notebook with another group must not make it a valid target
+    Group otherGroup = createGroup("other", pi);
+    addUsersToGroup(pi, otherGroup, anotherUser);
+    Notebook otherGroupNotebook =
+        folderMgr.createNewNotebook(
+            pi.getRootFolder().getId(), "other group notebook", new DefaultRecordContext(), pi);
+    assertTrue(shareNotebookWithGroup(pi, otherGroupNotebook, otherGroup, "write").isPresent());
+    assertFalse(
+        recordMgr
+            .move(doc.getId(), otherGroupNotebook.getId(), g.getCommunalGroupFolderId(), pi)
+            .isSucceeded());
     // attempt to move from shared folder into shared subfolder should be fine though
     Folder sharedSubfolder =
         folderMgr.createNewFolder(g.getCommunalGroupFolderId(), "sharedSubfolder", pi);
@@ -1055,6 +1066,33 @@ public class RecordManagerTest extends SpringTransactionalTest {
     assertTrue(
         recordMgr
             .move(doc.getId(), sharedSubfolder.getId(), g.getCommunalGroupFolderId(), pi)
+            .isSucceeded());
+  }
+
+  @Test
+  public void moveDocumentIntoNotebookWithinProjectGroupFolder() throws Exception {
+    User projectOwner = createAndSaveAPi();
+    User projectMember = createAndSaveRandomUser();
+    initialiseContentWithEmptyContent(projectOwner, projectMember);
+    Group projectGroup =
+        createProjectGroupForUsers(
+            projectOwner, projectOwner.getUsername(), "", projectOwner, projectMember);
+
+    logoutAndLoginAs(projectOwner);
+    Folder sharedFolder =
+        folderMgr.getFolder(projectGroup.getCommunalGroupFolderId(), projectOwner);
+    Notebook sharedNotebook =
+        folderMgr.createNewNotebook(
+            sharedFolder.getId(), "project notebook", new DefaultRecordContext(), projectOwner);
+    assertFalse(sharedNotebook.isShared());
+
+    StructuredDocument doc =
+        createBasicDocumentInRootFolderWithText(projectOwner, "project document");
+    assertTrue(doc.unsafeMove(projectOwner.getRootFolder(), sharedFolder, projectOwner));
+
+    assertTrue(
+        recordMgr
+            .move(doc.getId(), sharedNotebook.getId(), sharedFolder.getId(), projectOwner)
             .isSucceeded());
   }
 
--- a/src/test/java/com/researchspace/service/RecordManagerTest.java
+++ b/src/test/java/com/researchspace/service/RecordManagerTest.java
@@ -1040,6 +1040,14 @@ public class RecordManagerTest extends SpringTransactionalTest {
         recordMgr
             .move(doc.getId(), pi.getRootFolder().getId(), g.getCommunalGroupFolderId(), pi)
             .isSucceeded());
+    // attempt to move from shared folder into pi's private notebook should also fail
+    Notebook privateNotebook =
+        folderMgr.createNewNotebook(
+            pi.getRootFolder().getId(), "private notebook", new DefaultRecordContext(), pi);
+    assertFalse(
+        recordMgr
+            .move(doc.getId(), privateNotebook.getId(), g.getCommunalGroupFolderId(), pi)
+            .isSucceeded());
     // attempt to move from shared folder into shared subfolder should be fine though
     Folder sharedSubfolder =
         folderMgr.createNewFolder(g.getCommunalGroupFolderId(), "sharedSubfolder", pi);


nhanlon2 (commented):
From Claude: review of the move restriction change.

Three findings at High, no Critical. Scope reviewed: `BaseRecord` move validation, the `crudops.js` action visibility change, and the added Java and e2e coverage. CI is green apart from `mocked / e2e (17, firefox, 1)`, whose failures are all Inventory specs unrelated to this PR.

The refactor of `doMove` into `isMoveValid` / `satisfiesMoveRestrictions` is behaviour-preserving for every pre-existing branch, and the negative cases for the new rule are well covered. The three comments below are about the test that cannot fail, the strength of the audience check, and the one positive path with no coverage.

Only High and above posted here, as requested.

src/main/webapp/ui/src/__tests__/e2e/specs/workspace/filters.e2e.ts
```diff
@@ -127,6 +127,8 @@ test.describe(`Workspace filters`, () => {
         await recipient.workspace.open();
         await recipient.workspace.toolbar.toggleFilter("shared");
         await expect(recipient.workspace.table.row(docName)).toBeVisible();
+        await recipient.workspace.table.selectRecord(docName);
+        expect(await recipient.workspace.selectionBar.isActionVisible("Move")).toBe(false);
```
nhanlon2, line 132: From Claude: **High.** This assertion passes on `main` as well, so the one-line change in `crudops.js` has no test that would fail without it.

The Shared filter only ever lists records owned by someone else: `SHARED_WITH_USER_BASE` in `RecordGroupSharingDaoHibernateImpl` filters on `owner.id <> :userId`. Whether `#moveRecords` shows at all is gated by `cantMove`, which reads `isMoveable`, which is the `SEND` permission from `WorkspacePermissionsDTOBuilder` -> `recMgr.canMove` -> `isPermitted(record, SEND, user)`.

For a `LAB_GROUP`, `DefaultPermissionFactory.setUpACLForGroupSharedRootFolder` grants `SEND` only through `ACLElement.createRoleRestrictedGroupACL(grp, perm, PI, RS_LAB_ADMIN)`. The `recipient` here is `roleInGroup: "DEFAULT"` (line 107), so `SEND` is false and Move was already hidden for them before this PR. `RecordManagerTest.moveSearchResultsAsPI` shows the same thing from the other side: `assertTrue(permissionUtils.isPermitted(doc, PermissionType.SEND, pi))` holds only after the share, and only for the PI.

To actually exercise the change, assert from `appUser`'s (the PI's) session, or make the group a `PROJECT_GROUP`, where every member gets the folder-manipulation permission including `SEND`.

Related, and worth deciding on while you are here: hiding `#moveRecords` whenever `sharedFilter` is on also removes moves the backend still permits for those same users, including the shared-folder-to-shared-subfolder move asserted at `RecordManagerTest.java:1066-1069` and both positive cases added in this PR. The server rule is conditional, the UI rule is not.

src/main/java/com/researchspace/model/record/BaseRecord.java
```diff
@@ -593,36 +593,59 @@ public boolean unsafeMove(Folder from, Folder to, User u) throws IllegalAddChild
...
 
-    boolean removed = from.removeChild(this);
-    RecordToFolder added = to.addChild(this, u);
-    return added != null && removed;
+  /**
+   * Whether moving into {@code to} keeps this record visible to at least one user or group that
+   * {@code from} shared it with. The record's own owner does not count, since the owner keeps
+   * access wherever the record goes.
+   */
+  private boolean keepsSourceAudience(Folder from, Folder to) {
```
nhanlon2, line 641: From Claude: **High.** The guard requires one overlapping principal, but the move strips every principal the record shares with `from`.

`doMove` calls `from.removeChild(this)`, which runs `InheritACLFromParentsPropagationPolicy.onRemove`. That computes `parentACL.intersectionWith(removedACL)` and recursively removes all of it (minus the anonymous guest). Its own comment says as much: "WARNING!!! This means that the child record may lose some ACL groups (permissions) that it should rightfully have from some of its other parents!"

So when `from` carries more than one non-owner, non-anonymous principal and `to` matches only one of them, the move is allowed and the rest are dropped silently. That is the same silent unshare RSDEV-897 is about, narrowed rather than closed.

If "at least one" is the deliberate policy, the javadoc should say why partial audience loss is acceptable. If it is not, `destinationAudience.containsAll(audienceToKeep)` is the check that matches what `onRemove` actually does.

Separately: `sharedWithNames` compares raw ACL strings, and role-restricted group elements serialise as `groupUniqueName[PI,RS_LAB_ADMIN]` (see `ACLElement.createRoleRestrictedGroupACL`). A lab group shared folder therefore contributes two distinct entries for a single group. It works today only because the plain `groupUniqueName` element happens to be present on both sides.

src/test/java/com/researchspace/service/RecordManagerTest.java
```diff
@@ -1050,6 +1069,65 @@ public void moveSearchResultsAsPI() {
             .isSucceeded());
   }
 
+  @Test
+  public void moveDocumentIntoNotebookWithinProjectGroupFolder() throws Exception {
```
nhanlon2, line 1082: From Claude: **High.** The most common legitimate move has no test: a lab group shared folder into a notebook shared with that same lab group.

The new coverage is a project group notebook inside the group folder (this test), an individual share into a notebook shared with the recipient (line 1100), plus the negatives in `moveSearchResultsAsPI`: private notebook, and a notebook shared with a different group. Every negative is covered. The lab group positive, which is the path most users will take, is not.

It matters because that case is where the ACL name forms have to line up. The lab group shared folder ACL holds both `g1` and `g1[PI,RS_LAB_ADMIN]`, while `RecordSharingManagerImpl.doSharing` adds a plain `new ACLElement(toShareWith.getUniqueName(), perm)` when a notebook is shared with a group. The overlap depends entirely on the plain `g1` appearing on both sides. If that stops holding, the workflow breaks for every lab group PI and the only feedback is the generic `workspace.move.nothing.moved`.

Concretely, in `moveSearchResultsAsPI`: create a notebook in the PI's root, share it with group `g`, and assert the move from `g.getCommunalGroupFolderId()` succeeds.

nhanlon2:
From Claude: **Suggestions (Low and Medium priority only)**

These are optional. None of them block the PR. The High findings are posted separately as inline comments on [review 5208570170](https://github.com/rspace-os/rspace-web/pull/1115#pullrequestreview-5208570170). Each item below carries its own priority.

---

### S1 — Medium — An empty source audience fails closed

`BaseRecord.keepsSourceAudience` (line 636) ends on `audienceToKeep.stream().anyMatch(destinationAudience::contains)`. When `audienceToKeep` is empty the stream is empty, `anyMatch` is `false`, and the move is refused.

That is the case where the source folder shares with nobody except the record owner and the anonymous guest, so there is no audience left to lose and nothing the move can break. Refusing it is stricter than the rule the javadoc describes.

`FolderTest.moveFromSharedFolderToNotebookIgnoresAnonymousAudience` (FolderTest.java:168) locks this in: both ACLs hold only `ANONYMOUS_USER`, both sets come back empty after filtering, and the test asserts `assertFalse`. Worth deciding explicitly whether an empty audience should block or allow, then saying so in the javadoc either way.

```java
// if allowing:
return audienceToKeep.isEmpty() || audienceToKeep.stream().anyMatch(destinationAudience::contains);
```

### S2 — Medium — A blocked move gives the user no reason

`RecordManagerImpl.move` falls through to `return new ServiceOperationResult<>(null, false);` (RecordManagerImpl.java:287). The two-argument constructor defaults `message` to `""`, so `WorkspaceController.move` takes the `moveSuccessCount == 0` branch, finds no message matching `"into own notebook"`, and shows the generic `workspace.move.nothing.moved`.

`ServiceOperationResult`'s own javadoc says "an object of this class must have a message set if succeeded == false", and the three-argument constructor enforces it with `Validate.isTrue(!isBlank(message), ...)`. The two-argument path sidesteps that.

This is pre-existing, but the PR adds a new and much less obvious reason to land there. A user who has just been told the destination notebook is not shared with the right people sees only "nothing moved". Threading a message out of `doMove` would need a signature change, so a cheaper option is to have `RecordManagerImpl.move` distinguish the restriction failure and pass a specific key.

### S3 — Low — `getAclElements()` is documented as unsafe

`sharedWithNames` (line 644) calls `record.getSharingACL().getAclElements()`. That method's javadoc reads:

> UNSAFE!
> getAclElements should only be used with displaying share icons in the workspace home page.

The new use is read-only and collects into a fresh `HashSet`, so the mutation risk the warning is about does not apply here. It does widen a method that was explicitly scoped to one caller. Either add a short note at the call site explaining why it is safe, or add a narrow accessor such as `RecordSharingACL.getUserOrGroupNames()` and move the anonymous filter into it, which would also put that filter next to the identical one already in `InheritACLFromParentsPropagationPolicy.removeAnonymousFromAcl`.

### S4 — Low — The notebook branch depends on an invisible typing rule

`satisfiesMoveRestrictions` (line 623) reads:

```java
return to.isSharedFolder() || (to.isNotebook() && keepsSourceAudience(from, to));
```

The audience check only runs because a notebook sitting inside a shared folder is never typed `SHARED_FOLDER`. That is guaranteed by `Folder.updateRecordTypeIfAddingSubfolderToSharedFolder`, which skips children with `RecordType.NOTEBOOK`. If that exclusion were ever relaxed, every notebook in a shared folder would take the `to.isSharedFolder()` short circuit and silently skip the new rule.

`moveDocumentIntoNotebookWithinProjectGroupFolder` already depends on this, which is what `assertFalse(sharedNotebook.isShared())` is really pinning down. A one-line comment at the call site would make the coupling visible.

### S5 — Low — Unguarded `getOwner()` on a new path

`keepsSourceAudience` dereferences `this.getOwner().getUniqueName()` unconditionally. The existing owner dereference in `isMoveValid` (line 609) is guarded behind `from.isNotebook()`, so the shared-folder-to-notebook path did not previously require a non-null owner. A persisted record always has one, so this is theoretical, but it is a new unguarded assumption in a method reachable from `RecordManagerImpl.move`.

### S6 — Low — Test name reads opposite to its assertion

`FolderTest.moveFromSharedFolderToNotebookIgnoresAnonymousAudience` asserts `assertFalse(document.move(...))`. The name describes the mechanism, that the anonymous guest is filtered out, while the behaviour under test is that the move is blocked. On a failure the name points the reader at the filter rather than at the outcome. Something like `moveFromSharedFolderToNotebookBlockedWhenOnlyAnonymousAudienceIsShared` states the assertion.

### S7 — Low — `crudops.js` reads a global that React mutates

`crudops.js:200` reads `workspaceSettings.sharedFilter`, which is written by `Toolbar.setWorkspaceSettings` at `Toolbar.tsx:271`. It works, because `calculateOptionDisplay` runs on checkbox selection and therefore after any filter toggle has already run through the toolbar.

The file already depends on this global for `workspaceSettings.searchMode` (lines 22 and 35), so the pattern is not new and the risk is small. Flagging it only because this is now a permission-adjacent decision riding on an untyped global crossing the legacy jQuery / React boundary, with no fallback if the property is absent. `!workspaceSettings?.sharedFilter` would fail safe.

rs-fraser:
> Important
> 
> The new E2E fixture correction is sound, but the existing `BaseRecord` audience-comparison finding remains unresolved.
> 
> **Reviewed changes** since the prior Pullfrog review at `845a6f5b`, including the one new fixture correction and the complete current PR diff.
> 
>     * **Corrected project-group ownership** — Replaced the invalid `PI` role with `GROUP_OWNER`, satisfying the sysadmin API requirement and allowing the Shared-filter E2E scenario to reach its assertion.
> 
> 
> [Pullfrog](https://pullfrog.com)  ｜ [Fix it ➔](https://pullfrog.com/trigger/rspace-os/rspace-web/1115?action=fix&review_id=5210874087) ｜ [View workflow run](https://github.com/rspace-os/rspace-web/actions/runs/34977384422/job/104408486554) ｜ Using `GPT Sol` ｜ [𝕏](https://x.com/pullfrogai)

I looked into this, and cannot reproduce. Explanation from claude:
> Sharing a notebook with a group adds it to the group shared folder, and `InheritACLFromParentsPropagationPolicy.onAdd` unions the folder ACL onto it, so the notebook holds `g[PI,RS_LAB_ADMIN]` as well as `g`. Confirmed by printing the notebook ACL mid-test. `RecordManagerTest.moveSearchResultsAsPI` asserts exactly this move and passes in CI on 28edcde.

src/main/webapp/ui/src/__tests__/e2e/specs/workspace/filters.e2e.ts
rs-fraser: Fixed in https://github.com/rspace-os/rspace-web/commit/845a6f5b109e057ae842a57f0e9eb62eafac50f7. The test now uses a project group, where every member has move permission, so the assertion fails without the crudops change

src/test/java/com/researchspace/service/RecordManagerTest.java
rs-fraser: Fixed in https://github.com/rspace-os/rspace-web/commit/a48131b6af0ff949eb46f778949eafe56cc49eb2. `moveSearchResultsAsPI` now moves the document into a notebook shared with the same lab group

src/main/java/com/researchspace/model/record/BaseRecord.java
rs-fraser: Fixed `containsAll` in https://github.com/rspace-os/rspace-web/commit/a48131b6af0ff949eb46f778949eafe56cc49eb2.

Re. the ACL strings, this is by design. Sharing the doc with group places the doc in the groups shared folder, which causes the doc to inherit the permissions of that folder, so has both `group` and `group[PI.....]` type permissions.

rs-fraser:
> From Claude: **Suggestions (Low and Medium priority only)**
> 
> These are optional. None of them block the PR. The High findings are posted separately as inline comments on [review 5208570170](https://github.com/rspace-os/rspace-web/pull/1115#pullrequestreview-5208570170). Each item below carries its own priority.
> ### S1 — Medium — An empty source audience fails closed
> 
> `BaseRecord.keepsSourceAudience` (line 636) ends on `audienceToKeep.stream().anyMatch(destinationAudience::contains)`. When `audienceToKeep` is empty the stream is empty, `anyMatch` is `false`, and the move is refused.
> 
> That is the case where the source folder shares with nobody except the record owner and the anonymous guest, so there is no audience left to lose and nothing the move can break. Refusing it is stricter than the rule the javadoc describes.
> 
> `FolderTest.moveFromSharedFolderToNotebookIgnoresAnonymousAudience` (FolderTest.java:168) locks this in: both ACLs hold only `ANONYMOUS_USER`, both sets come back empty after filtering, and the test asserts `assertFalse`. Worth deciding explicitly whether an empty audience should block or allow, then saying so in the javadoc either way.
> 
> ```java
> // if allowing:
> return audienceToKeep.isEmpty() || audienceToKeep.stream().anyMatch(destinationAudience::contains);
> ```
> 
> ### S2 — Medium — A blocked move gives the user no reason
> 
> `RecordManagerImpl.move` falls through to `return new ServiceOperationResult<>(null, false);` (RecordManagerImpl.java:287). The two-argument constructor defaults `message` to `""`, so `WorkspaceController.move` takes the `moveSuccessCount == 0` branch, finds no message matching `"into own notebook"`, and shows the generic `workspace.move.nothing.moved`.
> 
> `ServiceOperationResult`'s own javadoc says "an object of this class must have a message set if succeeded == false", and the three-argument constructor enforces it with `Validate.isTrue(!isBlank(message), ...)`. The two-argument path sidesteps that.
> 
> This is pre-existing, but the PR adds a new and much less obvious reason to land there. A user who has just been told the destination notebook is not shared with the right people sees only "nothing moved". Threading a message out of `doMove` would need a signature change, so a cheaper option is to have `RecordManagerImpl.move` distinguish the restriction failure and pass a specific key.
> ### S3 — Low — `getAclElements()` is documented as unsafe
> 
> `sharedWithNames` (line 644) calls `record.getSharingACL().getAclElements()`. That method's javadoc reads:
> 
> > UNSAFE!
> > getAclElements should only be used with displaying share icons in the workspace home page.
> 
> The new use is read-only and collects into a fresh `HashSet`, so the mutation risk the warning is about does not apply here. It does widen a method that was explicitly scoped to one caller. Either add a short note at the call site explaining why it is safe, or add a narrow accessor such as `RecordSharingACL.getUserOrGroupNames()` and move the anonymous filter into it, which would also put that filter next to the identical one already in `InheritACLFromParentsPropagationPolicy.removeAnonymousFromAcl`.
> ### S4 — Low — The notebook branch depends on an invisible typing rule
> 
> `satisfiesMoveRestrictions` (line 623) reads:
> 
> ```java
> return to.isSharedFolder() || (to.isNotebook() && keepsSourceAudience(from, to));
> ```
> 
> The audience check only runs because a notebook sitting inside a shared folder is never typed `SHARED_FOLDER`. That is guaranteed by `Folder.updateRecordTypeIfAddingSubfolderToSharedFolder`, which skips children with `RecordType.NOTEBOOK`. If that exclusion were ever relaxed, every notebook in a shared folder would take the `to.isSharedFolder()` short circuit and silently skip the new rule.
> 
> `moveDocumentIntoNotebookWithinProjectGroupFolder` already depends on this, which is what `assertFalse(sharedNotebook.isShared())` is really pinning down. A one-line comment at the call site would make the coupling visible.
> ### S5 — Low — Unguarded `getOwner()` on a new path
> 
> `keepsSourceAudience` dereferences `this.getOwner().getUniqueName()` unconditionally. The existing owner dereference in `isMoveValid` (line 609) is guarded behind `from.isNotebook()`, so the shared-folder-to-notebook path did not previously require a non-null owner. A persisted record always has one, so this is theoretical, but it is a new unguarded assumption in a method reachable from `RecordManagerImpl.move`.
> ### S6 — Low — Test name reads opposite to its assertion
> 
> `FolderTest.moveFromSharedFolderToNotebookIgnoresAnonymousAudience` asserts `assertFalse(document.move(...))`. The name describes the mechanism, that the anonymous guest is filtered out, while the behaviour under test is that the move is blocked. On a failure the name points the reader at the filter rather than at the outcome. Something like `moveFromSharedFolderToNotebookBlockedWhenOnlyAnonymousAudienceIsShared` states the assertion.
> ### S7 — Low — `crudops.js` reads a global that React mutates
> 
> `crudops.js:200` reads `workspaceSettings.sharedFilter`, which is written by `Toolbar.setWorkspaceSettings` at `Toolbar.tsx:271`. It works, because `calculateOptionDisplay` runs on checkbox selection and therefore after any filter toggle has already run through the toolbar.
> 
> The file already depends on this global for `workspaceSettings.searchMode` (lines 22 and 35), so the pattern is not new and the risk is small. Flagging it only because this is now a permission-adjacent decision riding on an untyped global crossing the legacy jQuery / React boundary, with no fallback if the property is absent. `!workspaceSettings?.sharedFilter` would fail safe.

AI generated:
> Fixed in https://github.com/rspace-os/rspace-web/commit/a48131b6af0ff949eb46f778949eafe56cc49eb2. S1: an empty audience now passes and the javadoc says so; S4: comment added at the call site; S6: test renamed to state the outcome. S2: accepted, deferred to a follow-up ticket since the message plumbing is pre-existing. S3, S5, S7: declined, the use is read-only, owner is already dereferenced unguarded on the neighbouring lines, and the file reads the same global unguarded elsewhere.

rs-fraser:
@pullfrog

src/main/webapp/ui/src/__tests__/e2e/specs/workspace/filters.e2e.ts
```diff
@@ -98,12 +98,13 @@ test.describe(`Workspace filters`, () => {
 
       const recipient = await flowCreateUser("ROLE_USER");
 
-      await test.step("Given a second user exists in a lab group with me, and I own a document", async () => {
+      await test.step("Given a second user exists in a project group with me, and I own a document", async () => {
```
tchoi-rs, line 101: I think this test might have removed the existing e2e check (moving shared document within the shared folders)?

tchoi-rs (changes requested):

src/main/webapp/ui/src/__tests__/e2e/specs/workspace/filters.e2e.ts
rs-fraser: Can you explain further? Is it the change to PROJECT_GROUP that makes you think that?

nhanlon2 (approved):

tchoi-rs (approved):

src/main/webapp/ui/src/__tests__/e2e/specs/workspace/filters.e2e.ts
tchoi-rs: Yep, I've added a new test to cover this.

PR #1117, opened 2026-09-11
Fix UI token authentication across Operate As sessions

tchoi-rs:
## Description ##
Correctly invalidate UI token authentication when an Operate As session ends.

Fixes RSDEV-800


src/main/webapp/ui/src/common/ApiServiceBase.ts
```diff
@@ -54,6 +54,10 @@ class ApiServiceBase {
       error.response.status === 401 &&
       !(error.config.data && !error.config.data.__isRetryRequest)
     ) {
+      // A UI token is bound to the cookie-backed browser session. After an
+      // Operate As release, a token cached by another tab is rejected and must
+      // not be reused for the retry.
+      JwtService.destroyToken();
```
nhanlon2, line 60: This comment is from Claude. On a 401 this now clears the token, gets a new one for whoever the browser session currently is, and replays the original request with it. For a GET or a `delete()` there is no request body, so the guard on line 55 does not stop the replay, and a request started while operating as another user can re-run as the sysadmin. The new `fetchWithUiToken` avoids exactly this by reloading instead of replaying. Worth applying the same rule here, and only destroying the token once a `/whoami` probe confirms the token itself is the problem rather than an ordinary permission denial.

src/main/java/com/researchspace/api/v1/auth/AbstractApiAuthenticator.java
```diff
@@ -49,37 +45,8 @@ public User authenticate(HttpServletRequest request) {
...
-          log.info(
-              "Reusing the current session for API authentication, principal={}",
-              targetUser.getUsername());
-          return subject;
-        }
-      }
-    } catch (UnavailableSecurityManagerException ignored) {
-      // This exception only happens in some tests.
-      // When it does, we know that session  preservation is not possible.
+    if (UserAuthenticationMethod.UI_OAUTH_TOKEN.equals(targetUser.getAuthenticatedBy())) {
```
nhanlon2, line 48: This comment is from Claude. Whether the strict session check runs depends on `authenticatedBy`, which each subclass sets as a side effect inside `findUserForToken()`. Nothing in the abstract method's contract says it must, so a future authenticator that forgets it silently falls back to the stateless login path and reopens this bug, with no compiler or test failure. Consider an explicit abstract `UserAuthenticationMethod authenticationMethod()` that the base class asks for instead.

src/main/java/com/researchspace/api/v1/auth/AbstractApiAuthenticator.java
```diff
@@ -88,9 +55,30 @@ public User authenticate(HttpServletRequest request) {
...
-        || UserAuthenticationMethod.API_KEY.equals(targetUser.getAuthenticatedBy());
+  private User authenticateUiToken(User targetUser) {
+    try {
+      Subject shiroSubject = SecurityUtils.getSubject();
+      if (shiroSubject.getPrincipal() == null) {
+        throw invalidUiToken();
+      }
+
+      Session session = shiroSubject.getSession();
+      User sessionUser = (User) session.getAttribute(SessionAttributeUtils.USER);
```
nhanlon2, line 66: This comment is from Claude. `session.getAttribute` throws `InvalidSessionException` when the browser session has expired or been stopped, and only `UnavailableSecurityManagerException` is caught below. A user who leaves a tab open past session expiry then gets a 500 instead of a 401, and the new `fetchWithUiToken` recovery only reacts to a 401, so it would not kick in. Widening the catch to `ShiroException` and mapping it to `invalidUiToken()` covers both cases.

src/main/java/com/researchspace/api/v1/auth/AbstractApiAuthenticator.java
```diff
@@ -88,9 +55,30 @@ public User authenticate(HttpServletRequest request) {
...
+      if (sessionUser == null || !sessionUser.getUsername().equals(targetUser.getUsername())) {
+        throw invalidUiToken();
+      }
+
+      log.info(
+          "Reusing the current session for API authentication, principal={}",
+          targetUser.getUsername());
+      return sessionUser;
```
nhanlon2, line 74: This comment is from Claude. This returns the `User` cached in the browser session rather than `targetUser`, the fresh copy that `assertLoginAllowed` actually validated. The session copy is a detached snapshot from login time, so role or account changes made since are not reflected, and its `authenticatedBy` is `null` rather than `UI_OAUTH_TOKEN`, which downstream readers such as `CombinedApiAuthenticator` and `AnalyticsManagerImpl` rely on. Suggest keeping the username check against `sessionUser` but returning `targetUser`.

src/main/webapp/ui/src/modules/common/utils/fetchWithUiToken.ts
```diff
@@ -0,0 +1,47 @@
...
+let recovery: Promise<void> | undefined;
+
+async function recoverSession(authorization: string): Promise<void> {
+  // Resource permission failures also use 401. Only recover an invalid token.
+  const currentIdentity = await fetch("/api/v1/userDetails/whoami", {
+    headers: { Authorization: authorization, "X-Requested-With": "XMLHttpRequest" },
+  });
+  if (currentIdentity.status !== 401) return;
+
+  clearStoredToken();
```
nhanlon2, line 12: This comment is from Claude. The token is cleared before a replacement has been obtained, and all three early returns below (lines 16, 18 and 24) then leave the tab with no token and no reload. A transient network blip during recovery silently leaves the app unauthenticated until the user reloads by hand. Moving `clearStoredToken()` down to just before `saveStoredToken(token)` keeps the old token in place until a working one replaces it.

src/main/webapp/ui/src/modules/common/utils/fetchWithUiToken.ts
```diff
@@ -0,0 +1,47 @@
...
+  const tokenResponse = await fetch("/userform/ajax/inventoryOauthToken", {
+    headers: { "X-Requested-With": "XMLHttpRequest" },
+  });
+  if (!tokenResponse.ok) return;
```
nhanlon2, line 16: This comment is from Claude. When the cookie session has actually expired, `/userform/ajax/inventoryOauthToken` returns the login page with status 200, so `tokenResponse.ok` is true and the `.json()` on the next line throws into the outer catch. The user is left on a dead tab instead of being sent to the login page. `AuthStore.authenticate` already handles this by checking for an undefined token and redirecting to `/login`; mirroring that here would close the gap.

src/main/webapp/ui/src/modules/common/utils/fetchWithUiToken.ts
```diff
@@ -0,0 +1,47 @@
...
+  // Reload instead of replaying a pending mutation as the new user.
+  globalThis.location.reload();
+}
+
+/** Native-fetch UI API requests must recover from tokens rejected after Operate As. */
+export async function fetchWithUiToken(url: string, init: RequestInit): Promise<Response> {
+  const response = await fetch(url, init);
+  const authorization = new Headers(init.headers).get("Authorization");
+  if (response.status === 401 && authorization?.startsWith("Bearer ")) {
+    recovery ??= recoverSession(authorization)
```
nhanlon2, line 36: This comment is from Claude. The shared `recovery` promise is not keyed by the token that triggered it. If two requests are in flight with different tokens and the first 401 is a permission denial on a still-valid token, `recoverSession` returns early at line 10, and the concurrent request that failed with a genuinely stale token waits on that no-op and gets its 401 back with no recovery attempted. Queries recover on their next retry, but mutations do not retry by default. Keying the promise by the authorization string fixes it.

src/test/java/com/researchspace/api/v1/auth/OAuthTokenAuthenticatorTest.java
```diff
@@ -1,22 +1,141 @@
...
+
+    ApiAuthenticationException exception =
+        assertThrows(ApiAuthenticationException.class, () -> authenticator.authenticate(request));
+
+    assertEquals("api.errors.authentication.oauthTokenInvalid", exception.getMessageKey());
+    verify(subject, never()).login(any());
+  }
+
+  @Test
+  void runAsDoesNotAllowMismatchedUiToken() {
```
nhanlon2, line 87: This comment is from Claude. `runAsDoesNotAllowMismatchedUiToken` has the same setup and assertion as `mismatchedUiTokenIsRejected` above it, and never stubs `IS_RUN_AS` or `subject.runAs`, so it just runs the same scenario twice. The Operate As path this PR is named for has no unit coverage, while the test name suggests it does. Either build a real run-as fixture or drop the duplicate and rely on the new e2e spec.

nhanlon2 (commented):
This comment is from Claude. Reviewed this PR with six parallel reviewers covering security, layering, persistence, i18n, frontend and general quality. Posting 8 Critical/High findings inline: 3 on the Java authenticator, 4 on the new frontend recovery and axios 401 path, and 1 on test coverage. Lower-priority suggestions are not included here.

PR #1119, opened 2026-09-14
fix(e2e) address ci flaky tests teardown errors

nebay-abraha:
Fixes two real  ci observed e2e issues: PRT-1149

- Teardown cleanup no longer silently swallows real failures. 
- Fixed a flaky assertion in batch user registration. `clickCreateAll()` returned as soon as the click was dispatched, before the AJAX round-trip that populates the results table's status column.

Test:
You should be able to verify the fix, follow the comments on PRT-1149


src/main/webapp/ui/src/__tests__/e2e/api/clients/SysadminClient.ts
```diff
@@ -9,19 +9,38 @@ import type {
 } from "../models/sysadmin";
 import { BaseApiClient } from "./BaseApiClient";
 
+type ManagedUserState = "active" | "released";
+
 export class SysadminClient extends BaseApiClient {
-  readonly createdUserIds: number[] = [];
+  private readonly managedUsers = new Map<string, { id: number; state: ManagedUserState }>();
```
tchoi-rs: We don't read the state so it can be simplified that a new user is added to the `Map<string, number>` then its entry is deleted when it's released

src/main/webapp/ui/src/__tests__/e2e/components/system/accounts/BatchUserRegistrationComponent.ts
```diff
@@ -101,7 +101,14 @@ export class BatchUserRegistrationComponent {
...
 
   async clickCreateAll(): Promise<void> {
+    const batchCreateResponse = this.page
+      .waitForResponse(
+        (res) => res.url().includes("/system/userRegistration/batchCreate") && res.request().method() === "POST",
+        { timeout: 1_000 },
+      )
+      .catch(() => undefined);
     await this.createAllButton.click();
+    await batchCreateResponse;
```
tchoi-rs, line 111: We should also check the UI has updated before proceeding

tchoi-rs (changes requested):

tchoi-rs (approved):

Issue #319, opened 2025-05-01
[BUG]: BioPortal ontology suggestions are no longer appearing in tag selection


mKowalski256:
### Bug Description

After enabling BioPortal ontologies, as described in [documentation](https://documentation.researchspace.com/article/8ujmvpa1no-tagging-documents#bio_portal_ontologies), the tag suggestions should start including the tags coming from BioPortal definitions. I don't see them.

There is actually a [nightly-run unit test](https://github.com/rspace-os/rspace-web/blob/main/src/test/java/com/researchspace/service/impl/BioPortalOntologiesServiceNightlyTest.java) covering that integration, which also started to fail recently (April 2025). Seems like BioPortal no longer accepts the non-browser calls, i.e. the GET query to endpoint:

```
https://bioportal.bioontology.org/search/json_search?q=Tolstoy&subtreerootconceptid=&response=json
```
used to work fine in both browser and server (back-end) call, now it still works fine in browser, but back-end call returns 403 error with message along the line of `Enable JavaScript and cookies to continue`. 

### Steps to Reproduce

Easiest to reproduce by running [BioPortalOntologiesServiceNightlyTest](https://github.com/rspace-os/rspace-web/blob/main/src/test/java/com/researchspace/service/impl/BioPortalOntologiesServiceNightlyTest.java) unit test.

### Expected Behavior

The BioPortal ontology suggestions do appear in tag suggestions, if PI enabled the functionality for their Lab Group.

### Which RSpace version did you encounter the bug with?

2.9.0

### Mobile / Desktop / API

Mobile

### Operating system

Windows

### Browser

Chrome

### Browser version

latest

tilorspace:
@mKowalski256 copilot and I looked into this and came up with this:

**Current problem:**
RSpace’s integration uses a no longer documented BioPortal endpoint without an API key. BioPortal now blocks such backend requests, resulting in failed ontology tag suggestions (HTTP 403).

**API solution:**
Update the integration to use the official BioPortal Data API, which requires an API key and uses a documented endpoint:

https://data.bioontology.org/search?q={your_query}&apikey={YOUR_API_KEY}
API Documentation: https://data.bioontology.org/documentation
Search Endpoint Reference: https://data.bioontology.org/documentation#nav_search

**Alternative solution:**
Consider embedding BioPortal’s browser-based JavaScript widgets (e.g., Autocomplete, Search) directly in the RSpace frontend.

[...]

PR #1120, opened 2026-09-14
fix: BioPortal ontology search integration

nebay-abraha:
## Description ##
Restores BioPortal ontology suggestions by migrating the integration from the deprecated browser-facing endpoint to BioPortal's authenticated REST API.

## Design decisions
- Keep BioPortal as a backend-only integration.
- Preserve the existing ontology metadata format, group permissions, and ontology enforcement behavior.
- Use `suggest=true` for autocomplete BioPortal results.
- Cache successful searches.

## Testing notes
Detailed test steps are in ticket RSDEV-658

Issue:
- https://github.com/rspace-os/rspace-web/issues/319


nhanlon2:
This is from Claude: the findings below are **suggestions only**. They are low and medium priority, so none of them block this PR. Take them or leave them. Anything critical or high would be posted as an inline comment on the code instead, and on this PR there is nothing in those bands, so there are no inline comments to look for.

Reviewed the whole PR, all 4 commits from the merge base `2b97821` to `6ce455d`, not just the recent ones. The head is a merge of `main`, so a review of the last commit alone would have shown a merge and nothing else.

The rewrite is a clear improvement on what it replaces. The old version called `filterTerm.length()` with no null guard and blindly indexed `split("\\|\\|")[1]` on provider output, both of which are gone. The SSRF hardening is done the right way round: `validateApiBaseUri` checks the configured value and then returns a compile-time constant origin, so a stray port, user-info or query string on the property never reaches the request.

### Medium priority

**M1. `src/main/java/com/researchspace/client/BioPortalOntologiesClient.java:37` - `bioportal.api.base.url` is now documented as configurable but is effectively a fixed value** (priority: medium)

`ALLOWED_API_HOST` is hardcoded and `validateApiBaseUri` discards the parsed URI, returning `CANONICAL_API_ORIGIN`. The property therefore behaves as a boolean: match `data.bioontology.org` exactly, or the integration is off.

Both files this PR touches describe it as a real base URL. `RSpaceConfiguration.md` says "The BioPortal REST API base URL used to fetch search suggestions, e.g. `https://data.bioontology.org`", and `defaultDeployment.properties:275` says "the authenticated REST API used to fetch search suggestions". An operator reading either would reasonably point it at an on-prem BioPortal appliance or a staging endpoint. That value parses fine, fails the host check, logs one WARN for the entire life of the JVM because of the `AtomicBoolean` guard, and then every tag autocomplete silently returns nothing.

Whichever way this goes, the code and the documentation should agree:

1. If the allowlist is the point, and given the commit message says this addresses a `java/ssrf` finding it probably is, then document the property as fixed or drop it and inline the constant.
2. If a self-hosted BioPortal is meant to be supported, validate the parsed origin and use it, rather than validating it and throwing it away.

**M2. `src/main/java/com/researchspace/service/impl/BioPortalOntologiesService.java:63` and `BioPortalOntologiesClient.java:145` - the exception detail is dropped, and logged twice** (priority: medium)

Both sites log `e.getClass().getSimpleName()` only. A deserialization failure or a connect timeout produces the bare string `RestClientException` or `ResourceAccessException` with no message, no status, no URL and no stack trace. The previous code logged `e.getMessage()`, so this is a step backwards for anyone debugging a BioPortal outage from logs alone.

They also compound: the client logs a WARN and rethrows, then the service catches and logs another WARN. Tag autocomplete fires per keystroke, so a provider outage produces two WARN lines per keystroke per user.

Log the exception object itself, at one site rather than both. The client is the better place, since it has the URL.

### Low priority

**L1. `src/main/java/com/researchspace/service/impl/BioPortalOntologiesService.java:20` - the constant's name and javadoc contradict the guard that uses it** (priority: low)

`MIN_SEARCH_TERM_LENGTH = 2` is documented as "Below this length, a search term is not eligible", but line 33 reads `normalizedTerm.length() > MIN_SEARCH_TERM_LENGTH`, which also rejects a term of exactly 2.

The code is self-consistent and the test name `shouldReturnEmptyListWhenFilterTermIsTwoCharsOrFewer` describes the real behaviour, so nothing is broken. The risk is the next reader trusting the javadoc and "fixing" the comparison to `>=`, which would change behaviour. Either rename to something like `MAX_IGNORED_TERM_LENGTH` or reword the javadoc.

**L2. `src/main/java/com/researchspace/client/BioPortalOntologiesClient.java:143` - the cached list is mutable and shared** (priority: low)

`doSearch` returns Jackson's `ArrayList` directly, `search` puts that instance into the Guava cache, and every later caller receives the same object. The client test asserts exactly this with `assertSame(first, second)` at line 188.

Nothing mutates it today. But a future caller that sorts or filters the result in place corrupts the cached entry for every other user for the rest of the 10 minute TTL, and that is an unpleasant bug to track down. `Collections.unmodifiableList` before the `put` turns it into an immediate exception instead.

**L3. `src/main/java/com/researchspace/client/BioPortalOntologiesClient.java:69` - the cache key is case-sensitive but the upstream search is not** (priority: low)

`Tolstoy` and `tolstoy` occupy two entries and cost two upstream calls for one result set. With a 200 entry bound fed by per-keystroke prefixes, that is avoidable eviction pressure. Lower-casing the key while still sending the original term in the request would fix it.

**L4. `src/main/resources/deployments/defaultDeployment.properties:274` - typo in the new comment** (priority: low)

`## bioportal.base.url isBioPortal website` is missing a space and a word. Should read "is the BioPortal website".

**L5. `src/main/java/com/researchspace/client/BioPortalLinks.java:17` - `ui` is deserialized and never read** (priority: low)

The only `getUi()` anywhere in the tree is an unrelated match inside the vendored TinyMCE bundle. If it is there for a planned use, that is fine; otherwise it is dead weight in a class whose own sibling comment says only the needed fields are modelled.

### One thing I checked that is not a finding

`SimpleClientHttpRequestFactory` uses `HttpURLConnection`, which follows redirects on GET. The canonical origin stops a misconfigured property from reaching another host, but it would not stop `data.bioontology.org` itself answering with a 302 to somewhere internal. That requires the upstream provider to be hostile or compromised, which is a different threat model from the one this PR set out to close, so I am not raising it as a finding. Mentioning it only so the next person to look at the SSRF surface does not have to rediscover it.

nhanlon2 (approved):

PR #1121, opened 2026-09-14
RSDEV-1231: Inventory biobank operations, without concurrency handling

nhanlon2:
## Summary

This branch removes the compare-and-swap / version-based concurrency handling
that previously guarded Inventory operation requests (stock deductions,
subsample API updates, and related operation-wizard flows).

Concurrency protection for concurrent edits to the same inventory records is
**deferred to a later date**. It is not being reintroduced here; the
associated fields, validators, and tests were stripped along with it. Details
of the removal and rationale are recorded in ADR 0007 and
`DevDocs/DeveloperNotes/InventoryOperationWizard.md`.

Adds Inventory **operations**: a guided way to aliquot, passage, pool, derive, cryopreserve, revive and destroy subsamples, as a wizard in the UI and as seven typed endpoints in the API.

An operation is a transfer. It takes a measured amount out of one or more origin subsamples, creates a new sample from it, and links the result back to where it came from. Doing that by hand today means several API calls and getting the arithmetic right yourself.

## The API

Seven endpoints under `/api/inventory/v1/operations`, one per operation. Each takes the values a user would type and nothing else. The server builds the resulting sample, its fields, its links and its computed values from the operation's definition.

```http
POST /api/inventory/v1/operations/pool
{
  "origins": [
    { "globalId": "SS1234", "amountTaken": { "numericValue": 5, "unitId": 3 } },
    { "globalId": "SS5678", "amountTaken": { "numericValue": 8, "unitId": 3 } }
  ],
  "sampleName": "Pooled lysate",
  "count": 1,
  "eachAmount": { "numericValue": 13, "unitId": 3 }
}
```

```http
POST /api/inventory/v1/operations/destroy
{ "origin": { "globalId": "SS1234" } }
```

**All seven return the same envelope**, because an operation changes the origins as well as producing something:

```json
{ "sample": { ... } | null, "origins": [ ... ] }
```

`sample` is null for Destroy, which creates nothing. `origins` carries each origin as it stands afterwards, so a caller never needs a follow-up GET to see what is left. 201 with a `Location` header for the six creating operations, 200 for Destroy.

**Conventions**

| | |
|---|---|
| Origins | Global ids (`"SS1234"`), so a sample or container id is rejected at parse time. Pool takes `origins`, the other six take a singular `origin`. |
| `amountTaken` | Per origin, because Pool genuinely draws different amounts from each. Absent where the operation takes nothing (Passage) or everything (Destroy). |
| `expectedQuantity` | Optional per origin. Send it and the server compare-and-swaps against the live quantity, returning **409 EDIT_CONFLICT** if a colleague changed it meanwhile. Omit it and the operation takes whatever is there. |
| Required | `eachAmount` and `sampleName` on the six creating operations; `processName` on Derive; `storageTemp` on Cryopreserve. `count` defaults to 1. |
| `templateId` | Numeric, matching `POST /samples`. |

Field names match the operation definitions exactly, so a validation error about `amountTaken` names the field you sent.

## The design

**The server builds the sample, not the client.** The seven operations are declared in `operations_config.json`: their inputs, their bounds, the fields they generate, the links they create and the values they compute. One config-driven core reads that definition, validates the inputs against it, builds the sample, and runs the transfer in a single transaction. The seven endpoints are thin typed wrappers over that core.

This is what keeps the surface small. The rules live in one place rather than seven, and adding an eighth operation is a config entry plus a wrapper, not a new implementation.

**Concurrency is handled in the core.** Origins are locked in ascending id order, each amount is compare-and-swapped against the live quantity, parent totals are recomputed, and the whole effect rolls back together.

**Generated names, links and counters are resolved server-side** in the request locale, so a passage counter continues correctly and a pooled sample's field names are unique even when two origins share a name.

## UI

A wizard under Inventory: pick subsamples, choose an operation, fill in its fields, confirm. The confirmation step previews what will be created before anything is written.

## Notes for reviewers

The branch went through a design change late: the original endpoint had the client assemble the new sample and the backend verify that assembly, which needed roughly 1,300 lines of interpretive validation to police. That path has been removed in favour of the design above. Net effect across those commits is about 3,800 lines deleted.

The seven endpoints are published in `rspace_api_inventory_specs_2_26_0.yaml`, validated against the OpenAPI 3.0 meta-schema, with every example checked against its own schema and a generated Python client producing byte-identical request bodies. The generic config-driven endpoint stays internal and unpublished until user-defined operations ship.

Not yet done: sending the seven against a running instance. Everything else in the gate passed.

## Test plan

- [ ] Fast unit suite (per repo convention, run separately before merge)
- [ ] `InventoryOperationFacadesMVCIT` updated for the removed compare-and-swap path


mKowalski256 (commented):
I checked the functionality on test AWS build, looks good & works fine for me.

As for the code review though, I have a problem with a way how we're publishing new endpoints. `InventoryOperationsApi` is introducing eight new POST endpoints, where 7 are 'typed' and documented in API specs (`/operations/aliquot`, `operations/passage` etc.), with the eight being generic `/operations` endpoint, which is undocumented, but it is what UI is actually calling. The 'typed' endpoints only live in documentation and in MVCIT test.

Couple of issue with that approach:
* the Inventory UI is an API client, dogfooding /api/inventory/v1/ for everything. This feature is maybe a first place where the UI and the published API significantly diverge
* the API surface to test & secure is doubled - the 'typed' endpoints and 'generic' endpoint call the same manager class, but their input validation is different code

I think we should either keep one generic endpoint (and document it), or keep 7 typed ones, and use them in the UI. I'd say let's drop the generic endpoint and rewrite UI wizard so UI calls a 'typed' endpoint specific for each operation.

PR #1123, opened 2026-09-15
Fix required fields in export

### LINKED ISSUE #1114: [BUG]: Required fields are not preserved in the export
### Observed behavior

When I create a form/template with required fields, export it then import it to another RSpace account, the fields are not required in the imported template.

### Steps to Reproduce

1. Create a form with required fields.
2. Create a template from the form.
3. Export the template in XML.
4. Log into another RSpace account and import the template.
5. Open the imported template. The fields are not required.

### Expected Behavior

I expected the fields to remain required after importing the template into another RSpace account.

### Screenshots

_No response_

### Which RSpace version did you encounter the bug with?

2.25.0

### Mobile / Desktop / API

Desktop

### Operating system

linux

### Browser

Chrome

### Browser version

149.0.7827.200 (

### Additional Context

The xml export doesn't hold any information about the form fields being required so that's where I would start.

akos326:
## Description ##
fixes #1114 
In the xml export every field has an attribute called required, which is used to set the new form fields required during the import process.


src/test/java/com/researchspace/archive/model/ArchiveModelFactoryTest.java
```diff
@@ -131,4 +134,18 @@ public void archivalChoiceFieldHasPrintableVersionSetCorrectly() {
...
+  @Test
+  public void testMandatoryFields() {
+    TextFieldForm tff = TestFactory.createTextFieldForm();
+    tff.setMandatory(true);
+    DateFieldForm sff = TestFactory.createDateFieldForm();
+    sff.setMandatory(false);
+    RSForm form = new RSForm("form", "desc", createAnyUser("user"));
+    form.addAllFieldForms(tff, sff);
+    form.setId(1L);
+    ArchivalForm aform = factory.createArchivalForm(form, "asd");
```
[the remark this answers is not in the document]
akos326: Added a round-trip test

src/main/java/com/researchspace/service/archive/FormImporterImpl.java
```diff
@@ -84,11 +84,14 @@ public RSForm makeRSForm(ArchivalDocumentParserRef parserRef, User user) {
                   aff.getDecimalPlace(),
                   aff.getDefaultValue(),
                   FieldType.NUMBER,
-                  aff.getName()));
+                  aff.getName(),
+                  aff.isMandatory()));
           break;
         case STRING:
           String tr = aff.isPassword() ? "true" : "false";
```
rs-nicof: > Written by Claude, an AI coding agent.

Same bug you're fixing, one line up: `StringFieldDTO` only treats `"yes"` as a password (that's what the form editor posts), so `"true"` always evaluates to "not a password" and every password field comes back as an ordinary visible field after a round trip.

```suggestion
          String tr = aff.isPassword() ? "yes" : "no";
```

src/main/java/com/researchspace/service/archive/FormImporterImpl.java
```diff
@@ -98,13 +101,18 @@ public RSForm makeRSForm(ArchivalDocumentParserRef parserRef, User user) {
...
+                  aff.isSortAlphabetic(),
+                  aff.isMandatory()));
           break;
         case CHOICE:
           persistFieldForm(
               form,
               new ChoiceFieldDTO<>(
-                  aff.getOptions(), aff.getMultipleChoice(), aff.getSelection(), aff.getName()));
+                  aff.getOptions(),
+                  aff.getMultipleChoice(),
```
rs-nicof, line 112: > Written by Claude, an AI coding agent.

This value is always `"no"`: the export side hard-codes it in `ArchiveModelFactory.createArchivalFieldForm`'s CHOICE branch (next to a `why din't cfm set` comment) rather than reading `ChoiceFieldForm.isMultipleChoice()`. So a multi-select choice field comes back single-select after import, which is the same family as the bug this PR fixes.

src/main/java/com/researchspace/service/archive/FormImporterImpl.java
```diff
@@ -98,13 +101,18 @@ public RSForm makeRSForm(ArchivalDocumentParserRef parserRef, User user) {
                   aff.getSelection(),
                   aff.getName(),
                   aff.isDisplayAsPickList(),
-                  aff.isSortAlphabetic()));
+                  aff.isSortAlphabetic(),
+                  aff.isMandatory()));
```
rs-nicof, line 105: > Written by Claude, an AI coding agent.

Only the text-field branch is covered end to end by the new test. The other six pass the flag positionally, and this radio call passes three booleans in a row, so a transposition would compile and go unnoticed. `FormImporterTest` already has the Mockito setup, so one test over a form holding each field type would pin all seven cheaply.

src/main/java/com/researchspace/archive/model/ArchiveModelFactory.java
```diff
@@ -224,6 +225,7 @@ public ArchiveUser createArchiveUser(User u) {
   public ArchivalField createArchivalField(Field fd) {
     ArchivalField rc = new ArchivalField();
     rc.setFieldId(fd.getId());
+    rc.setMandatory(fd.isMandatory());
```
rs-nicof: > Written by Claude, an AI coding agent.

Nothing reads this back on import, and `Field.isMandatory()` just delegates to its field form, so the document XML now repeats what the form XML already carries. Fine if it's deliberate metadata for third parties reading the archive format, but if not, the form-side attribute alone fixes the issue.

rs-nicof (commented):

Reviewed the export/import round-trip change with five parallel reviewers (layering, i18n, security, persistence, quality). The core fix is sound and correctly plumbed all the way to the persisted field forms.

Four points below, one of which is a sibling bug on a line this PR rewrites. I checked and discarded a few things others might raise: the new `propOrder` entry is inert but consistent with the existing `fieldId` entry, and old archives stay importable because each archive validates against the schema bundled inside it.

rs-nicof (changes requested):

src/main/java/com/researchspace/service/archive/FormImporterImpl.java
akos326: Fixed in commit 5492d00

src/main/java/com/researchspace/archive/model/ArchiveModelFactory.java
akos326: Removed the xml attribute

src/main/java/com/researchspace/service/archive/FormImporterImpl.java
akos326: Updated the test

src/main/java/com/researchspace/service/archive/FormImporterImpl.java
```diff
@@ -84,11 +84,14 @@ public RSForm makeRSForm(ArchivalDocumentParserRef parserRef, User user) {
...
                   aff.getDefaultValue(),
                   FieldType.NUMBER,
-                  aff.getName()));
+                  aff.getName(),
+                  aff.isMandatory()));
           break;
         case STRING:
-          String tr = aff.isPassword() ? "true" : "false";
-          persistFieldForm(form, new StringFieldDTO<>(aff.getName(), tr, aff.getDefaultValue()));
+          String tr = aff.isPassword() ? "yes" : "no";
```
rs-nicof: > Written by Claude, an AI coding agent.

Both flag fixes are still unproven. The round-trip test only ever asserts the required flag, and every string field it builds is a non-password one, so reverting either fix leaves the suite green. Adding one password field and one multi-select field to that test, and asserting those two flags survive, would pin the part of this PR that is easiest to lose again.

src/main/java/com/researchspace/archive/model/ArchiveModelFactory.java
```diff
@@ -78,6 +78,7 @@ public ArchivalGalleryMetadata createGalleryMetadata(EcatMediaFile mediaFile) {
   ArchivalFieldForm createArchivalFieldForm(FieldForm fm) {
     ArchivalFieldForm rc = new ArchivalFieldForm();
     rc.setFormFieldId(fm.getId());
+    rc.setMandatory(fm.isMandatory());
```
rs-nicof: > Written by Claude, an AI coding agent.

Same family, still open: the importer reads a pick-list flag and a sort-alphabetic flag for radio fields, but nothing ever writes them on the way out, so a pick-list radio still comes back as a plain radio after a round trip. Happy either way on fixing it here or splitting it into a follow-up, but it shouldn't get lost.

src/main/java/com/researchspace/archive/model/ArchiveModelFactory.java
```diff
@@ -115,7 +116,7 @@ ArchivalFieldForm createArchivalFieldForm(FieldForm fm) {
         ChoiceFieldForm cfm = (ChoiceFieldForm) fm;
         rc.setOptions(cfm.getChoiceOptions());
         rc.setSelection(cfm.getDefaultChoiceOption());
-        rc.setMultipleChoice("no"); // why din't cfm set.
+        rc.setMultipleChoice(cfm.isMultipleChoice() ? "yes" : "no"); // why din't cfm set.
```
rs-nicof: > Written by Claude, an AI coding agent.

Small thing: the trailing comment no longer makes sense now that the value is read properly.

```suggestion
        rc.setMultipleChoice(cfm.isMultipleChoice() ? "yes" : "no");
```

rs-nicof (commented):

Thanks @akos326, all four points from the first round are properly addressed and I've resolved those threads.

Second round, now that the scope has grown well beyond the original fix. Nothing here is a defect in what you've written; it's mostly missing coverage on the new surface, plus one gap in the same bug family.

@akos326 **Could you also update the PR description?** It still describes only the required-attribute change, but the PR now also fixes password and multi-select flags being lost, rewrites how the form editor adds field rows, adds a new endpoint, and fixes the Multiple radio never being pre-selected when editing a choice field. That last one is a real user-visible fix worth naming. Reviewers and whoever reads this later will need the wider picture.

CI has now been triggered and is running.

src/main/java/com/researchspace/webapp/controller/RSFormController.java
```diff
@@ -319,6 +319,24 @@ public ModelAndView viewFieldById(@RequestParam("fieldId") long fieldId, Model m
     return new ModelAndView("workspace/editor/editform_fieldforms");
   }
 
+  @GetMapping("ajax/getFieldRow")
+  public ModelAndView viewFieldRow(@RequestParam("fieldId") long fieldId, Model model) {
```
rs-nicof: > Written by Claude, an AI coding agent.

This endpoint is now the only thing that renders a newly added field row, and it has no test. A controller test covering what it returns, the attributes it passes through, and a rejection for someone without edit rights would be cheap insurance against it breaking silently.

src/main/webapp/scripts/pages/workspace/editor/form.js
```diff
@@ -315,6 +318,38 @@ function initNonEditableTinyMCE() {
   });
 }
 
+function appendCreatedFieldRow(data) {
```
rs-nicof: > Written by Claude, an AI coding agent.

Two user-visible changes ride along with this rewrite: a newly added text field is now read-only, and a field added to an already published form no longer gets a Delete button. Both now match what a page reload has always shown, so they look like deliberate corrections rather than regressions, but neither is mentioned anywhere. Could you confirm they're intended and note them in the description?

rs-nicof (changes requested):

PR #1125, opened 2026-09-16
Ignore deleted form fields when validating document API fields

### LINKED ISSUE #1113: PUT /api/v1/documents/{id} always 422s for documents whose form has been edited (deleted FieldForm rows counted in the field-count check)
**Affects:** 2.25.0 and 2.26.0. Reproduced on a live 2.26.0 instance after upgrading, and
`DocumentsApiController.java` and `ApiFieldsHelper.java` are unchanged in `2.25.0...2.26.0`.

### Summary

Once a form has been edited in the UI, **no document created on that form can have its fields updated
through the API again**. Every `PUT /documents/{id}` carrying a `fields` array is rejected with

```
422  "fields" array should have 43 fields, but had 22
```

where 43 is the form's total `FieldForm` row count *including soft-deleted rows* and 22 is the
document's actual field count. Name-only and tag-only PUTs are unaffected.

### Reproduce

1. Create a form with N fields and publish it.
2. Create a document on that form. `POST /documents` succeeds.
3. Edit the form in the UI — rename a field, or add one. RSpace marks the old `FieldForm` rows
   `deleted = 1` and appends the new set to the same form record.
4. `PUT /api/v1/documents/{id}` with `{"fields": [{"id": <a real field id>, "content": "x"}]}`.

Expected: 200, one field updated. Actual: 422 as above.

Every body shape fails, so there is no client-side workaround:

| request body | result |
|---|---|
| `{"fields": [{"id", "content"}]}` (partial, 1 entry) | 422 `should have 43 fields, but had 22` |
| `{"fields": [ …22 positional… ]}` | 422 same |
| `{"fields": [ …21 blanks + 22 live… ]}` | 422 `Provided fields don't match document fields` |
| `{"fields": [ …22 live + 21 blanks… ]}` | 422 same |
| `{"tags": "x"}` / `{"name": "x"}` | 200 |
| partial PUT on a form that was never edited | 200 |

The length check demands 43; the content check accepts only 22.

### Cause

`DocumentsApiController.createNewRevision`:

```java
convertApiFieldsToMatchDocFields(apiDocument, doc.getFields());     // expands to the document's fields (22)
apiFieldsHelper.checkApiFieldsMatchingFormFields(
    apiDocument.getFields(), doc.getForm().getFieldForms(), user);  // validates against 43
```

`ApiFieldsHelper.preValidate` rejects on `apiFields.size() != formFields.size()`.

`AbstractForm.getFieldForms()` carries `@Filter(name = "notdeleted")`, but a Hibernate filter only
applies on a session where it has been enabled. `createNewDocument` obtains the form via
`formMgr.retrieveFormForApiForm(...)`, where the filter is active — 22 rows, and creation works.
`createNewRevision` reaches it through `doc.getForm()`, where it is not — 43 rows, including the
tombstones.

### Suggested fix

The call site wants the non-deleted subset. `AbstractForm` already draws that distinction deliberately —
`getNumActiveFields()` (*"the number of FieldForms defined in this Form that are **not** marked as
deleted"*) alongside `getNumAllFields()` (*"including those marked as deleted"*) — and it has a
`getActiveFieldForms()` that does exactly the right filtering, though it is currently `private`.

We have not built or tested a patch: exposing `getActiveFieldForms()` and passing it at the
`createNewRevision` call site looks like the smallest change, but you will know better whether the
filter belongs there or on the session that loads the form. It may also be worth auditing other
`getFieldForms()` callers that do not come through `FormManager`, since the same latent difference
applies to them.

### Impact

Any API-driven workflow that fills in or corrects a document after creation stops working permanently
the first time its template is edited, with an error message that points at the request body rather than
at the form. Nothing in the API surfaces the discrepancy — `GET /forms/{id}` reports the live field count
(22), identical to the document's — so it can only be diagnosed from the database or from this 422.

On our instance this affects documents on seven forms, and it is the reason two records carry a wrong value that we cannot correct programmatically.

Reported from a self-hosted 2.26.0 instance (Docker, Tomcat 10 / JRE 17, MariaDB 10.11).

rs-fraser:
Fixes #1113

`AbstractForm.getFieldForms()` carries `@Filter(name = "notdeleted")`, but a Hibernate filter only applies on a session where it was enabled, and this codebase enables it in exactly one place: `FormDaoHibernate.get`. `createNewDocument` loads the form through `FormManager` and sees live rows only. `createNewRevision` reaches it through `doc.getForm()`, an association traversal that skips that DAO method, so the count included tombstones and every field update 422'd.

Filter deleted rows in `ApiFieldsHelper` rather than at the call site, which covers both document paths. Inventory callers use the four-argument overload and already pass `getActiveFields()`.

Swept the other five `getFieldForms()` callers: all load via `FormDao.get` first, so none is reachable in a broken state. Added a comment on the accessor instead of changing them.


rs-fraser:
@pullfrog

rs-nicof (approved):

PR #1126, opened 2026-09-16
RSDEV-1325: Search and import an Instrument from a PID registry (UI)

rs-nicof:
## Description ##

Adds a **From PIDINST registry** entry to the Inventory Create menu, gated on `authStore.pidinstEnabled`, opening a dialog that searches the deployment's enabled PID registry and imports a published record as a new Instrument with its PID linked as the identifier. Closes the UI half of RSDEV-1033 (RSDEV-1325); the endpoints came from RSDEV-1326.

The dialog mirrors `FieldmarkImportDialog`: accented theme in Instrument colours, `AppBar variant="dialog"`, `DataGridWithRadioSelection`, toasts through `AlertContext`. Selecting a hit renders a read-only preview of its non-empty fields. A hit already linked to an instrument stays selectable and shows a `GlobalId` chip; Import refuses it with the reason rather than being disabled. On success the dialog closes and navigates to the new instrument.

**Two backend changes were added during review, so this is no longer frontend-only** (`/api/inventory/v1/pidinst/search`):

- **A search needs at least 4 characters**, after trimming, for either provider. Below that a free-text query matches much of the registry. The rule sits in `PidinstLookupManagerImpl` rather than the controller so a direct caller gets it too; it subsumes the controller's old blank check and answers a localized 422 instead of a bare `IllegalArgumentException`.
- **A DataCite search that finds nothing is retried as `doi:*<query>*`.** DataCite indexes the DOI as a keyword, so free text never matched a suffix: `qvtb-aw74` returned nothing although `10.82316/qvtb-aw74` is findable. Verified against api.test.datacite.org: bare suffix 0, `doi:*qvtb-aw74*` 1, `doi:*qvtb*` 1, matching across the `/` and case-insensitively, while `doi:*microscope*` returns 0 so the form invents no hits.

Also fixes a latent bug in `FieldmarkImportDialog`: this MUI X version renders a grid's toolbar slot only when `showToolbar` is passed, so neither dialog offered its Columns button and their hidden columns were unreachable.

## Changes from the internal review pass ##

A six-reviewer pass (layering, i18n, security, persistence, frontend, general quality) ran over the branch. What it changed:

- **A failed search left the previous hits on screen**, under the new query. The result set is now cleared on failure.
- **The results live region was inserted together with its text**, so the first search's summary was never announced. It is now always present and fills in.
- **Closing the dialog mid-import still navigated to the new instrument**, although the confirm promises only that the result will not be shown here. The success toast still links to it.
- **"1 of 1 records found"** - both result strings now carry ICU plurals. A direct PID lookup always returns exactly one hit, so this was the common case, not an edge one.
- **A search that matched nothing no longer reports the registry's total.** "0 of 128 records found at DataCite." sat directly above "No published instrument records match this search."; both were true, since 128 is what the index matched and 0 is what survives filtering to published instruments, but the pair read as a contradiction.
- **The whole Axios error was logged on both failure paths**, and it carries the bearer token in `config.headers`.
- **The `DOI_FRAGMENT` javadoc claimed the allow-list holds no character meaningful in Elasticsearch query syntax.** Both `/` and `-` are meaningful. The class is right, but on evidence rather than on syntax, so the claim is replaced by the evidence: verified against api.datacite.org, `doi:*qvtb/aw74*` answers 200 and `doi:*5281/zenodo*` (12.89M) narrows `doi:*5281*` (12.91M), so the wildcard spans the slash, while `doi:*"broken*` answers 400. Keeping `/` is what lets a pasted prefix/suffix pair match, and it is now pinned by a test rather than left to be removed on the strength of the syntax alone. ADR decision 7 and the connector's own "free-text" javadoc get the same correction.
- **The API spec documented neither the minimum nor the 422**, which are exactly what a direct caller needs.

Four guards could not fail, and each is now proven by mutation:

- Nothing asserted the minimum reaches the message, and `ExternalUpdateMessageKeysTest` feeds every code the same arguments, so "Enter at least Zz-Provider characters" would have passed.
- No test used a query of exactly 4 characters, so `<` and `<=` were indistinguishable.
- The only rejected DOI fragment was a space, which a class widened to admit query-string syntax would still reject.
- The close-during-import branch and the Create-menu wiring had no test at all; the dialog was mocked to `() => null`.

Two things were deleted rather than kept: `InstrumentLink` duplicated the existing `LinkableRecordFromGlobalId`, and a new `pidinst.svg` duplicated the bundled `PIDINST.svg`.

## Design decisions

Recorded as decisions 6 and 7 in `DevDocs/adr/0009-linked-identifiers-for-imported-pids.md`, which also gets its RSDEV-1325 forward references resolved. `CONTEXT.md`'s **PID lookup** entry notes the minimum.

- **Already-linked hits stay selectable** and Import refuses with the reason, rather than the button being disabled, so the user can still read the record and follow the chip to the instrument holding the link.
- **The DataCite retry is guarded**, running only on an empty first page and only for a bare `[A-Za-z0-9._/-]+` query, so user input cannot turn the wildcard into different query syntax and a search that already found something still costs one call.
- **The results grid has a fixed height** rather than `autoHeight`. With 41 hits the preview was pushed far below the fold. A first attempt capped the virtual scroller in `sx`, which looked correct but left `overflow-y: hidden` (MUI sets it inline under `autoHeight`) and stranded rows 6-41.
- **Contrast.** The Search button was unreadable at 2.25:1, and the app bar title at 1.44:1, because this dialog's `contrastText` was near-white where every other accent in the repo is dark, and the theme paints that token on the light accent background. Now 12.11:1 and 7.51:1.

## Testing notes

- `pnpm test src/Inventory/components/PidinstImportDialog.test.tsx src/Inventory/components/__tests__/CreateNew.test.tsx src/Inventory/components/FieldmarkImportDialog.test.tsx`
- `pnpm test-browser src/Inventory/components/PidinstImportDialog.spec.tsx` (3 tests per engine; covers the success toast's link, the refused-import toast and the already-linked popover, which need real layout)
- `mvn test -Dtest='PidinstLookupManagerImplTest,ExternalUpdateMessageKeysTest,PidinstRecordMapperTest' -Dfast=true`
- Walked end to end against a real b2inst-test registry and a DataCite deployment: menu gating both ways, search, preview, import, and the Identifiers card showing the linked PID in state Accepted.

`PidinstLookupApiControllerMVCIT` gains a parameterized test pinning the 422 at the wire for a short, a whitespace-only and an empty query. It is not run locally; CI's `integration-tests` job runs it on both MariaDB versions, and the log confirms it executed rather than being skipped (`Tests run: 9, Failures: 0, Errors: 0, Skipped: 0`).

**Firefox browser tests.** `PidinstImportDialog.spec.tsx` is excluded from the Firefox CI leg, alongside `FieldmarkImportDialog.spec.tsx` and `IgsnTable.spec.tsx`. On Firefox a spec that mounts TinyMCE leaves the next file's requests unintercepted, so MSW passes them through and the component sees a 404 instead of its mock; each file passes alone and fails in that pair. This is a fault in the test tooling rather than in these screens, but it does mean the dialog has no automated Firefox coverage. Tracked as RSDEV-1508.

## Follow-ups raised, deliberately not in this PR

- **RSDEV-1505** - the search names the linked Instrument's Global ID regardless of whether the caller may see it. Pre-existing from RSDEV-1326; this dialog is simply the first consumer to surface it.
- **RSDEV-1506** - `search` is advised into a REQUIRED transaction, so provider HTTP (now up to two calls) holds a pooled DB connection, and a refused short query costs a begin plus rollback.
- **RSDEV-1507** - the grid toolbar and loading overlay now exist in four copies; deduplicating them would have meant touching three components this PR does not otherwise touch.
- **RSDEV-1508** - the Firefox/MSW interception fault above, and the resulting gap in browser coverage.

The 4-character minimum is enforced at both layers: the server answers 422 whatever the caller, and the dialog keeps Search disabled below it and names the minimum under the field, so the ordinary case never makes the round trip.


rs-nicof:
AWS instance: https://rsdev-1325-ui-search-import-pid-650ee552-17.researchspace.com/

src/main/java/com/researchspace/service/inventory/impl/PidinstLookupManagerImpl.java
```diff
@@ -167,6 +187,21 @@ private void searchB2inst(String query, ApiPidinstSearchResult result) {
...
   }
 
+  /**
+   * DataCite indexes the DOI as a keyword, so free text never matches a suffix or part of one, and
+   * a user who pasted half a DOI gets nothing. A {@code doi:*...*} wildcard does match, so an empty
+   * first page is retried that way (ADR 0009 decision 7).
+   */
+  private DataCiteDoiSearchResult searchDataCite(String query) {
+    DataCiteDoiSearchResult hits =
+        dataCiteConnector.searchInstrumentDois(query, MAX_HITS, InventorySettingType.PIDINST);
```
nhanlon2: This comment is from Claude. Two things about this method:

1. The retry below escapes its input via `DOI_FRAGMENT`, but this first call passes the user's raw text into DataCite's `query` parameter, which the new javadoc on `DataCiteConnector` correctly says is Elasticsearch query-string syntax, not plain text. So a search containing a quote or a bracket gets a 400 back and the user sees an error toast instead of "no results". Worth applying the same allow-list (or a reject/escape) to this call too.

2. `DOI_FRAGMENT` is `[A-Za-z0-9._/-]+`, which any ordinary word matches, so the retry fires on every fruitless one-word search rather than only on a pasted DOI fragment. Most misses now cost two registry calls and two cache entries. Restricting the retry to queries that actually look like a DOI (containing `/`, or a `10.` fragment) would keep the case the ADR argues for without the extra traffic.

src/main/webapp/ui/src/Inventory/components/PidinstImportDialog.tsx
```diff
@@ -0,0 +1,616 @@
...
+        }),
+      );
+    } finally {
+      setSearching(false);
+    }
+  }
+
+  async function importRecord(record: PidinstRecord): Promise<boolean> {
+    setImporting(true);
+    closedDuringImport.current = false;
```
nhanlon2: This comment is from Claude. `closedDuringImport` is a single flag reset here on every import, so it can be cleared by a later import while an earlier one is still in flight. Close mid-import, reopen the dialog and start a second import: this line sets the flag back to false, the first request then resolves, sees false, and navigates the user away in the middle of their second import. Tying the flag to the specific request (a per-call token compared when it resolves) avoids it.

src/main/webapp/ui/src/Inventory/components/PidinstImportDialog.tsx
```diff
@@ -0,0 +1,616 @@
...
+                />
+                <Button type="submit" variant="outlined" disabled={searching || importing || queryTooShort}>
+                  {t("common:actions.search")}
+                </Button>
+              </Stack>
+            </Box>
+            {/* rendered even with nothing to say: a live region inserted together with its text
+                is not announced, so the first search's summary would be silent */}
+            <Typography variant="body2" aria-live="polite" role="status">
+              {result && result.hits.length > 0 && (
```
nhanlon2, line 464: This comment is from Claude. The live region only renders text when there is at least one hit, and the zero-result message lives in the grid's `noRowsLabel`, which is not a live region. So a search that finds nothing announces nothing and, to a screen-reader user, the search appears to have done nothing at all. Announcing a short "no records match" here would close the gap.

src/main/webapp/ui/src/Inventory/components/PidinstImportDialog.tsx
```diff
@@ -0,0 +1,616 @@
...
+  const queryTooShort = query.trim().length < MIN_QUERY_LENGTH;
+
+  const providerLabel = (provider: string) =>
+    provider === "PIDINST_B2INST" ? t("pidinstImport.providers.b2inst") : t("pidinstImport.providers.datacite");
+
+  async function runSearch() {
+    const trimmed = query.trim();
+    if (trimmed.length < MIN_QUERY_LENGTH) return;
+    setSearching(true);
+    setSelectedPid(null);
```
nhanlon2, line 284: This comment is from Claude. The selection is cleared before the request but `result` is not, so while the next search is running the summary above the grid still reports the previous query's count and the old rows sit under the overlay. The failure path already clears it for exactly this reason; a `setResult(null)` next to this line covers the in-flight window too.

src/main/webapp/ui/src/Inventory/components/PidinstImportDialog.tsx
```diff
@@ -0,0 +1,616 @@
...
+   * because the in-flight import reads it after the close, outside the render that started it.
+   */
+  const closedDuringImport = React.useRef(false);
+
+  const hits = result?.hits ?? [];
+  const selected = hits.find((hit) => hit.pid === selectedPid) ?? null;
+
+  const queryTooShort = query.trim().length < MIN_QUERY_LENGTH;
+
+  const providerLabel = (provider: string) =>
```
nhanlon2: This comment is from Claude. Any provider value that is not `PIDINST_B2INST` is labelled "DataCite", and the DataCite branch is never asserted with a rendered summary (the only DataCite test returns zero hits, so the summary line does not render). If the provider string is ever wrong or a third provider is added, every affected user quietly sees the wrong registry name with nothing failing. An explicit `PIDINST_DATACITE` check with the raw value as fallback, plus one test with a DataCite hit, would pin it.

nhanlon2 (commented):
This comment is from Claude. Six reviewers (layering, i18n, security, persistence, frontend, general quality) ran over this branch. No critical issues; 5 higher-priority findings are left inline. A longer list of smaller suggestions was reviewed locally and not posted, and the three follow-ups you already raised (RSDEV-1505/1506/1507) were left alone.

src/main/webapp/ui/src/Inventory/components/PidinstImportDialog.tsx
```diff
@@ -0,0 +1,616 @@
...
+    setSearching(true);
+    setSelectedPid(null);
+    try {
+      const { data } = await axios.get<PidinstSearchResult>("/api/inventory/v1/pidinst/search", {
+        params: { query: trimmed },
+        headers: {
+          Authorization: `Bearer ${await getToken()}`,
+        },
+      });
+      setResult(data);
```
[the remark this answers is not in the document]
rs-nicof: Confirmed and fixed in 77cc6c196. The dialog stays mounted while it is closed, so a slow search really did land in the next opening. Fixed at the root with a session counter bumped whenever the dialog resets, which also settles the import case in your other comment. Added the test you asked for: close before the search resolves, reopen, then resolve it.

src/main/webapp/ui/src/Inventory/components/PidinstImportDialog.tsx
```diff
@@ -0,0 +1,616 @@
...
+            {selected && <RecordPreview record={selected} />}
+          </Stack>
+        </DialogContent>
+        <DialogActions>
+          <Stack direction="row" spacing={1} sx={{ ml: "auto" }}>
+            <Button onClick={() => handleClose()}>{t("common:actions.close")}</Button>
+            <ValidatingSubmitButton
+              onClick={() => {
+                if (selected) {
+                  void importRecord(selected).then((success) => {
```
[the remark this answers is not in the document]
rs-nicof: Confirmed and fixed in 77cc6c196. Same root cause as the search one, so the same session counter covers it: a completed import now reports failure to its caller once its session is gone, which is what stops the second close. The test asserts a single close callback, and a separate one starts a later import to prove the abandoned one cannot navigate into it.

src/main/java/com/researchspace/service/inventory/impl/PidinstLookupManagerImpl.java
rs-nicof: Checked both halves against the live registry.

The first one is real, fixed in e5a808523. A quote, a bracket, an unbalanced paren, a caret, a bang and even a dangling AND/OR/NOT all answer 400 rather than an empty page, so the free-text call now escapes what the user typed. The slash is the one exception: escaping it answers 400 where leaving it answers 200, so it would break the pasted fragments the search is for.

The second one I'm leaving as it is. Narrowing the retry to something DOI-shaped would delete the case it exists for, since the documented example is a pasted suffix with neither a slash nor a "10." in it and it only matches through the wildcard. Both calls are cached, so a repeated fruitless search costs nothing extra, and the retry still only fires on an empty first page.

src/main/webapp/ui/src/Inventory/components/PidinstImportDialog.tsx
rs-nicof: Agreed, and fixed in 77cc6c196. The flag is gone entirely, replaced by a counter bumped on every reset, so an abandoned import can no longer be re-enabled by a later one. A test now starts a second import while the first is still in flight.

src/main/webapp/ui/src/Inventory/components/PidinstImportDialog.tsx
rs-nicof: Agreed, fixed in 77cc6c196. The empty result now also goes into the live region, hidden so the sentence is not printed twice on screen.

src/main/webapp/ui/src/Inventory/components/PidinstImportDialog.tsx
rs-nicof: Agreed, fixed in 77cc6c196. Cleared when the search starts rather than only when one fails; the clear on the failure path was redundant after that, so it has gone.

src/main/webapp/ui/src/Inventory/components/PidinstImportDialog.tsx
rs-nicof: Agreed, fixed in 77cc6c196. An unrecognised provider is now named by its own value rather than called DataCite, and there are two tests: one with a DataCite hit and a rendered summary, and one with a provider that has no label, which is the half that actually fails without the change.

rs-nicof:
Answered every comment inline. Summary of this pass:

**Fixed**
- Both async findings from the automated review, plus the related one raised internally, turned out to be one hole: the dialog stays mounted while it is closed, so a search or an import that finished afterwards wrote into whatever session was open next. Replaced the single mid-import flag with a session counter bumped on reset (77cc6c196).
- A search that matches nothing is now announced, the previous result is cleared when the next search starts, and a provider with no label is named by its own value instead of being called DataCite (77cc6c196).
- The DataCite free-text query is escaped (e5a808523). This one was worth the probe: a quote, a bracket, an unbalanced paren, a caret, a bang and a dangling boolean operator all answer 400 from the live registry rather than an empty page, which users would have seen as an error rather than "no results". The slash is deliberately left alone, since escaping that one answers 400 where the raw form answers 200.

**Not fixed**
- Narrowing the wildcard retry to DOI-shaped queries. It would delete the case the retry exists for, and both calls are cached. Reasoning is in the thread.

Every fix is pinned by a test that fails when the fix is reverted, checked by reverting them. The one exception is the DataCite-hit summary test, which is coverage that was missing rather than a pin; the unknown-provider test is the half that fails without the change.

Copilot was quota-blocked again on both review requests, so it has still not reviewed this PR.

src/main/java/com/researchspace/service/inventory/impl/PidinstLookupManagerImpl.java
rs-nicof: Following up on the second half with something better than my own reasoning: the ticket's acceptance criteria name `qvtb` explicitly as a search that must find the record. That is four letters, no slash and no "10.", so restricting the retry the way you suggest would fail a stated acceptance criterion, not just lose an edge case. Leaving it.

src/main/java/com/researchspace/service/inventory/impl/PidinstLookupManagerImpl.java
```diff
@@ -55,6 +56,45 @@ public class PidinstLookupManagerImpl implements PidinstLookupManager {
...
+   * {@code foo^} and {@code zeiss &&} all answer 400 while every escaped form answers 200.
+   *
+   * <p>{@code /} is deliberately absent although query-string syntax reserves it: DataCite answers
+   * 400 for {@code 10.5281\/zenodo} and 200 for {@code 10.5281/zenodo}, so escaping it would break
+   * the pasted DOI fragments this search exists to match. Escaping costs nothing where it is not
+   * needed, checked the same day: {@code \Zeiss} and {@code Zeiss} both answer 71,
+   * {@code spectrometer\*} and {@code spectrometer} both 146.
+   */
+  private static final Pattern DATACITE_RESERVED =
+      Pattern.compile("([\\\\+\\-&|!(){}\\[\\]^\"~*?:])");
```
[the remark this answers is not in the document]
rs-nicof: Checked this against the live registry and it does not hold, so I have not added them.

The 6,539 figure is not evidence of an operator: that is exactly what `Zeiss 4` returns, and what `Zeiss#4`, `Zeiss,4`, `Zeiss;4`, `Zeiss@4` and `Zeiss|4` all return too. The analyser splits on the character and searches two terms. Comparing against `Zeiss` alone measures the second term, not the syntax.

The operators do work, but only attached to a field, and the colon that builds one is already escaped. `publicationYear:>2020` returns 95,411,191; `publicationYear\:>2020`, which is what we actually send, returns 15, identical to the plain `publicationYear 2020`.

Also worth noting for anyone reading this later: these three cannot be escaped anyway. A backslash before one is ignored, so `Zeiss\>4` and `Zeiss>4` return the same thing. Removing them would be the only option, and it would corrupt text the user typed for no gain.

I swept every punctuation character raw against escaped while checking this. Everything escaping can neutralise is neutralised, `/` has to stay raw (escaping it answers 400, and raw it is inert), and `_` is a word character. So the set is complete. Recorded in the ADR and pinned by a test in f41bec230, since adding these three looks right from the syntax alone and has now been asked for once.

nhanlon2 (approved):

tchoi-rs (approved):
Frontend looks fine!

PR #1130, opened 2026-09-16
RSDEV-1231: Inventory biobank operations, operations as typed code not config

nhanlon2:
## Summary

Same feature as #1090 and #1121: Inventory **operations** — a guided way to aliquot, passage, pool, derive, cryopreserve, revive and destroy subsamples, as a wizard in the UI and seven typed endpoints under `/api/inventory/v1/operations`. An operation is a transfer: it takes a measured amount out of one or more origin subsamples, creates a new sample from it, and links the result back to where it came from.

This branch is a design variant of that same epic (RSDEV-1231). Where #1121 keeps a config-driven core (`operations_config.json` + a generic validator + a generic `POST /operations`) reading the same declared rules for all seven operations, this branch deletes that core entirely. **Each operation is its own Java class owning its own value rules and the sample it builds** — there is no generic endpoint left, only the seven typed ones.

## What changed vs. the config-driven design

- `operations_config.json`, the config registry, the generic request builder, the two config-driven validators, and the generic `POST /operations` are gone.
- Value rules split three ways: what a field can say about itself moves onto the request bodies as Jakarta validation annotations; the unit-dependent quantity/temperature rules move to `OperationQuantityRules`; the origin-list rules move to `OperationOriginRules`.
- `ApiInventoryOperationPost` stops being a wire format and becomes the internal carrier the transactional core executes.
- Pool gains `takeAll`, which empties every origin of its live quantity the same way Destroy does.
- Concurrent edits to the same record are handled with an explicit **edit-session lock** taken by the wizard on its origin records (see `DevDocs/adr/0007-operation-wizard-decisions.md` and `DevDocs/DeveloperNotes/InventoryOperationWizard.md`), rather than the compare-and-swap `expectedQuantity` field from #1090/#1121.
- The seven endpoints are documented in `rspace_api_inventory_specs_2_27_0.yaml`.

## Why

Config-driven validation for seven fixed, well-understood operations added an interpretation layer (a registry + two generic validators) for flexibility this feature doesn't currently need — every operation's rules are already known at compile time. This branch trades that flexibility for a plain type per operation: the rules live next to the code they govern, and a validation error names the field the way the Java type does, not the way a config key does.

## Notes for reviewers

- Latest commit is a fix from a six-reviewer parallel pass over the no-configuration refactor commits: a `GlobalIdentifier` aliasing bug that let one Pool request double-decrement an origin under two spellings of the same id (`SS100` / `SS0100`), and an NPE.
- Compare against #1090 (the original config-driven branch) and #1121 (config-driven, no concurrency handling) for the same feature under different design trade-offs.

## Test plan

- [ ] Fast unit suite
- [ ] Spring-transactional tier
- [ ] `InventoryOperationFacadesMVCIT` and related MVCITs for the seven typed endpoints


src/main/java/com/researchspace/service/inventory/operations/OperationQuantityRules.java
```diff
@@ -0,0 +1,164 @@
...
+
+  private OperationQuantityRules() {}
+
+  /**
+   * The quantity each created subsample is given: a real amount unit, storable at 3dp, and greater
+   * than zero. Zero would create a subsample holding nothing while still deducting real stock from
+   * the origin.
+   */
+  public static void createdAmount(ApiQuantityInfo amount, String field, Errors errors) {
+    if (amount == null || amount.getNumericValue() == null) {
```
nhanlon2: This is Codex: "[High] Reject an eachAmount object whose numericValue is missing. The @NotNull constraint applies only to the ApiQuantityInfo object, so a payload such as {numericValue: null, unitId: 3} reaches this return without any error. CreatingOperation then copies that null quantity into every output subsample while the valid amountTaken still decrements the origin, producing empty output stock (or a persistence-time failure) instead of a 400. Only a null amount object should defer to bean validation; a present object with a null number needs a field error here."

src/main/java/com/researchspace/service/inventory/operations/OperationQuantityRules.java
```diff
@@ -0,0 +1,164 @@
...
+   * Celsius bounds. Bounds are compared on the temperature the value DENOTES, not its raw number,
+   * so a value sent in Kelvin or Fahrenheit is judged correctly.
+   */
+  public static void temperature(
+      ApiQuantityInfo temperature,
+      String field,
+      BigDecimal minCelsius,
+      BigDecimal maxCelsius,
+      Errors errors) {
+    if (temperature == null || temperature.getNumericValue() == null) {
```
nhanlon2: This is Codex: "[High] Validate the full temperature value here, not just the wrapper and operation-specific bound. Cryopreserve requires a non-null storageTemp object, but {numericValue: null, unitId: CELSIUS} returns here with no error; likewise -300 C is storable and below the -18 C ceiling, so it passes this method despite being below absolute zero. The later custom sample validator also treats a missing number as unresolved and does not run @ValidTemperature, leaving these malformed requests to fail during entity persistence as a 500. Please reject a missing numericValue and enforce TemperatureValidator/absolute-zero semantics before building the sample."

src/main/java/com/researchspace/service/impl/UserManagerImpl.java
```diff
@@ -343,6 +346,102 @@ public UserPreference setPreference(Preference preference, String value, String
...
+  @CachePut(value = "com.researchspace.model.UserPreference", key = "#subject + 'UI_JSON_SETTINGS'")
+  @CacheEvict(value = CacheNames.INTEGRATION_INFO, key = "#subject + 'UI_JSON_SETTINGS'")
+  public UserPreference mergeUiJsonSetting(String key, String valueJson, String subject) {
+    // Null-checked separately: Set.of is an immutable set, whose contains(null) throws NPE rather
+    // than answering false.
+    if (key == null || !UI_JSON_SETTINGS_KEYS.contains(key)) {
+      throw new IllegalArgumentException(
+          messages.getMessage("errors.preference.invalidKey", new Object[] {key}));
+    }
+    if (valueJson != null && valueJson.length() > MAX_UI_JSON_SETTING_VALUE_CHARS) {
```
nhanlon2, line 402: This is Codex: "[Medium] Bound the serialized merged preference, not only this one value. Every one of the 18 allowlisted keys is accepted with up to 6500 characters, so sequential valid requests can build an ASCII object well above the 65535 limit; JSON key/escaping overhead and utf8mb4 byte width make the stated 56692-character budget unsafe even under the expected-use arithmetic. UserPreference only enforces a 65535 Java-character limit, while a MySQL TEXT column is limited in bytes, so save(user) can still fail with a database exception/500. Serialize after settings.set and reject based on a conservative aggregate byte limit (or use key-specific budgets) before persisting."

src/main/webapp/ui/src/Inventory/components/ContextMenu/ProcessAction.tsx
```diff
@@ -0,0 +1,88 @@
...
+    const origins = selectedResults.filter((r): r is SubSampleModel => r instanceof SubSampleModel);
+
+    const releaseAll = (records: Array<SubSampleModel>) => Promise.allSettled(records.map((r) => r.releaseLock(true)));
+
+    // Holds an edit lock on each origin while the wizard is open.
+    const openWithOriginsLocked = async () => {
+      const results = await Promise.allSettled(origins.map((o) => o.acquireEditLock()));
+      const failure = results.find((r) => r.status === "rejected");
+      if (failure) {
+        await releaseAll(origins.filter((_, i) => results[i].status === "fulfilled"));
```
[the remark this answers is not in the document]
nhanlon2: fixed in: c4a6e325a

`openWithOriginsLocked` now keeps only the origins whose acquisition fulfilled with `LOCKED_OK`, and both release paths use that list: the partial-failure cleanup directly, and the normal close via a ref the open path sets. A `WAS_ALREADY_LOCKED` origin is never released.

Two tests in `ProcessAction.test.tsx` cover it, a mixed `LOCKED_OK` / `WAS_ALREADY_LOCKED` selection with one rejection, and a close after `WAS_ALREADY_LOCKED`. The close-path test was proven by mutation: point `releaseAll` back at `origins` and it fails.

src/main/webapp/resources/rspace_api_inventory_specs_2_27_0.yaml
```diff
@@ -4940,6 +5294,374 @@ components:
...
+    PoolOperation:
+      description: |
+        Pool request. Two or more origins, all holding the same measurement category. `count` is
+        optional and defaults to 1; it may not exceed 100.
+
+        Either every origin carries its own `amountTaken`, or `takeAll` is true and none does.
+      properties:
+        origins:
+          type: array
+          minItems: 2
```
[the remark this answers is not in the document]
nhanlon2: fixed in: c4a6e325a

`maxItems: 100` added to `PoolOperation.origins`, matching `@Size(max = 100)` on `ApiInventoryOperationRequests.Pool`.

src/main/webapp/resources/rspace_api_inventory_specs_2_27_0.yaml
```diff
@@ -2508,6 +2518,343 @@ paths:
...
+        temperature, linked to the origin with an `IsDerivedFrom` field.
+
+
+        `storageTemp` is required and must be at or below -18 degrees Celsius. It is stored as
+        both the storage temperature bounds of the new sample. Send it in any temperature unit;
+        the limit is compared on the temperature it denotes, not on its number.
+
+
+        `cryomedium` is optional and is stored as a Cryomedium field on the new sample. Omitting
+        it still creates that field, with no content.
```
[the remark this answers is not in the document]
nhanlon2: fixed in: c4a6e325a

Aligned the description with the implementation, since `CryopreserveOperation.textFields` and its test both pin no-field-on-blank. It now reads "Omitting it, or sending it blank, creates no such field", which also covers the blank case the old wording missed.

src/main/java/com/researchspace/service/inventory/operations/OperationQuantityRules.java
nhanlon2: fixed in: c4a6e325a

Confirmed and fixed. `createdAmount` now returns early only for a null object; a present one with a null `numericValue` is rejected with `errors.inventory.operation.createdAmountNotPositive`, so bean validation still owns absence and this owns a malformed present value.

Covered at both levels: `OperationQuantityRulesTest.aCreatedAmountObjectSentWithoutANumberIsRejected`, and end to end by `InventoryOperationFacadesMVCIT.anEachAmountCarryingOnlyAUnitIsRejectedRatherThanCreatingEmptyStock`, which posts `eachAmount: {"unitId": <gram>}` to `/operations/aliquot` and asserts a 400 naming `eachAmount` with the origin unchanged.

src/main/java/com/researchspace/service/inventory/operations/OperationQuantityRules.java
nhanlon2: fixed in: c4a6e325a

Both halves confirmed and fixed, in `OperationQuantityRules.temperature`.

A present `storageTemp` with a null `numericValue` is now rejected with `errors.inventory.operation.inputRequired` before anything else reads the number.

For absolute zero the rule now defers to `TemperatureValidator.validate`, which is the canonical semantics you pointed at, raising a new `errors.inventory.temperature.belowAbsoluteZero`. It sits after the unit and 3dp checks, so it is effectively just the absolute-zero judgement, and before the min/max comparison, so it fires even where the operation sets no lower bound. That is exactly the -300 C under Cryopreserve case.

Three tests: `aTemperatureObjectSentWithoutANumberIsRejected` and `aTemperatureBelowAbsoluteZeroIsRejectedEvenInsideTheOperationsBounds`, plus two MVCIT cases posting each shape to `/operations/cryopreserve` and asserting a 400 naming `storageTemp` with the origin unchanged.

src/main/java/com/researchspace/service/impl/UserManagerImpl.java
nhanlon2: fixed in: c4a6e325a

Confirmed and fixed. `mergeUiJsonSetting` now serializes after `settings.set` and rejects on the UTF-8 byte length of the merged object, against 65535 to match `UserPreference.value`, which the schema shows as `text` on a `utf8mb4` table. The per-key ceiling stays as the cheap early check.

Your two separate points both held. The accumulation one was already partly guarded by `SettingsType.validate`; the byte-versus-character one was not guarded at all, and the new test proves it: 25000 euro signs are 25000 Java characters but 75000 bytes, so the old code sailed past every check and would have failed inside the INSERT.

src/main/webapp/ui/src/Inventory/components/ContextMenu/ProcessAction.tsx
```diff
@@ -0,0 +1,99 @@
...
+          error: new AggregateError(results.filter((r) => r.status === "rejected").map((r) => r.reason)),
+          title: t("operations.wizard.originsLocked"),
+          message: t("contextMenu.edit.someoneEditingThem"),
+          beingEditedBy: (name) => t("contextMenu.edit.beingEditedBy", { name }),
+          addAlert: getRootStore().uiStore.addAlert,
+        });
+        if (!shown) console.error("Could not lock the operation origins", failure.reason);
+        return;
+      }
+      taken.current = newlyLocked;
```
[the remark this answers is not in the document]
nhanlon2: This response is from Claude: fixed in: 0cadcc94f05b5438d6b82c9d4addf733d4ebaf31

Confirmed and fixed. An in-flight ref guard now permits one acquisition lifecycle at a time, and it is released at the wizard's close rather than at the acquisition's completion, so an open wizard cannot be re-entered either. A ref rather than state, since both clicks can land before React re-renders.

Covered by a new test in ProcessAction.test.tsx that fires both clicks in the same tick and completes the second (WAS_ALREADY_LOCKED) attempt last, then asserts the lock the first attempt took is still released on close. It fails on the previous code with "expected acquireEditLock to be called 1 times, but got 2 times".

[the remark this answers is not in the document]
nhanlon2: This is Codex: "[High] The `0cadcc94f` serialization fix closes the double-click race, but the successful path still treats an initial `WAS_ALREADY_LOCKED` as safe to proceed. The tracker identifies ownership only by username, so this can be a stale edit form in another tab. A fully sequential UI sequence can lose stock: tab A edits and locks a quantity; tab B opens this wizard, gets `WAS_ALREADY_LOCKED`, and commits a decrement; tab A then saves its explicitly edited stale quantity, which `SubSampleModel.paramsForBackend` sends and the server accepts, overwriting the decrement. `quantityEdited` does not protect this case because the quantity really was edited. Please refuse the initial wizard open when any origin is already locked by this user, or add per-session/tab ownership tokens; preserving the other tab lock without blocking the operation does not prevent its stale write."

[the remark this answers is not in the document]
nhanlon2: This is Codex: "[Medium] The lock lifecycle still is not serialized at close. `onCloseHandler` starts `releaseAll` without awaiting it and immediately clears `busy`; a rapid reopen can therefore receive `WAS_ALREADY_LOCKED` from the old lock and open just before the old DELETE removes that lock. Separately, `extendOriginLocks` is fire-and-forget, so pressing Next/Back and immediately closing can let DELETE win first and the late POST then recreate a five-minute lock after the wizard is gone. Please make renewal/release part of one ordered lifecycle (and keep reopen blocked until release settles), with tests that control POST/DELETE completion order."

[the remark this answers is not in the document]
nhanlon2: This response is from Claude: fixed in: 82e57a9b8f7b9d793d7cbb9dd9337f5bcb39a0a1

Both Codex findings accepted; all three defects are fixed in that commit.

[High] initial WAS_ALREADY_LOCKED. Agreed, and the reasoning about the stale write is correct: the tracker identifies a holder by username alone, so it cannot distinguish this user's other tab, and preserving that lock without blocking the operation does not stop the other tab saving over the decrement. ProcessAction now refuses to open when any origin comes back WAS_ALREADY_LOCKED, releases only the locks it took on the way in, leaves the pre-existing one untouched, and explains why via a new operations.wizard.originsOpenElsewhere message. Per-session ownership tokens would be the more precise fix but need a server-side lock-identity change, so refusing is the correct guard at this layer.

[Medium] close not serialized. Both halves fixed. onCloseHandler now keeps the re-entry guard up until releaseAll settles, so a reopen cannot acquire the lock the close is still giving back. extendOriginLocks now records its renewal, and both close paths (closeUnlessSubmitting and the post-perform close) await it, so a late POST can no longer recreate a five-minute lock behind a wizard that is gone.

Tests, each written first and confirmed failing on the previous code:
- ProcessAction.test.tsx "refuses to open when an origin is already locked by this user elsewhere"
- ProcessAction.test.tsx "blocks reopening until the release started at close has settled", which controls DELETE completion order via a deferred releaseLock
- new OperationWizard.lockLifecycle.test.tsx "does not close until an in-flight lock renewal has settled", which controls POST completion order via a deferred acquireEditLock

One existing test, "leaves a pre-existing same-user lock in place when the wizard closes", was removed rather than updated: its premise was the wizard opening with a WAS_ALREADY_LOCKED origin, which the High fix now forbids. The guarantee it protected, that such a lock is never released by this wizard, is asserted by the new refusal test.

Verification: 366/366 across the Operations and ContextMenu suites, pnpm tsc clean, biome clean on the changed files. No Java changed, so the MVCIT and IT tiers have nothing relevant to exercise and were not run.

src/main/webapp/ui/src/Inventory/components/Operations/OperationWizard.tsx
```diff
@@ -0,0 +1,930 @@
...
+   * Renewal is a POST that recreates a five-minute lock, so it is ordered against the release the
+   * caller performs on close: closing while one is in flight lets the DELETE go first, and the late
+   * POST then re-locks an origin whose wizard is already gone, with nothing left to release it.
+   * The latest renewal is kept here and closeAfterRenewals waits for it (RSDEV-1231).
+   */
+  const renewals = React.useRef<Promise<unknown>>(Promise.resolve());
+
+  /** Extends the edit-session lock on each origin as the wizard's steps complete (RSDEV-1231). */
+  const extendOriginLocks = () => {
+    renewals.current = Promise.allSettled(
```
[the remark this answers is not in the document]
nhanlon2: This response is from Claude: fixed in: 746faf3a5f0f157a00837c128aa3181389971af0

Correct, and the Next/Back/Cancel sequence you describe is exactly how it fails: assigning to the ref replaced the previous batch rather than retaining it, so an earlier renewal still in flight was no longer awaited by close.

Each batch is now folded into the tracked promise instead of replacing it:

    renewals.current = Promise.allSettled([renewals.current, batch]);

I deliberately did not chain the batches to each other (`renewals.current.then(() => ...)`), because that would serialize the renewal POSTs and make a slow one delay every later one. Folding them keeps the requests concurrent while close still waits for all of them.

Added the test you asked for, in OperationWizard.lockLifecycle.test.tsx: "waits for every pending renewal, including one that settles after a later one". Next starts A and Back starts B, B is resolved first, Cancel is then clicked and onClose must not have fired, and only resolving A releases the close. It fails on the previous code with onClose "called 1 times" at the Cancel assertion.

Verification: 367/367 across the Operations and ContextMenu suites, pnpm tsc clean, biome clean on both changed files.

src/main/webapp/ui/src/Inventory/components/Operations/OperationWizard.tsx
```diff
@@ -0,0 +1,931 @@
...
+      origins.map((o) =>
+        o.acquireEditLock().catch((error: unknown) => {
+          console.warn("Could not extend the edit lock on an operation origin", error);
+        }),
+      ),
+    );
+    // Folded in, not replaced: stepping again while a batch is still in flight would otherwise drop
+    // the older one, and it could then land after the release. Batches are not chained to each
+    // other, so a slow renewal never delays a later one; only close waits for them all.
+    renewals.current = Promise.allSettled([renewals.current, batch]);
```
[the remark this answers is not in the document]
nhanlon2: This response is from Claude: fixed in: d1656fc89af36f7e0f7de370145d807531b6bea4

Confirmed: closeAfterRenewals captured the accumulator as it stood, so a Next or Back taken while it waited appended a batch the captured promise never covered, and the release could go first.

You offered two acceptable outcomes and I took the first: prevent any further renewal from being initiated once close starts. A `closing` ref is set by both close paths (closeAfterRenewals and the post-perform close), and extendOriginLocks returns immediately when it is set. A wizard on its way out has no lock left to keep alive, so this is a plain flag rather than a second accumulator, and it makes a repeated Cancel idempotent as a side effect. The alternative, re-checking for batches appended during the wait, needs a loop with no guaranteed fixed point while the user can keep clicking.

That choice changes the regression sequence you specified. Since a transition after Cancel now starts nothing, the assertion is that no renewal B exists at all rather than that resolving A alone fails to close. The new test in OperationWizard.lockLifecycle.test.tsx, "starts no further renewal once close has begun", clicks Next to put renewal A in flight, clicks Cancel, then clicks Back, and asserts acquireEditLock was called exactly once and that resolving A then closes. It fails on the previous code with "expected [ [Function], [Function] ] to have a length of 1 but got 2".

On Back and Next remaining enabled while close is pending: they are now inert with respect to locks, which is the part that could corrupt state. Disabling them as well is a presentation change I have left out of this commit; say the word if you would rather they were greyed out.

Verification: 368/368 across the Operations and ContextMenu suites, pnpm tsc clean, biome clean on both changed files.

src/main/java/com/researchspace/service/inventory/impl/InventoryOperationInFlightOrigins.java
```diff
@@ -0,0 +1,89 @@
...
+   */
+  public Claim claim(Collection<String> originGlobalIds) {
+    Claim claim = new Claim();
+    long now = clock.getAsLong();
+    for (String globalId : new TreeSet<>(originGlobalIds)) {
+      Entry winner =
+          claimed.compute(
+              globalId,
+              (id, entry) ->
+                  entry == null || now - entry.sinceMillis() >= STALE_AFTER_MILLIS
```
nhanlon2: This is Codex: "[High] Do not expire a claim solely because 30 seconds elapsed. The guarded `perform` transaction has no timeout, Pool may update 100 origins, and the UI API client waits up to 360 seconds, so a legitimate operation can still be executing when a second same-user tab replaces its claim here. The edit lock then reports `WAS_ALREADY_LOCKED` to that second server request and allows it through, reopening the overlapping-write/double-decrement path this class exists to prevent. The stated crash cleanup does not justify the timeout: this map is process-local, so a process crash discards the map, while ordinary exceptions already close the try-with-resources claim. Keep a live claim until `Claim.close()` (or use an ownership/heartbeat mechanism that can prove the original request ended) rather than treating age as proof of abandonment."

src/main/java/com/researchspace/service/inventory/impl/InventoryOperationInFlightOrigins.java
```diff
@@ -0,0 +1,100 @@
...
+   * <p>This is a judgement, not a proof: {@code perform} declares no transaction timeout, so
+   * nothing establishes that every live operation finishes inside the limit. A request still
+   * running when its claim is replaced would overlap the request that replaced it. Five minutes is
+   * chosen to put that far outside the range any real operation occupies while still releasing an
+   * origin orphaned by a crash within one edit-lock lifetime, rather than holding it until the
+   * process restarts. Bounding {@code perform} below this limit would turn the judgement into a
+   * guarantee.
+   */
+  static final long STALE_AFTER_MILLIS = Duration.ofMinutes(5).toMillis();
+
```
nhanlon2: This is Codex: "[High] Five minutes still does not make age a safe ownership test. The API client timeout is 360000 ms (six minutes), `perform` remains unbounded, and a 100-origin Pool can legitimately remain live past this constant; a forward wall-clock adjustment can also cross it immediately because this uses `currentTimeMillis`. At that point the second same-user UI request still steals the live claim and overlaps the first. The crash rationale remains incorrect for this process-local map: a JVM crash destroys the map along with the request, while caught failures execute `Claim.close()`. Please restore the owner-release-only implementation from `4be2829b5`, or establish and enforce a real operation timeout below the claim lifetime (using a monotonic elapsed clock). Merely increasing the guess leaves the data-integrity race intact."

src/main/java/com/researchspace/service/inventory/impl/InventoryOperationInFlightOrigins.java
nhanlon2: This response is from Claude: fixed in: d3915e49a0a2aad0d0495bf1c266a287ed22b4f9 and 856078e63d142c04606145a53a290a6918b6a7d2

Accepted. The diagnosis was right: age was being used as proof of abandonment while nothing bounded how long a legitimate operation could run, so a live request could have its claim taken and the overlapping-write path reopened.

The maintainer decided to keep an age-based release rather than adopt owner-release-only, so the fix makes the age test true instead of removing it. performBiobankOperation (renamed from perform, so the name-bound advice cannot be claimed by another manager) now declares timeout="480", and the claim is released at eight minutes ten seconds. The transaction is therefore always rolled back before its claim can be taken, with ten seconds absorbing the rollback and the release. That ordering is pinned by InventoryOperationTransactionRuleTest, which fails if either value moves out of order, and also fails if any other tx:method gains a timeout, since txAdvice is shared by every *Manager advisor.

Your point about the crash rationale was correct and I have stopped relying on it: the surviving justification is that a claim outliving an enforced transaction timeout cannot belong to a live request.

Kept open on the follow-up thread and now also fixed: the wall-clock issue you raise there.

src/main/java/com/researchspace/service/inventory/impl/InventoryOperationInFlightOrigins.java
nhanlon2: This response is from Claude: fixed in: d3915e49a0a2aad0d0495bf1c266a287ed22b4f9 and 856078e63d142c04606145a53a290a6918b6a7d2

Both objections accepted, and this comment was written against the five-minute state, which no longer exists.

You offered two ways out. The maintainer chose the second, so rather than restoring 4be2829b5 the operation timeout is now established and enforced: performBiobankOperation declares timeout="480" and the claim is released at eight minutes ten seconds. An operation is rolled back before its claim can be taken, so a 100-origin Pool can no longer be legitimately live past the constant. InventoryOperationTransactionRuleTest pins the ordering and fails if either value moves independently, or if any other tx:method gains a timeout, txAdvice being shared by every *Manager advisor.

The monotonic point was correct and is fixed in 856078e63: the clock is System.nanoTime, which no calendar adjustment moves, so a forward step can no longer cross the limit in an instant. Only differences are compared and the constant is expressed in nanoseconds.

On the six-minute API client timeout: the client now gives up before the server does. That is deliberate. The claim outlives the client, so a retry after six minutes is refused with the in-progress error while the original is still writing, rather than being allowed to overlap it. Sequencing it the other way, with the client outliving the transaction, is what would let a retry land on a released claim.

The crash rationale you and the earlier comment both challenged is gone. The justification now is only that a claim outliving an enforced transaction timeout cannot belong to a live request.

src/main/java/com/researchspace/service/inventory/impl/InventoryOperationInFlightOrigins.java
```diff
@@ -0,0 +1,98 @@
...
+  /**
+   * A claim this old is taken to have been orphaned, and the next request replaces it.
+   *
+   * <p>Ten seconds past the transaction timeout on {@code performBiobankOperation}, which is what
+   * makes this safe rather than merely likely: a live operation is rolled back at eight minutes, so
+   * no request can still be writing when its claim is released at eight minutes ten. The margin
+   * absorbs the rollback and the release itself. Change one and the other has to move with it;
+   * {@code InventoryOperationTransactionRuleTest} fails if the ordering is ever broken.
+   */
+  public static final long STALE_AFTER_MILLIS = Duration.ofMinutes(8).plusSeconds(10).toMillis();
```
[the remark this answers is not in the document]
nhanlon2: This response is from Claude: fixed in: 3989601ae2f5f79bc9d66fa0e19a0e18e343b14e

Accepted in full, and the criticism of the test is fair: it compared two configured numbers, which is not evidence that a transaction has finished unwinding. A deadline is not an interrupt, and a statement started near it plus cancellation and rollback has no bound I can assert.

Age-based eviction and the transaction timeout that was meant to justify it are both removed. The claim map is now a token per origin, with no clock and no timestamp, and a claim ends only when its owner closes it. There is no takeover point left to get wrong.

The rule test now asserts the opposite of what it did: that no tx:method declares a timeout at all, since txAdvice is shared by every *Manager advisor and one added there would cap unrelated manager calls.

src/main/java/com/researchspace/service/inventory/impl/InventoryOperationInFlightOrigins.java
nhanlon2: This response is from Claude: correction to my previous reply on this thread.

That reply described a design that no longer exists, so please disregard it. I said the age test had been made safe by bounding the transaction at 480 seconds and releasing the claim at eight minutes ten. Both of those are now removed, in 3989601ae2f5f79bc9d66fa0e19a0e18e343b14e.

The outcome is the one you originally asked for: a live claim is kept until Claim.close(). The map holds a token per origin with no clock and no timestamp, so age is no longer treated as proof of abandonment anywhere in this class.

Your reasoning about the crash rationale was right and it is not relied on any more. The reason there is no expiry now is simply that nothing can establish a request has ended except that request releasing its own claim.

src/main/java/com/researchspace/service/inventory/impl/InventoryOperationInFlightOrigins.java
nhanlon2: This response is from Claude: correction to my previous reply on this thread.

My earlier reply chose the second of the two options you offered, establishing an enforced operation timeout below the claim lifetime. That has since been reversed. Please disregard it.

The first option is what is now in the branch: 3989601ae2f5f79bc9d66fa0e19a0e18e343b14e restores owner-release-only claims, exactly as 4be2829b5 had them. The transaction timeout on performBiobankOperation is gone too, because with no expiry there is no limit for it to stay below.

Two things you raised survive as corrections I should record rather than defend. The monotonic-clock point was valid and was fixed in 856078e63 while the age test still existed; that code is now deleted along with the clock itself, so the concern is moot rather than addressed. And my defence of the six-minute API client timeout sequencing no longer applies to anything, since the claim is no longer released on a schedule at all.

src/main/java/com/researchspace/service/inventory/impl/InventoryOperationInFlightOrigins.java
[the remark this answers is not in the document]
nhanlon2: This response is from Claude: fixed in: 3989601ae2f5f79bc9d66fa0e19a0e18e343b14e

This thread was resolved without a reply from me, so for the record: your recommendation was adopted.

Age-based replacement is removed from this process-local guard, as you suggested. A claim is held for the complete lifetime of its owning request and ends only at Claim.close(). The map is a token per origin with no clock and no timestamp.

The branch took a detour before arriving here. It briefly moved the limit to five minutes, then bounded performBiobankOperation at 480 seconds to make the limit provable. Both are gone. Your original point stands as the reason: try-with-resources already releases the claim on normal and exceptional exits, and a process crash discards the map with the JVM, so the orphan case never needed a timer.

DevDocs/adr/0007-operation-wizard-decisions.md
mKowalski256, line 1: This ADR comes from state on 2026-08-24 and is still saying about `operations_config.json` and a thin generic endpoint, which was dropped from final version.

src/main/webapp/resources/rspace_api_inventory_specs_2_27_0.yaml
```diff
@@ -2508,6 +2518,343 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/OperationResult"
+        "400":
+          description: >
+            Rejected, leaving the origin unchanged. `sampleName` or `eachAmount` is missing;
+            `count` is below 1 or above 100; an `amountTaken` other than zero was sent, which
+            this operation cannot honour; the origin is empty; or `templateId` names a template
+            the created sample does not conform to. Each error names the field that was sent.
```
mKowalski256, line 2611: AI agent review comment:

> the "amountTaken other than zero was sent, which this operation cannot honour" part of description is incorrect here - `PassageOperation.takesAmount()` returns false, so `OperationOriginRules.validateAmountTaken` rejects any non-null `amountTaken`, zero included. It also contradicts L2581 in the same operation, which correctly says "Do not send amountTaken". And the schema it refers to, PassageOperation.origin → OperationOrigin (L5399), does not declare the property at all.

better description for 400 response from this endpoint:

> description: >
>   Rejected, leaving the origin unchanged. `sampleName` or `eachAmount` is missing;
>   `count` is below 1, above 100 or not a whole number; an `amountTaken` was sent, which
>   this operation does not accept; the origin is empty; or `templateId` names a template
>   the created sample does not conform to. Each error names the field that was sent.
>

src/main/webapp/resources/rspace_api_inventory_specs_2_27_0.yaml
```diff
@@ -2508,6 +2518,343 @@ paths:
...
+          description: Success. `sample` is null, and the origin is reported holding nothing.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/OperationResult"
+        "400":
+          description: >
+            Rejected, leaving the origin unchanged. The origin is missing, is not identified by a
+            subsample global id, is already empty, or an `amountTaken` was sent that is not the
+            origin's entire remaining quantity. Each error names the field that was sent.
```
mKowalski256: similar issue with 'amountTaken'; the more correct description of 400:

> description: >
>   Rejected, leaving the origin unchanged. The origin is missing, is not identified by a
>   subsample global id, or is already empty; or an `amountTaken` was sent, which this
>   operation does not accept: it always takes whatever the origin holds when it runs.
>   Each error names the field that was sent.

src/main/webapp/resources/rspace_api_inventory_specs_2_27_0.yaml
mKowalski256, line 1: as the `/operations` implementation adds some cases where `409` responses are returned, I think it would be worth for the seven new endpoints to document that `409` in `responses` section.

mKowalski256 (approved):
I reviewed the backend code up to API level, and did some testing on latest AWS build - looks good to me. I've added a few minor comments around documentation.

src/main/webapp/resources/rspace_api_inventory_specs_2_27_0.yaml
nhanlon2: This response is from Claude: fixed in: 96b115bcc81649b899d08e2e8b2c235feb482990

Confirmed and adopted your wording. OperationOriginRules.validateAmountTaken rejects on `origin.getAmountTaken() != null`, so for an operation whose takesAmount() is false the value is irrelevant and zero is rejected like any other. "other than zero" described a rule that does not exist, and as you note it contradicted the operation's own description a few lines above and the schema, which does not declare the property.

I checked the count clause before taking it: the request body carries @Min(1), @Max(100) and @Digits(fraction = 0), so "below 1, above 100 or not a whole number" is exactly right, where the old text omitted the third.

src/main/webapp/resources/rspace_api_inventory_specs_2_27_0.yaml
nhanlon2: This response is from Claude: fixed in: 96b115bcc81649b899d08e2e8b2c235feb482990

Adopted verbatim. Destroy's old text said a 400 followed from an `amountTaken` that was "not the origin's entire remaining quantity", which implies sending the right amount would be accepted. It would not: DestroyOperation.takesAmount() is false, so the same null check rejects any value. Your phrasing also explains why, which the old text did not.

src/main/webapp/resources/rspace_api_inventory_specs_2_27_0.yaml
nhanlon2: This response is from Claude: fixed in: 96b115bcc81649b899d08e2e8b2c235feb482990

Agreed, and done for all seven. Two guards raise a 409 on these endpoints, and neither was documented:

- InventoryOperationInProgressException, when another in-flight request already holds one of the origins
- InventoryEditLockHeldException, when another user's edit session holds one

Both map to 409 EDIT_CONFLICT in ApiControllerAdvice. Since the two are indistinguishable to a caller in what they should do about it, they share one `OperationOriginBusy` response component that covers both causes, states that no origin was changed, and says to retry once the other request or edit session finishes. Each of the seven endpoints now references it, ordered after 404.

DevDocs/adr/0007-operation-wizard-decisions.md
nhanlon2: This response is from Claude: fixed in: 96b115bcc81649b899d08e2e8b2c235feb482990

You are right that a reader is misled. The ADR does already carry the corrections, in "Amended 2026-09-11: the seven typed endpoints are the public API" and "Amended 2026-09-16: operations are code, not config", but those sit around 400 lines below the Foundation section, so the superseded design is what you meet first and the correction is only found if you read to the end.

Rather than rewrite the 2026-08-24 text, I marked it. Status now says to read the amendments first and names what did not survive, and each superseded section carries a banner where it begins linking to the amendment that replaced it: Foundation, "The one thing a future reader must know first", "Why the backend validates against that config copy", "Operation definitions are validated at registry construction" and "Computations are a dev-only code registry". I kept the original prose because the amendments argue against that reasoning and stop making sense without it.

While marking it I found one more staleness you did not flag: "Origin quantity model" still described `amountMode` and a per-origin `expectedQuantity` as being on the wire. D5 removed both, and neither appears anywhere in the Java or the spec now, so those paragraphs are marked too. The `amountTaken` rules above them are still current and are left unmarked.

If you would rather the superseded sections were deleted outright than marked, say so and I will cut them.

DevDocs/adr/0007-operation-wizard-decisions.md
```diff
@@ -0,0 +1,529 @@
...
+(`QuantityInfo`) is rejected the same way, not rounded: the stored decrement
+would differ from the validated one (0.0004 ml would take nothing). Created amounts are independent of the amount taken (fresh medium
+can be added during Derive). A zero decrement is a complete no-op, for
+operations that link to an origin without consuming it (Passage).
+
+> **Superseded from here to the end of this section** by D5 of
+> [Amended 2026-09-16](#amended-2026-09-16-operations-are-code-not-config):
+> `amountMode` and `expectedQuantity` are gone from the wire entirely, and Pool's
+> take-all is `takeAll: true` on the body (D4). The `amountTaken` rules above still
+> hold.
```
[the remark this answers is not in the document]
nhanlon2: This response is from Claude: fixed in: 80a5b28c18a321062f8c4e128a62ce25fda76cd7

Both parts accepted; the marker was mine from the previous commit and it was too wide.

It now reads "Superseded, the next two paragraphs only" and says the section resumes being current from "Off by default" onwards, so the availability gate, the edit-session lock and the in-flight claim are no longer swept up.

The zero-decrement point is the better catch, because my marker actively protected it by asserting the amountTaken rules above still held. They did not, entirely. OperationOriginRules rejects any amountTaken for an operation whose takesAmount() is false, and requires signum() > 0 for one where it is true, so zero is invalid on every path and nothing treats it as a no-op. That sentence now states both rules instead.

src/main/webapp/resources/rspace_api_inventory_specs_2_27_0.yaml
```diff
@@ -3839,6 +4201,15 @@ components:
         application/json:
           schema:
             $ref: "#/components/schemas/Error"
+    OperationOriginBusy:
+      description: >
+        The operation did not start, and every origin is unchanged. Either another request is
+        already acting on one of the origins, or another user's edit session holds one of them.
+        The message names the origin. Retry once that request or edit session has finished.
```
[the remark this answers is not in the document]
nhanlon2: This response is from Claude: fixed in: 80a5b28c18a321062f8c4e128a62ce25fda76cd7

Correct, and your suggestion is applied as written. withOriginsLocked adds each origin's parent sample to the sorted lock set and raises InventoryEditLockHeldException with whichever global id it could not take, so the message can name an SA record while my description promised an origin.

Worth noting the two causes differ in scope, and your wording happens to keep that straight: the in-flight claim covers the origin global ids only, so "another request is already acting on one of the origins" stays accurate for that path, while the edit-session lock is the one that can also block on a parent sample.

nhanlon2:
@pullfrog

DevDocs/adr/0007-operation-wizard-decisions.md
```diff
@@ -0,0 +1,531 @@
...
+can be added during Derive). A zero decrement is no longer accepted anywhere
+(amended 2026-09-16): an operation that takes an amount requires it to be
+greater than zero, and one that does not, such as Passage, rejects the field
+whatever its value rather than reading zero as a no-op.
+
+> **Superseded, the next two paragraphs only**, by D5 of
+> [Amended 2026-09-16](#amended-2026-09-16-operations-are-code-not-config):
+> `amountMode` and `expectedQuantity` are gone from the wire entirely, and Pool's
+> take-all is `takeAll: true` on the body (D4). What follows them in this section,
+> from "Off by default" onwards, is current.
```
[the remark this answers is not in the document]
nhanlon2: This is from Claude: fixed in: ddbe54a02

DevDocs/adr/0011-inventory-operation-wizard.md
```diff
@@ -0,0 +1,201 @@
...
+which recorded three months of decisions as a config-driven design plus the amendments
+that overturned it. Only the reasoning that still describes the shipped system is kept
+here. The mechanics live in the code and its tests; this ADR records the WHY and the
+traps. The developer notes are `DevDocs/DeveloperNotes/InventoryOperationWizard.md` and
+`InventoryOperationWizardRequirements.md`.
+
+## Context
+
+Inventory needed seven subsample operations (Aliquot, Passage, Pool, Derive, Cryopreserve,
+Revive, Destroy): each creates a sample with subsamples and provenance links and decrements
```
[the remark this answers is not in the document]
nhanlon2: This is from Claude: fixed in: ddbe54a02

DevDocs/adr/0011-inventory-operation-wizard.md
```diff
@@ -0,0 +1,201 @@
...
+extension (`WAS_ALREADY_LOCKED`), so the edit lock cannot see one user's two overlapping
+requests, and that pair decrementing an origin twice from one read was the remaining lost
+update. `InventoryOperationInFlightOrigins` is a `ConcurrentHashMap` of the origins a
+request is acting on right now. The controller claims every origin (ascending, all or
+none) before taking the edit locks and releases the claim after the edit locks are given
+back, so the claim outlives the manager's transaction. A second request naming a claimed
+origin is a 409 `EDIT_CONFLICT`, whoever sent it. Presence is the whole rule: no timing
+window, no expiry. A time-based eviction was built and removed: no transaction timeout
+bounds the manager call, so a slow request is indistinguishable from an abandoned one, and
+guessing would hand a live request's origins to a second one. try-with-resources releases
```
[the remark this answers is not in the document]
nhanlon2: This is from Claude: fixed in: ddbe54a02

DevDocs/adr/0011-inventory-operation-wizard.md
```diff
@@ -0,0 +1,201 @@
...
+process name only, so one saved on a volume origin is offered on a mass one; a restored
+amount whose category no longer fits is repaired before it reaches the form.
+
+## Smaller decisions worth keeping
+
+- **Every operation has a process name**, fixed to the operation key when there is no
+  visible field: remembered values and derived sample naming key off it, and one uniform
+  key beats per-operation branches.
+- **Disposal dates are text fields holding an ISO date**: extra fields have no date type.
+- **"Template from the origin's sample" may leave a stray template on failure**: template
```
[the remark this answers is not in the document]
nhanlon2: This is from Claude: fixed in: ddbe54a02

DevDocs/adr/0011-inventory-operation-wizard.md
```diff
@@ -0,0 +1,201 @@
...
+- **Category and precision rules also apply to the created subsamples**: each new subsample
+  quantity must be in the origin's category (or the template's, when one is chosen) and
+  storable at 3dp. There is no top-level quantity on the wire; the server derives it.
+- **The documentation link targets an ELN document, notebook or Gallery file**; a link to an
+  Inventory record is rejected.
+- **Process is offered in every Inventory context menu except the picker**: epic RSDEV-1228
+  names both the list view and the item view as entry points.
+- **Bean Validation cascades over `origins`** only; the shared sample DTOs are not on the
+  wire. `origins` carries the 100 cap at binding so an over-long list is rejected before
+  Jackson materialises every element.
```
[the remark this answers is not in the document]
nhanlon2: This is from Claude: fixed in: ddbe54a02

src/test/java/com/researchspace/api/v1/controller/InventoryOperationFacadesMVCIT.java
```diff
@@ -0,0 +1,606 @@
...
+import java.util.List;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpHeaders;
+import org.springframework.test.context.web.WebAppConfiguration;
+import org.springframework.test.web.servlet.MvcResult;
+
+/**
+ * The seven typed operation endpoints (DevDocs/adr/0011 M6), each posted in the shape frozen in
```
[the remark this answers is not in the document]
nhanlon2: This is from Claude: fixed in: ddbe54a02

src/main/webapp/ui/src/Inventory/components/ContextMenu/ProcessAction.tsx
```diff
@@ -0,0 +1,130 @@
...
+    const busy = React.useRef(false);
+
+    const releaseAll = (records: Array<SubSampleModel>) => Promise.allSettled(records.map((r) => r.releaseLock(true)));
+
+    const openWithOriginsLocked = async () => {
+      if (busy.current) return;
+      busy.current = true;
+      let opened = false;
+      try {
+        const results = await Promise.allSettled(origins.map((o) => o.acquireEditLock()));
```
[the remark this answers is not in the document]
nhanlon2: This is from Claude: fixed in: ddbe54a02

src/main/java/com/researchspace/service/inventory/impl/SampleApiManagerImpl.java
```diff
@@ -377,6 +393,55 @@ private SubSample createSubSampleFromIncomingApiSample(
...
+      if (StringUtils.isNotBlank(name)) {
+        mergeTargets.putIfAbsent(name.trim().toLowerCase(Locale.ROOT), inherited);
+      }
+    }
+    Iterator<ApiExtraField> generated = apiSample.getExtraFields().iterator();
+    while (generated.hasNext()) {
+      ApiExtraField field = generated.next();
+      if (field == null
+          || StringUtils.isBlank(field.getOperationFieldKey())
+          || ExtraFieldTypeEnum.LINK.equals(field.getType())
```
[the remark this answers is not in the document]
nhanlon2: This is from Claude: fixed in: ddbe54a02

nhanlon2:
@pullfrog

src/main/webapp/ui/src/Inventory/components/ContextMenu/ProcessAction.tsx
```diff
@@ -0,0 +1,161 @@
...
+    const busy = React.useRef(false);
+
+    // The toolbar owns the selection, so emptying it unmounts this action mid-acquisition. Nothing
+    // is left to close the wizard then, so a lock granted after that point has to be given back
+    // here or it stays held until the server expires it.
+    const unmounted = React.useRef(false);
+    React.useEffect(
+      () => () => {
+        unmounted.current = true;
+        void releaseAll(taken.current);
```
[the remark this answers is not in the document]
nhanlon2: This is from Claude: fixed in: 5d1fc37fc

nhanlon2:
@pullfrog

nhanlon2:
@pullfrog

nhanlon2:
@pullfrog

nhanlon2 (commented):
This is Codex: I re-reviewed the entire current PR cumulatively against `main` at head `e6695d7e89abdfdc14861f5845b5fb01a2699a3f`, with adversarial focus on authorization, input validation, quantity/unit edge cases, template conformance, transaction rollback, generated-field persistence, preference storage, and the frontend edit-lock lifecycle. I found no new material findings to post. The process-local/multi-JVM concurrency limitation is explicitly accepted and documented in ADR 0011, so I did not raise it again. This was a static, read-only review; I did not run tests or modify code.

nhanlon2:
@pullfrog

DevDocs/DeveloperNotes/InventoryOperationWizardRequirements.md
```diff
@@ -0,0 +1,53 @@
+# Inventory Operation Wizard: Requirements and Implementation
```
tchoi-rs: Do we need this file?

src/main/java/com/researchspace/api/v1/controller/InventoryOperationsApiController.java
```diff
@@ -0,0 +1,265 @@
...
+    }
+    List<String> taken = new ArrayList<>();
+    // The in-flight claim is what refuses a second overlapping request from the SAME user (a
+    // double submit), which the edit lock treats as an extension. Claimed first, released last:
+    // the claim must outlive the edit locks, or a second request could claim the origin and then
+    // meet this request's still-held parent-sample lock, or (same user) pass it and run on after
+    // this request's release had stripped the locks from under it.
+    InventoryOperationManager.OperationOutcome outcome;
+    InventoryOperationInFlightOrigins.Claim claim = inFlightOrigins.claim(originGlobalIds);
+    try {
```
tchoi-rs: If you do `try (var claim = inFlightOrigins.claim(originGlobalIds)) {` then you don't need the extra try clause

src/main/java/com/researchspace/model/units/QuantityInfo.java
```diff
@@ -52,6 +52,24 @@ public static QuantityInfo of(String quantityString) {
...
 
+  /** Integer digits the DECIMAL(19,3) column holds: precision 19 minus scale 3. */
+  private static final int MAX_INTEGER_DIGITS = 16;
+
+  /**
+   * Whether {@link #setNumericValue(BigDecimal)} would store this value exactly. Quantities persist
+   * at 3 decimal places (HALF_UP), and the column is DECIMAL(19,3), so a value with more than
+   * {@value #MAX_INTEGER_DIGITS} integer digits does not fit at all: it passes a scale-only check
+   * (1E+30 has a NEGATIVE scale) and then fails at the INSERT. Trailing zeros lose nothing, so
+   * 0.5000 is storable; null is not.
```
tchoi-rs: It's a bit suspicious that we are not trimming trailing zeros

src/main/webapp/ui/src/Inventory/components/Operations/useOperationWizardLauncher.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const unmounted = React.useRef(false);
+  // The cleanup can run without an unmount (hot reload, StrictMode, a hidden Activity), so setup
+  // undoes it and `taken` is kept: the open wizard's next renewal re-takes the locks, and close
+  // must still give them back.
+  React.useEffect(() => {
+    unmounted.current = false;
+    return () => {
+      unmounted.current = true;
+      const held = taken.current;
+      void renewals.current.then(() => releaseAll(held));
```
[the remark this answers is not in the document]
nhanlon2: > Written by Claude, an AI coding agent.

Won't fix: this is dev-only. An effect restart without an unmount needs hot reload, StrictMode or a hidden `<Activity>`, and the Inventory app renders under none of them, so no user can reach this state. The operation itself is still guarded server-side by the controller's own origin locks and in-flight claim (ADR 0011 D6).

src/main/java/com/researchspace/service/inventory/operations/LabelResolver.java
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  /**
+   * ICU-formats the pattern only when args are supplied: an unconditional format() would mangle a
+   * literal apostrophe in an argument-free name, since ICU MessageFormat treats it as an escape
+   * character.
+   */
+  static LabelResolver fromMessageSource(MessageSource messages, Locale locale) {
+    return (key, args) -> {
+      String pattern = messages.getMessage("inventory:" + key, null, "inventory:" + key, locale);
```
tchoi-rs: This would hide label i18n keys from static analysis

src/main/java/com/researchspace/service/inventory/operations/OperationQuantityRules.java
```diff
@@ -0,0 +1,174 @@
...
+          field,
+          "errors.inventory.quantity.unitInvalid",
+          new Object[] {unitId},
+          "The quantity's unit id is not a unit.");
+    } else if (!RSUnitDef.getUnitById(unitId).isAmount()) {
+      errors.rejectValue(
+          field,
+          "errors.inventory.quantity.unitNotAmount",
+          new Object[] {unitId},
+          "The quantity's unit is not a mass, volume or dimensionless unit.");
```
tchoi-rs: New error messages do not need fallback English messages

src/main/java/com/researchspace/service/inventory/operations/PassageOperation.java
```diff
@@ -0,0 +1,89 @@
...
+ * Passage: carries a culture forward a generation. It creates a new sample derived from the origin
+ * and stamps it with the next passage number.
+ */
+@Component
+public class PassageOperation extends CreatingOperation<ApiInventoryOperationRequests.Passage> {
+
+  static final String NUMBER_FIELD_KEY = "operations.passage.numberField";
+
+  /** JS's Number.MAX_SAFE_INTEGER: the ceiling above which the counter restarts. */
+  private static final long MAX_SAFE_INTEGER = 9007199254740991L;
```
tchoi-rs: I don't know how far the passage value is supposed to be, but we should cap it at a much lower number and prevent passage beyond that number instead of silently allowing the passage counter to restart

src/main/webapp/ui/src/Inventory/components/Operations/__tests__/testOperations.ts
```diff
@@ -0,0 +1 @@
+export { operations } from "../operations";
```
tchoi-rs: Don't expose barrel imports

src/main/webapp/ui/src/Inventory/components/Operations/WizardTemplatePicker.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+  const preselected: TemplateOption | null =
+    selectedTemplateId != null
+      ? { id: selectedTemplateId, name: selectedTemplateName ?? "", globalId: "", record: null }
+      : null;
+  const [value, setValue] = React.useState<TemplateOption | null>(preselected);
+  const [inputValue, setInputValue] = React.useState(preselected?.name ?? "");
+
+  React.useEffect(() => {
+    void search.fetcher.performInitialSearch(null);
```
tchoi-rs: This doesn't actually clear the search results and causes confusion when the search term is cleared but the dropdown still contains the old search results.

Found via AI:
 
1. Open subsample **SS16 → Process → Passage → Next**.
2. Choose **an existing template** and open the picker: **six templates** appear.
3. Type `Browser`: results narrow to **two templates**.
4. Delete the search text.**Actual:** The input is empty, but the request still contains `query=Browser`, and only those two templates remain. Entering a single character (`C`) also resends `query=Browser`.

**Expected:** Empty or single-character input should restore the unfiltered list.

src/main/webapp/ui/src/Inventory/components/Operations/OperationWizard.tsx
```diff
@@ -0,0 +1,882 @@
...
+            ? { amountMode, perSubsampleAmounts: amountMode === "perSubsample" ? perSubsampleAmounts : {} }
+            : {}),
+        };
+        // One save of all three, applied to what the server holds now rather than this page's copy.
+        const latest = await fetchLatestOperationPreferences().catch((error: unknown) => {
+          console.warn("Could not read the latest remembered values; saving over this page's copy", error);
+          return normalizeOperationPreferences(storedOperationPreferences);
+        });
+        const name = operation.effect.processNameFrom ? String(values[operation.effect.processNameFrom] ?? "") : "";
+        setOperationPreferences({
```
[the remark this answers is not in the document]
nhanlon2: No. This is an impossible edge case. It also exists on main. A real user would have to somehow click perform in two tabs within milliseconds. Never happening.

src/main/webapp/ui/src/util/error.ts
```diff
@@ -68,3 +68,62 @@ export function getErrorMessage(error: unknown, fallback: string): string {
...
+ *   See {@link ORIGIN_INDEX}.
+ */
+export function getApiErrorDetail(
+  error: unknown,
+  fallback: string,
+  formatOriginIndex: (reason: string, index: number) => string,
+): string {
+  return Parsers.objectPath(["response", "data", "errors"], error)
+    .flatMap(Parsers.isArray)
+    .flatMap(([first]) => (typeof first === "undefined" ? Result.Error<unknown>([]) : Result.Ok(first)))
```
tchoi-rs: I don't think we should silently discard other errors here

src/main/webapp/ui/src/util/error.ts
```diff
@@ -68,3 +68,62 @@ export function getErrorMessage(error: unknown, fallback: string): string {
...
+    .flatMap(Parsers.isString)
+    .map((detail) => withOriginIndex(detail, formatOriginIndex))
```
tchoi-rs: Are other API responses guaranteed to have an `origin`?

src/main/webapp/ui/src/Inventory/components/Operations/useOperationWizardLauncher.tsx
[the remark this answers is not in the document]
tchoi-rs: If something fails in StrictMode, it will need to be fixed so that it runs in StrictMode correctly

src/main/webapp/ui/src/Inventory/components/Operations/useOperationWizardLauncher.tsx
```diff
@@ -0,0 +1,146 @@
...
+    void releasing.finally(() => {
+      busy.current = false;
+      onClose?.(performed.current);
+    });
+  };
+
+  const wizard =
+    open || origins.length > 0 ? (
+      <OperationWizard
+        key={open ? 1 : 0}
```
tchoi-rs: This codepath has a smell; what is this line trying to solve?

src/main/webapp/ui/src/__tests__/realI18n.tsx
```diff
@@ -0,0 +1,35 @@
+import { createInstance, type i18n as I18nInstance } from "i18next";
```
tchoi-rs: This should reuse __tests__/helpers/realI18n.tsx

PR #1131, opened 2026-09-17
RSDEV-1370: low-severity CodeQL cleanups

rs-fraser:
## Description

Low-severity CodeQL cleanups. Mostly defence in depth; the reachable ones are called out below.

- Unused cookie helpers removed in bbd1c1aa4
- archiveType constrained to xml, html or eln in 5d0e159a0
- Form permission query values bound in c266e879c
- Ambiguous and injectable regexes replaced in 6f8c8308a
- Search term parsing made consistent in b7e5ab655
- Markup built and parsed without string interpolation in b19df9d7d
- Zip entries checked against the extraction directory in d2470de7a
- B2INST submit links restricted to the configured server in 580d184e2

## What was actually reachable

- **Email validation ran an ambiguous regex before checking length**, on form fields with no size cap, so a long value meant quadratic matching. `6f8c8308a`
- **Search term splitting ran a quadratic separator-with-whitespace pattern on user-controlled terms**, in the validator, query builder and Lucene config. All three change together. `b7e5ab655`
- **A notebook name containing a double quote was submitted truncated** at the quote; name validation only blocks `/`. `b19df9d7d`

The form permission query interpolated four values that are all constrained upstream today (usernames are alphanumeric, group unique names are generated, permission strings are not user input); binding them removes the dependence on those distant guarantees.

## Behaviour changes

- Email addresses with more than one `@` (`a@b@c.com`) are now rejected. The pattern is shared, so this affects community signup as well as cloud group creation.
- `validateToken` now treats a newline-containing token as JWT-shaped where it previously did not. Not reachable via the Authorization header, which splits on whitespace; it is reachable via the `refresh_token` grant parameter, where it only changes which error a garbage token gets, since the token must still match a stored hash in `refreshGrant`.


rs-fraser:
@pullfrog

src/main/java/com/researchspace/dao/hibernate/FormDaoHibernate.java
```diff
@@ -138,61 +148,55 @@ private ISearchResults<RSForm> getForms(
...
+        andClauses.add(sqlIdSub);
+      }
+      for (String propName : constraintBasedPermission.getPropertyConstraints().keySet()) {
+        PropertyConstraint pc = constraintBasedPermission.getPropertyConstraints().get(propName);
+        if (pc.getName().equals("global")) {
+          if (requestedAction.equals(PermissionType.READ)) {
+            String access1 = " form.worldPermissionType in (" + join(RW, true) + ")";
+            andClauses.add(access1);
+          } else if (requestedAction.equals(PermissionType.WRITE)) {
+            String access1 = " form.worldPermissionType = " + requestedAction;
```
nhanlon2, line 163: This comment is from Claude. `= WRITE` goes into the native SQL unquoted, so the database reads it as a column name, and `WRITE` is a reserved word besides. The READ branches just above quote their values via `join(RW, true)`. Same on lines 171 and 179. A form search that carries a write-permission check would fail with a SQL error instead of returning results. It predates this PR, but this is the natural change to bind it in, since every other literal in the method just became a bound parameter.

src/main/java/com/researchspace/model/dtos/CreateCloudGroupValidator.java
```diff
@@ -14,10 +13,9 @@
 @Component("CreateCloudGroupValidator")
 public class CreateCloudGroupValidator implements Validator {
 
-  private Matcher matcher;
-
   /** Valid Email pattern for creating new group */
-  public static final Pattern EMAIL_PATTERN = Pattern.compile("^[^\\\\<>' \"]+@[^\\\\<>' \"]+$");
+  // "@" is excluded from both sides so exactly one split point exists and matching stays linear
+  public static final Pattern EMAIL_PATTERN = Pattern.compile("^[^\\\\<>' \"@]+@[^\\\\<>' \"@]+$");
```
nhanlon2, line 18: This comment is from Claude. The tightened pattern still allows carriage return and newline inside both halves, so an address like `victim@example.com\r\nBcc:attacker.test` passes. That was true of the old pattern too, so it is not something this change introduced, but since the line is being hardened anyway, adding `\r\n` to both excluded sets would close it. Worth confirming the mail layer rejects control characters if you would rather leave it as is.

src/main/java/com/researchspace/dao/hibernate/RSQueryBuilder.java
```diff
@@ -284,7 +284,7 @@ private List<SearchPredicate> getPredicateList(SearchPredicateFactory f, List<Te
       if (term.field().equals(FieldNames.MODIFICATION_DATE)
           || term.field().equals(FieldNames.CREATION_DATE)) {
         // Validator should have already checked that the string contains ';'
-        String[] toAndFrom = term.text().split("\\s*[,;]\\s*", -1);
+        String[] toAndFrom = splitDateRangeTerm(term.text());
         Date dateTo, dateFrom;
```
nhanlon2, line 289: This comment is from Claude. `splitDateRangeTerm` can return a single element (a term with no `,` or `;`), but lines 292 and 295 index `[0]` and `[1]` unconditionally. The only thing preventing that is a validator in a different package, which the new javadoc already calls out as a coupling. Throwing `SearchQueryParseException` from the helper when `halves.length != 2` would let the query builder defend its own precondition, and it is already imported in this file.

nhanlon2 (commented):
This comment is from Claude. Reviewed the whole PR (9 commits, 27 files) across layering, security, persistence, i18n, frontend and general quality. Three issues worth acting on are inline below; everything else the reviewers raised was lower priority and is not posted here.

tchoi-rs (approved):
Frontend part looks fine to me!

nhanlon2 (approved):

PR #1132, opened 2026-09-17
PRT-947: Hide chemistry editor actions when the user has Chemistry disabled

rs-fraser:
## Description

Fixes PRT-947.

Chemistry has two gates: the sysadmin setting (`chemistry.available` with `chemistry.provider=indigo`) and the user's Chemistry card on the Apps page. Some editor paths only checked the system gate, so users with Chemistry disabled could insert `.mol` files from the Gallery as chemistry elements and see unusable Ketcher and stoichiometry actions.

This change uses the combined system/user gate for Gallery insertion and chemistry-specific toolbar actions. Generic details, download and resize actions remain available.

Regression tests cover both gate states and toolbar contents.

## Side effects

- Deployments with a non-indigo or empty provider now show generic toolbar actions on existing chemistry elements.
- With Chemistry disabled, Gallery thumbnails for `.mol` and other chemistry files use the generic file-type icon and insertion produces a plain attachment.


src/main/webapp/scripts/externalTinymcePlugins/contexttoolbars/plugin.min.js
```diff
@@ -77,17 +77,17 @@ function addKetcherChemContextToolbar(editor){
...
 		position: 'node',
 		scope: 'node'
 	});
 	editor.ui.registry.addContextToolbar('chemicalElementEditableKetcher', {
 		predicate: function (node) {
-			return !(isChemFileAttachment(editor.selection.getNode())) && editor.dom.is(node, 'img.chem') && RS.chemistryProvider === 'indigo';
+			return !(isChemFileAttachment(editor.selection.getNode())) && editor.dom.is(node, 'img.chem');
 		},
-		items: 'downloadImage | resizeImage | ketcherEditable | stoichiometry',
+		items: 'downloadImage | resizeImage' + (RS.chemistryEnabled && RS.chemistryProvider === 'indigo' ? ' | ketcherEditable | stoichiometry' : ''),
```
nhanlon2, line 90: This comment is from Claude. The `stoichiometryTableOnly` toolbar just below still offers `stoichiometry` with no gate, so a user with Chemistry switched off can still reach the stoichiometry action from an existing table block. The dialog does re-check and shows a disabled message, so it is a UI inconsistency rather than a hole, but it looks like the same gate belongs there.

src/main/webapp/scripts/pages/workspace/editor/tinymce5_configuration.js
```diff
@@ -800,7 +800,9 @@ function initTinyMCE(selector) {
 		}
 
 		const chemistryProvider = properties["chemistry.provider"];
-		chemistryAvailable = integrations.CHEMISTRY.available;
+		// Both gates must be on: the sysadmin setting and the user's Apps page toggle.
+		chemistryAvailable = chemistryEnabled;
+		RS.chemistryEnabled = chemistryEnabled;
```
nhanlon2, line 805: This comment is from Claude. This flag is set in the integrations `done` handler while the toolbar reads it at editor-construction time, so the buttons depend on an ordering rule spread across two files. It is safe today and fails closed, but a default `RS.chemistryEnabled = false` up front (or a one-line comment here) would stop a future reshuffle silently hiding chemistry for everyone.

src/main/webapp/ui/src/tinyMCE/__tests__/chemistryIntegration.test.ts
```diff
@@ -0,0 +1,159 @@
...
+    ui: {
+      registry: {
+        addButton: vi.fn(),
+        addContextToolbar: (name: string, toolbar: ContextToolbar) => toolbars.set(name, toolbar),
+      },
+    },
+  };
+  const sandbox = {
+    $: vi.fn(),
+    RS: { chemistryEnabled, chemistryProvider: "indigo" },
```
nhanlon2, line 111: This comment is from Claude. Every case here hard-codes `chemistryProvider: "indigo"`, so the new non-indigo behaviour the PR description calls out is not covered. One more case with a non-indigo provider and `chemistryEnabled: true`, asserting the generic-only `items` string, would pin it.

nhanlon2 (commented):
This comment is from Claude. Automated review of the three changed files (plus the downstream consumers of `chemistryAvailable`) by three reviewers run in parallel: 3 issues worth raising, listed inline. One note that did not fit a changed line: the server side does not check the CHEMISTRY integration entitlement at all (RSChemController and StoichiometryApiController resolve the user and rely on record-level permissions), so the Apps-page toggle stays a UI-level gate only. That is pre-existing and out of scope here, just worth not relying on.

tchoi-rs (approved):

src/main/webapp/scripts/externalTinymcePlugins/contexttoolbars/plugin.min.js
nhanlon2: > Written by Claude, an AI coding agent.

This is from Claude: fixed in: 3d20046be

src/main/webapp/scripts/pages/workspace/editor/tinymce5_configuration.js
nhanlon2: > Written by Claude, an AI coding agent.

This is from Claude: fixed in: 3d20046be

src/main/webapp/ui/src/tinyMCE/__tests__/chemistryIntegration.test.ts
nhanlon2: > Written by Claude, an AI coding agent.

This is from Claude: fixed in: 3d20046be

nhanlon2 (approved):

PR #1139, opened 2026-09-22
RSDEV-1505: Don't name a linked Instrument the searcher cannot read

rs-nicof:
A PID registry search still tells every user that a hit is already linked to an Instrument in this
RSpace, so Import is refused for a stated reason, but it names that Instrument only to someone
allowed to read it (RSDEV-1505, subtask of RSDEV-1033, follow-up of RSDEV-1325).

## Why

`PidinstLookupManagerImpl` copied `getConnectedRecordGlobalIdentifier()` onto every hit and into the
409 message without ever looking at who was asking. The registry record itself is public, so nothing
about it needs withholding, but the existence and Global ID of an RSpace Instrument are not. A user
in no shared group could paste a PID, be told it was taken, and be handed `IN52` along with it.

The disclosure was pre-existing from RSDEV-1326 and only became visible once the RSDEV-1325 dialog
started rendering that Global ID as a clickable chip.

## What it does

Each hit gains an always-serialised `alreadyLinked` boolean, so a client can refuse Import before the
409 whoever is asking. `linkedInstrumentGlobalId` stays, but is now set only when the caller passes
`InventoryPermissionUtils.canUserReadOrLimitedReadInventoryRecord`, and is omitted from the JSON
otherwise. `importInstrument` runs the same gate to choose between the existing named message and a
new `pidinstAlreadyLinkedNoAccess` one.

The dialog branches on the two fields at its three consumer sites, all of them gating on
`alreadyLinked` and letting the Global ID choose only the wording. Where the chip cannot be shown,
the cell carries the same warning "No access" chip the link fields already use for an unreadable
target, with the explanation in its tooltip.

Additive on the wire, so no existing client breaks. No new endpoint, no schema change, no new
dependency. A PID nothing links imports exactly as before, and the importing user always owns what
they created, so they keep seeing their own Global ID.

## Decisions

Settled with Nico on 2026-09-21, recorded as an amendment to `DevDocs/adr/0009` decision 4 and a new
**Already-linked marker** term in `CONTEXT.md`:

- **Mark and name are separate.** Everyone learns the PID is taken; only a permitted reader learns by
  which Instrument. Refusing without a reason was rejected as user-hostile, and refusing with the
  name was the bug.
- **Read or limited read, not full read.** That is the rule deciding everywhere else in Inventory
  whether a caller gets a record or only its no-access view. It is deliberately looser than the rule
  behind the *No access* link-target pill, which uses plain read: a viewer with only limited read,
  through a container they can read or a document whose List of Materials lists the Instrument, is
  named it here, because both of those contexts already show its Global ID.
- **One sentence, reused.** "This PID is already linked to an instrument in RSpace that you cannot
  access." for the preview alert, the Import refusal and the 409.
- **No new ADR.** The trade-off belongs to 0009 and the principle is already in 0002.

## Things a reviewer may want to check

- **This stops the search volunteering the Global ID; it is not a secrecy guarantee.** I had assumed,
  and the plan stated, that a withheld chip would otherwise lead to a 404. It would not.
  `GET /instruments/{id}` deliberately makes no read assert and answers 200 with a name-only public
  view, which the UI renders as "You do not have permission to see any of the details of this
  instrument". Only the endpoints going through `assertUserCanReadInstrument` (image, thumbnail,
  revisions) actually refuse. So a guessed numeric id still yields an Instrument's name. The
  predicate choice is still the right one and the ticket's ask is met, but the ADR, the glossary and
  the Javadoc now say "does not volunteer" rather than "cannot reach". Flagging it in case anyone
  wants enumeration treated as a separate problem.
- **Sysadmins keep seeing the Global ID**, because they pass the same predicate. Verified rather than
  assumed: a sysadmin still opens another user's Instrument with a 200 and a full record.
- **The visibility check runs once per already-linked hit**, bounded by the 50-hit page. For a caller
  who cannot plainly read the holder it reaches the list-of-materials query inside limited read. The
  Javadoc says so now rather than claiming the whole method is one query. Memoizing per holder was
  considered and rejected: one identifier per record makes distinct holders the normal case, so it
  would not touch the worst case.
- **A row whose identifier has no record counts as hidden, not unlinked.** The PID is taken either
  way. Unreachable in production, but it decides a disclosure, so it is pinned by a test rather than
  only described.
- **The limited-read half of the predicate is not exercised against a running deployment**, only
  through the unit tests. It needs a List of Materials or sharing fixture, and the predicate itself
  is long-standing shared code that this change only calls.

## Verification

66 fast backend tests green (`PidinstLookupManagerImplTest` 47, `ExternalUpdateMessageKeysTest` 19),
`PidinstLookupApiControllerMVCIT` 10 run and 0 skipped locally, 23 jsdom tests on the dialog, the
browser-mode spec green on Chromium, and tsc, biome and both i18n checks clean.

Each new guard was proved by reverting it and watching the named test go red, including at the wire:
removing the permission clause makes the MVCIT leak the owner's Global ID to the outsider, and
removing the null-holder guard makes the unit test that pins it fail.

Also verified end to end against a running deployment and the real B2INST test registry with three
users. The owner sees the marker with the Global ID on the search and a 409 naming it; a user in no
shared group sees the marker with the field absent from the JSON entirely and a refusal naming
nothing; and in the dialog that user gets the no-access cell, the anonymous preview alert, and an
Import that creates nothing.

## Review pass

A six-reviewer pass (layering, i18n, security, persistence, frontend, dead code and coverage) raised
nothing critical. What it changed, in the two commits after the first:

- The no-access cell was **clipped and unreadable**: a grid cell is truncated with an ellipsis and
  only gets a hover title when it has no custom renderer. Confirmed in a browser, then fixed.
- The docs claimed the marker follows the **same rule as the No access link-target pill**. It does
  not, as described under Decisions. Corrected in `CONTEXT.md` and ADR 0009.
- The annotation Javadoc still claimed "one query rather than one per hit" after the permission
  check moved inside the loop.
- The wire flag was called `linked`, which **already means something else** in the same API document
  (`ApiInventoryDOI.linked`, a PID minted outside RSpace). Renamed to `alreadyLinked`.
- The 409 exception carried a Global ID field **no production code read**; the error response is
  built from the message alone. Deleted, with the tests moved onto the message.


rs-nicof:
AWS instance: https://rsdev-1505-search-link-instrument-user-cannot-see-7d7680db-5.researchspace.com/

nhanlon2 (commented):

Parallel review (layering, i18n, security, persistence, frontend, quality). No code issues found; the disclosure gate looks correct and fails closed. Two doc mismatches below.

DevDocs/adr/0009-linked-identifiers-for-imported-pids.md
```diff
@@ -56,6 +56,22 @@ and API field names were ported instead.
    in the deployment is refused with 409 naming the instrument. That check matches the
    provider's record id as well as the PID, because a B2INST identifier this deployment
    registered stores the record id: its Handle is minted on publish and never written back.
+   *Amended 2026-09-21 (RSDEV-1505):* the 409 names the instrument only when the caller may
+   read it, by the read-or-limited-read rule that decides everywhere else whether a caller gets
+   a record or only its no-access view; otherwise it says only that an instrument they cannot
+   access already holds the PID. The search response marks every such hit `linked` and carries
```
nhanlon2: > Written by Claude, an AI coding agent.

The field is `alreadyLinked`, not `linked` (renamed in this PR), so readers of the ADR would look for a key that does not exist.

src/main/webapp/resources/rspace_api_inventory_specs_2_27_0.yaml
```diff
@@ -3027,8 +3027,11 @@ paths:
         not returned, and pasting such a PID yields no hit.
 
 
-        Each hit carries `linkedInstrumentGlobalId` when an Instrument in this deployment already
-        links that PID, so a client can disable importing it before the `409`.
+        Each hit carries `alreadyLinked` when an Instrument in this deployment already links that
```
nhanlon2: > Written by Claude, an AI coding agent.

This reads as if `alreadyLinked` only appears on linked hits, but the schema below and the code say it is always present. Suggest: "Each hit carries `alreadyLinked`, true when an Instrument in this deployment already links that PID, so a client can refuse importing it before the `409`; and `linkedInstrumentGlobalId` only when...". There is also a stray line break after "PID, so".

DevDocs/adr/0009-linked-identifiers-for-imported-pids.md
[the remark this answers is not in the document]
rs-nicof: > Written by Claude, an AI coding agent.

Confirmed and fixed in 32351f578. Worth noting it was not a dangling reference: `linked` is still a real field on the identifier payload, documented two decisions earlier in the same ADR, so the amendment resolved to the wrong field rather than to none. That collision is what prompted the rename. The two remaining mentions of `linked` in the ADR are that other field and are correct as they stand.

rs-nicof: > Written by Claude, an AI coding agent.

Agreed, fixed in 32351f578. Slightly worse than a missing key: `linked` is a real field on the identifier payload that the same ADR describes earlier, so a reader would land on the wrong one. Left the two other mentions alone, since those genuinely are that field.

src/main/webapp/resources/rspace_api_inventory_specs_2_27_0.yaml
rs-nicof: > Written by Claude, an AI coding agent.

Agreed, fixed in 32351f578 with your wording. The schema entry directly below already said "Always present", so the two halves of the spec contradicted each other, and a client reading only the endpoint text would have taken an absent key as "not linked" rather than as one it should never see. Reparsed the file afterwards to confirm it still loads.

On the line break: YAML folds it to a space, so the published text never showed it. Tidied in the same edit anyway.

rs-nicof:

Both reviews landed on the same two documentation mismatches, fixed in 32351f578 and answered in their threads. No code changed, so the behaviour the reviews checked is untouched.

- The ADR amendment still used the pre-rename name for the search flag. That name belongs to a different, real field on the identifier payload, so the reference pointed at the wrong thing rather than at nothing.
- The API endpoint text implied the flag only appears on already-linked hits, contradicting its own schema entry. It is on every hit.

Verification was a reparse of the spec plus a read of the rendered description, since nothing in the suite loads that file. The pullfrog thread is resolved; the two human threads are left open.

nhanlon2 (approved):

src/main/java/com/researchspace/service/inventory/PidinstAlreadyLinkedException.java
```diff
@@ -2,19 +2,14 @@
...
+ * the localized message alone. That message names the instrument only when the caller may read it;
+ * otherwise it says only that an instrument they cannot access holds the PID (RSDEV-1505; ADR 0009
+ * decision 4, amended).
  */
 public class PidinstAlreadyLinkedException extends RuntimeException {
 
-  private final String linkedInstrumentGlobalId;
-
-  public PidinstAlreadyLinkedException(String localizedMessage, String linkedInstrumentGlobalId) {
+  public PidinstAlreadyLinkedException(String localizedMessage) {
```
tchoi-rs: I would prefer exception messages to carry only the data necessary to render the error, then let the actual site (controller) decide how to render the exception (e.g. if we, in the future, pass this exception down as JSON and let the client render the error)

src/test/java/com/researchspace/service/inventory/impl/PidinstLookupManagerImplTest.java
```diff
@@ -86,11 +92,18 @@ void setUp() {
     lenient().when(b2instConnector.isConfiguredAndEnabled()).thenReturn(true);
     // nothing is linked unless a test says so; search annotates every page through this one query
     lenient().when(doiDao.findActiveByIdentifiersAndType(any(), any())).thenReturn(List.of());
+    lenient().when(messages.getMessage(anyString())).thenAnswer(inv -> inv.getArgument(0));
```
tchoi-rs: nit: FYI I removed all usages of `lenient()` a while back - I'll do another pass after this PR is merged but the idea is that each test should define their mocks precisely.

tchoi-rs (approved):

src/main/java/com/researchspace/service/inventory/PidinstAlreadyLinkedException.java
rs-nicof: > Written by Claude, an AI coding agent.

Agreed, done in da3d0eb99. It now carries the message key and its arguments and the advice resolves them, which is what four other exceptions handled there already do, so this stops being the odd one out. A caller that later wants the parts as JSON has them.

One thing I kept where it was: the choice of which key to throw. That depends on whether this caller may read the linking instrument, which only the service can answer, so it still picks between the naming key and the no-access one. The no-access key now carries no arguments at all, so the hidden Global ID is not in the exception for any future renderer to find, which is a bit stronger than what we had.

Nothing on the wire changes, and the two wire-level tests assert the same 409 text. Both new boundary tests were checked by mutation.

src/test/java/com/researchspace/service/inventory/impl/PidinstLookupManagerImplTest.java
rs-nicof: > Written by Claude, an AI coding agent.

Understood, leaving it for your pass rather than colliding with it. One trap worth knowing before you get here: one test asserts the permission checker is never touched at all, so its stub has to stay unused in that test. Making these strict as they stand fails loudly rather than quietly, but the tempting shortcut of deleting the stub would erase the thing that test exists to prove. Pushing each stub down into the tests that use it is the safe shape.

Unrelated but in passing, the change above made one of these stubs genuinely dead, so it is gone in the same commit. FYI there are still a handful of other test classes using it, so the pass is a bit bigger than this file.

